Issue #1997, opened 2014-11-26
Gas-Generator output limit fixed based on hydrogen


Dannerick:
The output limit of a gas-generator is based on the generation using hydrogen (default 400). If using ethylene (default 705), the production is significantly higher, but the output limit of the gas-generator means you can not utilize its full potential. This is because the output limit is set when the object instance is created. The output limit needs to be adjusted when fuel is added to the generator.

PR #2001, opened 2014-11-28
Fix for #1997 gasGenerators incorrectly limiting output

Dannerick:
This change fixes the issue I was seeing where gasGenerators are limiting energy output based on hydrogen burning even though ethylene, when burned, produces more energy. This fix only checks and sets the output when a new fuel is first put into the generator, or when the when the generator is loaded up in the world.


aidancbrady:
Thanks!

PR #2089, opened 2015-01-13
Update pt_BR.lang

al-myr:


LorenzoDCC:
Dude, in no way "paxel" means enxó.  Enxó means adz, and paxels are tools that only exist within mekanism :neutral_face:
Also, "calças" is right, man ._.

al-myr:
Ok!

PR #2090, opened 2015-01-13
Fixed server crash in 7.1.2

halvors:
This fixes that the server crashes due to that the in enableAmbientLighting and ambientLightingLevel MekanismGeneratorsClient is client only.

I'm not sure if this is the right way to fix this.

I've confirmed that this solves the crash problem.


halvors:
Closing this as it was fixed by @unpairedbracket :)

PR #2128, opened 2015-02-13
Shift-click items to top (fill) slot

chrixian:
I don't think a day goes by where I don't shift-click my jetpack into the gas tank and when it ends up in the bottom slot I cry out in pain :cry: 

If 99.9999% of the time you open the gas tank to manually fill a jetpack or scuba tank, it makes sense to have items shift-click into the top slot rather than the bottom one. (And items in either of the gas tank slots will shift-click to the player inventory now as well.)


thecodewarrior:
I agree, one stop-gap measure I use is putting an empty gas tank in the bottom slot, so the top is the only free slot and the jetpack goes there.

PR #2140, opened 2015-02-18
Fix for setting generator output when ethylene already in tank when plac...

Dannerick:
If a gas generator entity has ethylene in the tank when it is placed, the max output will not be calculated correctly until a restart or the tank is emptied


aidancbrady:
Thanks

Issue #2069, opened 2015-01-03
Cardboard box blacklist


ZeekDaGeek:
It seems like the Cardboard Box processor doesn't consider the new format for item ID names. I've attempted to blacklist warded items in two ways, neither work.

Attempt 1: http://pastebin.com/aVwkkG1M
Attempt 2: http://pastebin.com/rchgFm8P

The way item IDs work now is ModID:ItemName, correct? The way processing of the blacklist works it prevents this from being accepted:

https://github.com/aidancbrady/Mekanism/blob/master/src/main/java/mekanism/common/BoxBlacklistParser.java

Possibly it needs to be updated to ModID:ItemName;META or some other separator that isn't allowed in an item name like a space?

FakoTheGreat:
I've been trying to blacklist items as well, and using unlocalized names also does not solve the issue, rendering the feature unusable with modded blocks (it will take vanilla blocks just fine).

aidancbrady:
Should be fixed now

PR #2141, opened 2015-02-18
[#2069] Fix for modded items being added to cardboardbox blacklist

Dannerick:
This is a fix for being able to put modded items into the cardboard box blacklist (issue #2069).


aidancbrady:
Thanks

Issue #2146, opened 2015-02-20
add config for Disassembler usage


Greylocke:
Please add a config for the Disassembler usage. In my opinion, it does not use enough power to balance its benefit. Some might agree, some might not... but with a config we can all set it to whatever we like.

thanks!

aidancbrady:
Done, thanks to @Dannerick!

PR #2165, opened 2015-03-03
New config entries

Dannerick:
I added two new config entries. 
One for the atomic disassembler energy usage for request #2146.
The second for the ethylene burn time as I always felt that ethylene produced too much gross power compared to how much is produced from a single piece of wheat. Lowering to burn time from 40 to 20 or even 10 seemed to make a better balance on the servers I have run.


aidancbrady:
Excellent PR, thanks

PR #2186, opened 2015-03-08
Completely remade Seismic Reader.

### LINKED ISSUE #1565: Seismic Reader GUI enhancements
I was chatting with Vaygrim the other day and he had some nice ideas about increasing the ease-of-use for the Seismic reader. Namely:
1. Add tooltip/mouseovers for the legend along the right-hand side of the Reader GUI. Currently there is no explanation as to what the colors mean, so the player is left to guess.
2. increase the size of the color blocks on the reader, Certainly horizontally. Expanding vertically would require that the reader GUI be increased, unless you implement a "zoom" type feature, as described in (a) below
3. Add tooltip/mouseovers for the readout on the GUI. This will greatly help players with color perception issues.

some extra feature ideas for the system:
a. Zoom-in to an area. Click on a section of the readout, and the 64-block chart becomes a 20-block chart with more detail, centered on the area you clicked. Click the chart again to get back to the zoomed-out mode.
b. Advanded Reader allows DigiMiner-type filters. Reports "pings" for the filtered materials. Unlike the DM, the Seismic Reader should allow for "pinging" caves and fluids.
c. Advanced Reader allows a third level of zoom, which turns the 20-block chart into a "% of total" chart for each type of material in the selected level..(Or maybe an average of the selected level +/- 1 level, for a better idea of what's down there)

Archtikz:
It may no be 100% as efficient as possible. I've tried to bug test as much as possible! Resolves #1565


Archtikz:
Note, I havn't done any proper gui localisation due to not wanting to mess with the lang files just yet.

Archtikz:
If you want me to change anything also, tell me and I'll go make the change.

PR #2197, opened 2015-03-11
Makes Upgrades Configurable

iPencil:
Allows setting how many upgrades should go into a machine. Works for <8
and >8 equally well. Also adjusts new max stacksize of Upgrades
depending on how many allowed in machines.


iPencil:
A lot of people have been complaining on my server that Energy Upgrades are too op, this should solve it :)

It will also allow upgrades to go beyond the current 8x and they work just like expected!

unpairedbracket:
Would reducing max upgrade multiplier not work?

aidancbrady:
Sorry, but I'd rather not do this - you can do exactly what you want by reducing the multiplier, as @unpairedbracket said.  The upgrade system has been refactored for v8, and I don't want to have extra config entries when I don't need them.

iPencil:
Reducing the max multiplier will have what effect exactly? It's going to screw over speed upgrades as well. I also don't see how more config entries are harmful, most users don't even look at them.

PR #2218, opened 2015-03-18
Added a nullary constructor to AdvancedMachineInput

ganymedes01:
On line 587 of https://github.com/aidancbrady/Mekanism/blob/development/src/main/java/mekanism/common/recipe/RecipeHandler.java

Class.newInstance() can only be called by classes that contain an accessible constructor that takes no arguments, throwing an exception otherwise.


aidancbrady:
Fixed :)

ganymedes01:
Thank you:3
I've re-written the mekanism support in AOBD to use the new IMC you implemented.
It's working like a charm now!

PR #2221, opened 2015-03-18
Cleanup, cleared some warnings.

halvors:
No scary changes, just cleared some warnings.
Basically added default perameter to switch statements.


aidancbrady:
Ah, thanks

PR #2247, opened 2015-03-20
Update de_DE.lang

Vexatos:
Comments will come back later today. I was in a hurry


aidancbrady:
Thanks for the speedy update!

PR #2252, opened 2015-03-20
Added Norwegian translation

halvors:
I'm from Norway, so i took the time to transelate Mekanism into Norwegian ;-)


aidancbrady:
Thanks :)

PR #2272, opened 2015-03-21
Chemical Infuser is now improved by Speed upgrades.

### LINKED ISSUE #2263: [v8] Speed upgrades in Chemical Infuser seem to have no effect
Putting eight speed upgrades does not appear to make the Infuser work any faster. The upgrade display says "Effect: 256.0x" but the Infuser does not produce gas any faster. The input tanks remain full of Hydrogen and Chlorine (displaying "9999" which tells me it only processes 1 per tick), and the output rate is the same as it is without any upgrades.

codewarrior0:
Uses the same logic as Electrolytic Separator, adding a `scale` option to the Infuser's recipes and calling `getUpgradedUsage` to find out how many operations to do.

Fixes #2263

### DIFF
--- a/src/main/java/mekanism/common/recipe/inputs/ChemicalPairInput.java
+++ b/src/main/java/mekanism/common/recipe/inputs/ChemicalPairInput.java
@@ -36,23 +36,30 @@ public class ChemicalPairInput extends MachineInput<ChemicalPairInput>
 	}
 	
 	public ChemicalPairInput() {}
-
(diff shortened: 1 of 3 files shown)


aidancbrady:
Self-merged, thanks for the help :)

codewarrior0:
Gee, that's a great way of saying you used my code without giving me any credit for it! >:(

aidancbrady:
Hehe, no such intent

codewarrior0:
There isn't even any reference to the pull request, or the original issue. I feel robbed.

Issue #2330, opened 2015-03-23
[8.0.0] Digital Miner loses replacement configuration when breaked and replaced.


halvors:
Digital Miner loses replacement configuration when breaked and replaced.

Referencing this issue #2286, similar problem. But this works now, just that the settings are not stored somehow.

aidancbrady:
8.0.0 or 8.0.1?

halvors:
I'm running version 8.0.1, fixed the problem in pull request #2331.

PR #2331, opened 2015-03-23
Fixed OreDictionary filter in Miner not writing "requireStack" to NBT.

halvors:
Fixed OreDictionary filter in Miner not writing "requireStack" to NBT.

This solves issue #2330.


aidancbrady:
Great PR, thanks

PR #2373, opened 2015-03-26
Fix API package name

yogpstop:
invalid class name in Optional.Interface
So couldn't strip interface exactly.


aidancbrady:
Self-merged. Thanks for bringing this to my attention!

Cracks777:
has this been merged into the dev builds?

PR #2547, opened 2015-04-16
Update fr_FR.lang

Mazdallier:


Mazdallier:
how many caracters before \n in tooltips ?

aidancbrady:
Thanks!  I don't have a designated amount, just whatever looks nice.

Issue #2541, opened 2015-04-15
Computercraft APIs for new Multiblocks


mazen:
Since the new induction matrix can store more then 2^31 RF it causes trouble with the 32 bit behavior of ComputerCraft. OpenPeripherals easily overflows if getEnergyStored() / getMaxEnergyStored() is called in this case. .

Can we have some methods which can provide the energy stored / max energy stored (maybe, returning a table with a value + factor) ?

Kodos-Atoz:
Computronics adds special handlers for both mekanism V8 as well as Draconic Evolution's large power storage.  I'd recommend adding that, as it works for both OpenComputers and ComputerCraft.

emperor06:
As you stated, getEnergyStored() and getMaxEnergyStored() are part of Open Peripheral, which has nothing to do with Mekanism nor ComputerCraft. Since those methods are for the RF api, I doubt they will ever return a double but it may be a good idea to ask the author of Open Peripherals.

I implemented those 2 functions on my induction matrix (as well as an average output) and it works perfectly fine (ComputerCraft handle doubles natively). I can send a pull request if the Meka team wants it.

aidancbrady:
I'd rather not do any kind of additional integration along this path.  I'd be happy to accept a feature PR though :)

PR #2551, opened 2015-04-16
Added ComputerCraft support to Induction Matrix.

emperor06:
I've never done a pull request before. Anyway, this one adds a few functions for ComputerCraft with Induction Matrices:
- getEnergyStored() and getMaxEnergyStored() which both return RF values (so it matches the common signature of Open Peripherals, making existing CC scripts compatible with the Induction Matrix)
- getEnergyStoredMJ() and getMaxEnergyStoredMJ(), same as above but using Joules
- getLastOutput() and getOutputCap().

Honestly, I'm not very fond of those method names. But well, they are easy to change.
Nota: I have put this support on the basic induction block so a computer can be plugged pretty much anywhere on the structure (including edges and corners). Maybe it's best to put it on induction ports? I don't know.

### DIFF
--- a/src/main/java/mekanism/common/tile/TileEntityInductionCasing.java
+++ b/src/main/java/mekanism/common/tile/TileEntityInductionCasing.java
@@ -4,6 +4,7 @@ import io.netty.buffer.ByteBuf;
 
 import java.util.ArrayList;
 
+import mekanism.api.MekanismConfig.general;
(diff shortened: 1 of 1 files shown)


emperor06:
Oh, forgot to mention, that's in regard to #2541

aidancbrady:
Lovely PR.  Induction casing was the right block to integrate with.

PR #2579, opened 2015-04-23
Fixing NPE from EnergyAcceptorWrapper.java not handling ticks

RAWRwins254:
Code was incorrectly handling non-TileEntity possibilities thus causing an NPE. NPE is now fixed as the system now does not return a null value as easily.


RAWRwins254:
i should probably also note that these crashes seem to be chunk loading related. I'd come back near my base after being far away and then the server would crash with an NPE. Until you guys can come up with a better fix, i recommend accepting this pull request, merging the development branch with master and then releasing a stable build. I know a lot of peeps are waiting on the fluid fix for Mekanism v8 with forge 1365+.

PR #2750, opened 2015-06-16
Make jetpack size configurable

Canisaur:
As a server admin, I thought the jetpack hydrogen tank was waaay too large, so I went to the config to reduce it... but there wasn't a config option!  So, I added one.

(I haven't tested this because I can't get the mod to build, but it's a really simple edit)


Canisaur:
Ok, I got the mod to build now that I'm not derping gradle.  It worked fine, it even took my existing jetpack immediately down to the new max as soon as I used it.

aidancbrady:
Just implemented this alongside with support for the Flamethrower and Scuba Tank, thanks!

PR #2780, opened 2015-06-30
Maintain java 6 source compatability in API classes

tterrag1098:


unpairedbracket:
Oh, I didn't realise these things  weren't in Java 6. Do we reliably have access to com.google.common? It's a library used by vanilla MC, I presume?

tterrag1098:
Yes.

aidancbrady:
Would you mind resubmitting this to the development branch?
Thanks.

halvors:
Why would you keep Java 6 compatibility? Most people are already using Java 8. The people still using Java 6 should simply update...

aidancbrady:
It's a bigger problem than you may think - a study by Plumbr earlier this year showed 20% of Java users still fixed on JDK/JRE 6.  As horrible as it is, I'm willing to make this small compromise to allow a bit of backwards-compat

tterrag1098:
@halvors I have no problem with the mod itself forcing j7 or j8 (well maybe a little problem) but the API should remain on java 6 as long as minecraft itself does, due to the fact that other modders use it in their environments (such as myself) and have their source compliance set to j6, causing compile errors.

@aidancbrady I will, one moment

tterrag1098:
@aidancbrady Also you can use the diamond operator and still compile to java 7, this doesn't have anything to do with backwards compat

aidancbrady:
By backwards compat I'm talking about maintaining legacy Java 6 support - the diamond was added in 7 and doesn't compile with 6 (which I'm assuming was what this PR was for).

Also thanks for switching to dev ^^

tterrag1098:
Nope that's not how it works. You can compile diamond operators to j6
bytecode, it's just syntax sugar.

aidancbrady:
My old workspace errored at me when I tried to use the diamond in a JDK6-built Eclipse workspace - I was under the assumption you need JDK 7 to work with the version's syntax enhancements.  I'm pretty sure I can't build any projects I use lambdas in w/o JDK 8, for example.  Correct me if I'm wrong, I should know this~

tterrag1098:
It depends on the feature. There is also a difference between source
compliance and compilation target. You can specify with javac (in your
gradle script) what jre to target.

halvors:
@aidancbrady 
sourceCompatibility = "1.6"
targetCompatibility = "7"

tterrag1098:
@halvors The opposite, no? He wants source compatibility 7, and compilation target 6.

PR #2789, opened 2015-07-01
Added the default transparent greyed out background to GuiScreen's.

halvors:
Added the default transparent greyed out background to the
GuiPasswordEnter and GuiPasswordModify GuiScreen's.


halvors:
@aidancbrady Did you understand what i meant here? Since the GuiScreen is the base class of GuiContainer which is used for every other GUI than GuiPassword* the GuiContainer class has this greyed out background, but this is because of this missing from the GuiPassword* classes that derive directly from GuiScreen.

aidancbrady:
Great, thanks.

PR #2838, opened 2015-07-23
Register transmitter items with AE2 P2P

ruifung:
So the respective transmitters can be used to attune a AE2 P2P bus.

For Energy P2P, if IC2 is installed, it uses IC2 - Otherwise it uses RF instead.
Since it probably wouldn't work to register both RF and EU P2P tunnels.

Please tell me if I'm missing anything on the mekanism side.


aidancbrady:
Nice implementation. Thanks!

PR #2856, opened 2015-07-26
Another bugfix.

### LINKED ISSUE #2844: Crash at startup after upgrade
http://paste2.org/jvhK5wKn

ruifung:
Fixes #2844.

Also, @aidancbrady you probably should release a 8.1.7 soon with the bugfixes. This is the second bug that crashes minecraft in singleplayer (bin bug was first), and will pretty much crash any server which uses logistical sorters on anything that uses sided inventories.


aidancbrady:
Thanks!

PR #2950, opened 2015-09-06
Force the player's experience to sync when changing dimensions

### LINKED ISSUE #2943: Xp temporarily dissappears when using teleporter
When using the teleporter to transfer between dimensions, my xp bar temporarily dissapears. The xp isn't lost but as far as enchanting tables and other items that use xp, I might as well be a fresh lvl 0. Relogging or killing something causes the xp to come back to where it was. At first i thought it was something to do with RFtools dimensions, but it even happened when teleporting to the nether. Pics below show what happens.

--Before using the portal to the nether--

-- After arriving at the nether, xp is gone--

-- after using a bottle of enchanting my xp is back --

shadowfacts:
Thank you, @Gebatzens

Fixes #2943

Also add misc IDEA files to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@ output
 .gradle
 .idea
 *.iml
+*.ipr
+*.iws
+out
 
 Mekanism*.jar
 MDK.zip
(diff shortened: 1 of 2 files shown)


aidancbrady:
Thanks!

PR #3014, opened 2015-11-06
Fix for CC peripherals always throwing error.

timstableford:
At least under a Linux and Windows environment using Java 7 or 8 when doing a call from ComputerCraft such as:
peripheral.call("right", "getStored") on an energy cube the call would always return "Error.". This addresses that bug.


aidancbrady:
Thanks!

PR #3028, opened 2015-12-04
A little fix for Electric Pump

exploser:
Fix Electric Pump not pumping heavy water when it's filled with over9000 mb. That is in no way an important fix, but it is a bit annoying that the pump just stops on 9k.

### DIFF
--- a/src/main/java/mekanism/common/tile/TileEntityElectricPump.java
+++ b/src/main/java/mekanism/common/tile/TileEntityElectricPump.java
@@ -38,6 +38,9 @@ public class TileEntityElectricPump extends TileEntityElectricBlock implements I
 	
 	public TileComponentUpgrade upgradeComponent = new TileComponentUpgrade(this, 3);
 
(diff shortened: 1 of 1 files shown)


src/main/java/mekanism/common/tile/TileEntityElectricPump.java
```diff
@@ -119,7 +122,12 @@ else if(tempStack.isItemEqual(inventory[1]) && tempStack.getMaxStackSize() > inv
 		{
 			if(MekanismUtils.canFunction(this))
 			{
-				if(getEnergy() >= usage.electricPumpUsage && (fluidTank.getFluid() == null || fluidTank.getFluid().amount + FluidContainerRegistry.BUCKET_VOLUME <= fluidTank.getCapacity()))
```
justin2tek: Het

dobegor:
@aidancbrady why was this PR closed?

Issue #2832, opened 2015-07-21
Mek and Forestry


Twigflinger:
Just wondered if Mekanism is supposed to be able to use Forestry saplings.  I was hoping to run extra saplings from tree farms through the crusher and eventually produce plastics.  Something I pulled up on google led me to believe Forestry saplings worked at one point in time.  I am running Mekanism 8.1.5.246, Forestry 3.6.2.19 and Forge 10.13.4.1448

othermusketeer:
I am also noticing this as well. I even tried using modtweaker to re-add the recipes. With and without modtweaker, forestry oak saplings show up in crusher recipes, but nothing happens when you place any forestry saplings in. MinefactoryReloaded and IC2:Experimental trees work fine. I have some test info of when it was working and not working in a test enviroment. If you want additional information or tests, let me know.

### Test enviroment

Forge Mod Loader version 7.99.30.1492 for Minecraft 1.7.10
Java is version 1.7.0_71 64-Bit, running on Windows 7:amd64:6.1

### Supporting Mods

buildcraft-7.0.14
buildcraft-compat-7.0.8
CodeChickenCore-1.7.10-1.0.7.46-universal
CoFHCore-[1.7.10]3.0.3-303
CoFHLib-[1.7.10]1.0.3-175
forestry_1.7.10-3.6.3.20
<Mekanism, MekanismGenerators, and MekanismTools. Various versions>

### Testing Method

In creative mode, in a fresh local world, place a creative energy cube. Then place a crusher. Open the crusher GUI (confirming power). Use creative GUI to give yourself a Forestry Oak sapling. Place it in the input slot of the crusher GUI and observe if the crusher begins to operate.

### Results

Mek version | Working
------------ | -------------
7.1.1.127 | Yes
7.1.2.185 | Yes
8.0.0.526 | No
8.1.1.225 | No
8.1.7.252 | No

### Thoughts
Forestry saplings all share the same id and metadata, and only differ by the _HUGE_ block of NBT data for each of them. Could this be the problem? The Buildcraft autocrafter thinks an unanalyzed sapling is completely different than an analyzed one, despite it being the same id and meta.

#### Crude workaround
Use the following in a MineTweaker script to allow converting forestry saplings into plain Oak saplings in a crafting grid or autocrafter.
`recipes.addShapeless(<minecraft:sapling>,[<Forestry:sapling:0>]);`

#### Possibly related
 #2787   #2500

aidancbrady:
[...]

PR #3031, opened 2015-12-06
Allow Forestry saplings to work in crusher

TehStoneMan:
Adds check for ore dictionary entry "treeSapling" and use different
comparison criteria if found.
Also rearranged test conditions to take advantage of Java conditional
short-circuiting.

Addresses issue #2832

### DIFF
--- a/src/main/java/mekanism/api/util/StackUtils.java
+++ b/src/main/java/mekanism/api/util/StackUtils.java
@@ -3,6 +3,8 @@ package mekanism.api.util;
 import java.util.ArrayList;
 import java.util.List;
 
+import mekanism.common.Mekanism;
+import mekanism.common.util.MekanismUtils;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraftforge.oredict.OreDictionary;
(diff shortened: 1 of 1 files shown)


dobegor:
I think this should be rather implemented by server owner / client using [ModTweaker addon for MineTweaker](minetweaker3.powerofbytes.com/wiki/ModTweaker:Mekanism_Support).

XFactHD:
No, not everyone is familiar with MineTweaker/ModTweaker and it doesn't create any other problems as it is independent from any mod APIs.

dobegor:
But one may not want this to be in-mod and it would be a bit more complicated to remove a recipe with ModTweaker instead of adding this. But that's IMHO so to be chosen by @aidancbrady.

XFactHD:
Why wouldn't you want it in the mod? Saplings are saplings meaning that they should all be treated the same.

TehStoneMan:
Technically, the recipe to use Forestry saplings was already there, it was just the extra NBT data they have that caused the machine to not recognize them as valid.

aidancbrady:
Fix implemented, thanks so much!

PR #3040, opened 2015-12-15
Fix, Logistical Sorter ItemStack filter is not ignoring the item damage.

emris:
The filter will only accept items with the same damage that was used to
make the ItemStack filter.

### DIFF
--- a/src/main/java/mekanism/api/util/StackUtils.java
+++ b/src/main/java/mekanism/api/util/StackUtils.java
@@ -62,6 +62,15 @@ public final class StackUtils
 		return wild.getItem() == check.getItem() && (wild.getItemDamage() == OreDictionary.WILDCARD_VALUE || wild.getItemDamage() == check.getItemDamage());
 	}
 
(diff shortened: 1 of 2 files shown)


aidancbrady:
I'm sorry, but I'm afraid the issue you speak of is intentional behavior. If item damage was not taken into account all Mekanism dusts would be considered equivalent.

emris:
If this is intentional behavior, then the Logistical Sorter is useless when it comes to all other items that use the damage value for item damage and not for different items. Why not add a filter for Mekanism dusts, and other Mekanism items that need it.

XFactHD:
Mekanism is not the only mod that uses item damage to distinguish multiple items from each other. If at all, it should be a setting in the GUI for the user to choose.

emris:
Yes, GUI option was how i started this PR, but it turned out to be a lot of new code, maybe too much code for someone who is new to this project, so i went with the simple fix .. but sometimes a simple fix is not the needed solution :)

PR #3049, opened 2015-12-27
Port IPeripheral usage to IComputerIntegration

habnabit:
>Some classes were still using IPeripheral directly, when there was
>already an internal abstraction for integrating with computer systems.

>Added bonus: fixes a bug with reactor logic reactors expecting to get an
>Integer from computer integration, when none of the computer mods
>provide Integer objects (only Doubles).

I'm not familiar enough with the implementation of any of the computer mods to say *why* but this was required to get mekanism-specific methods available for induction casings and fusion reactors. All other components (i.e. the ones using `IComputerIntegration`) were correctly exposing their methods.


aidancbrady:
These changes have already been fixed in development, but thank you for that extra fix, didn't catch that :)

habnabit:
Unpushed commits...?

aidancbrady:
Development branch. Hoping to push a release soon

habnabit:
Oh! I didn't see that branch at all. Thanks!

PR #3087, opened 2016-02-03
DON'T DO THIS, IT IS CURRENTLY USELESS

Stephen789Git:
I'm fairly certain that this check will stop a world unloading glitch triggered by turning off ore generation in some other mods

wait, that makes no sense
what's going on here?


XFactHD:
This value can't be null, it is a fixed value in the ore dictionary class.

Stephen789Git:
There seems to be a problem with the map that stores the ore values...
whatever it is, it causes a crash when you launch with deactivated ore generation

PR #3109, opened 2016-02-21
Add Danish locale

LFalch:
Please tell me, if something is off.


LFalch:
I've tested it now, and couldn't find any problems: Should be safe to merge now.

PR #3151, opened 2016-03-27
Update pt_BR lang

EzerArch:
Full Brazilian Portuguese translation of Mekanism. Adds a ton of new entries, syncs with en_US, translates a handful of left-overs and general spruce-up.

European Portuguese to come (has to be created).


aidancbrady:
Thank you!

PR #3155, opened 2016-03-28
Update fr_FR.lang for v9

HanatakeYurii:
Hi, I tried to update the translation as far as I found differences. Feel free to tell me or correct yourself if I missed any change or made a mistake.


IcieFlake:
Thats it?
Sorry im offtopic but if there is no one to do Swedish and Polish translation then i guess that i'll sign up for it ^^, If there is any need or interest in this.

PR #3186, opened 2016-04-25
Peripherals api includes invalid indexes for methods with string parameters

Buildstarted:
The method indexes (6 and 7) correspond to "removeOreFilter" and "reset" instead of "addOreFilter" and "removeOreFilter".

Fixed the two method indexes listed to the correct ones.


aidancbrady:
Thanks!

Buildstarted:
Still busted anyway. Going to have to debug this to figure out why the api doesn't completely work. Hopefully I'll have another PR soon. (also can't figure out why turtle's destroy the miner completely)

PR #3209, opened 2016-05-11
Fake Player location change

DraksterAU:
This PR should ensure the fake player that breaks the blocks for a digital miner actually reports its location as the digital miner, not the location of the block being broken.

See https://github.com/BuildCraft/BuildCraft/issues/3323 for more information - BuildCraft have implemented this.

@aidancbrady If this is wrong, I apologise. Pretty new to java :)


aidancbrady:
Added to development branch, credit given in commit. Thanks :)

PR #3211, opened 2016-05-12
Fixed TileEntityBin Packet Security Problem

calvinbalke13:
Added network side checking code for TileEntityBin.  There may be other
packets which also have less severe security problems.


aidancbrady:
I've gone ahead and completed this around the board. Thanks for letting me know

PR #3278, opened 2016-05-27
1.8.9 fixes

miuirussia:
And maybe fixes for 1.9. See commits for changelog


aidancbrady:
Thanks! As 1.8.9 is not going to be supported, I'll merge whatever fixes you (or anyone else) finds as long as they're appropriate.

PR #3340, opened 2016-06-09
Update pt_BR.lang

AlterAkima:
Missing tooltips added

### DIFF
--- a/src/main/resources/assets/mekanism/lang/pt_BR.lang
+++ b/src/main/resources/assets/mekanism/lang/pt_BR.lang
@@ -11,6 +11,7 @@ item.EnergyTablet.name=Tablete de Energia
 item.SpeedUpgrade.name=Upgrade de Velocidade
 item.EnergyUpgrade.name=Upgrade de Energia
 item.FilterUpgrade.name=Upgrade de Filtro
+item.MufflingUpgrade.name=Upgrade Silenciador
 item.GasUpgrade.name=Upgrade de Gás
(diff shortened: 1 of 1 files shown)


EzerArch:
Pretty spot-on, you got some entries I'd overlooked, specially the "Lingote"x"Barra" one. Thank you.

Just a few notes:

1. "Auto-ejetar" no longer has hyphen according to the new orthography rules for vowel-ended prefixes.

2. `gui.on=Ligado` and `gui.off=Desligado`: I used "SIM" and "NãO" because longer words didn't fit in the Digital Miner UI.
 - As to "NãO": Minecraft doesn't recognize the "Ã" character. :S

3. `gui.digitalMiner.silk=Toque Suave`: also because of the Digital Miner UI. Even in English it doesn't fit, so it's just "Silk" rather than "Silk Touch". "T.S." or "Toque" can be good choices.

4. "Quantum Entangloporter" is indeed a tricky one. The machine uses [Quantum entanglement](https://pt.wikipedia.org/wiki/Entrela%C3%A7amento_qu%C3%A2ntico) to perform [Quantum teleportation](https://pt.wikipedia.org/wiki/Teletransporte_qu%C3%A2ntico). My suggestions are pretty ugly: "Entrelaceportador Quântico", "Tele-Entrelaçador Quântico" and "Transentrelaçador Quântico".

AlterAkima:
I'll hold until tomorrow to fix my oversights.

PR #3393, opened 2016-06-22
Update en_US.lang to include Lapis Lazuli

onVoid:


aidancbrady:
No clue how that went missing, thanks

PR #3412, opened 2016-06-27
Update to 1.10

miuirussia:
Updated to Minecraft 1.10


mimikal:
Mekanism is for 1.9. Also, there is no Forge for 1.10 yet. So basically you
have to use 1.9.

miuirussia:
http://files.minecraftforge.net/maven/net/minecraftforge/forge/index_1.10.html

mimikal:
Oh, it probably only came out a week ago or something then. Either way,
Mekanism 9 only works on Forge 1.9.

aidancbrady:
I've ported your changes to the 1.10 branch. Appreciate it :)

PR #3492, opened 2016-07-16
Opaque

chaos234:
Hi @aidancbrady ,

I have added the opaque option but for this it was necessary to load the config file before initialising the tiers otherwise it will end up in a crash from preInit to Ini Event during the loading phases.

Tests worked fine for me and FPS won't be disrupted if there are many pipes with rendered content :).

I hope it will be merged for the next version.


chaos234:
I must close this because smth. went wrong. Please refresh your development branch because it is not uptodate with the 1.9.4 branch.

PR #3497, opened 2016-07-19
Digital Miner Alternate Operation

TehStoneMan:
This offers an alternative operating boundary shape to the digital miner. The change makes the miner start with a dome above the miner with the size determined by the radius setting, then continues with a cylinder down to the bottom of the map. This can be seen through the visualizer. Also with this, the max/min y setting is ignores. These changes can be turned on or off through the "MinerAltOperation" config option.


aidancbrady:
Seems this was automatically closed due to the renaming of the branch.  Current working branch is 1.10, feel free to update and reopen :)

PR #3090, opened 2016-02-05
Water to heavy water

chaos234:
Added a secound way to aquire Deuterium because the current functionallity of the pump produces server lags and also it eats up all the water so that it will dissapear and must be replaced with new water.

The process is desinged like in issue #2988 but not over a new SEP (thought as GSP) to work like the Girdler-Sulfid-Process. It is desinged to work with the C. Infuser, C. Washer and Electrolytic Seperator to get enriched and finally to get Deuterium.

For detailed informations look at the image below and feel free to ask me, if there are some questions left.

Image:
newdeuteriumprocessing

Edit 1:

If there a false tab widht let me know this so that I can correct it in my IntelliJ IDEA.

Issue #2988, opened 2015-10-11
[Suggestion] for making Heavy  Water


gp009:
as it currently is only mekanism pumps can make heavy water with a filter upgrade installed. The problem with current with the way mekanism pumps work is you have to have like 100 of those to filter heavy water fast enough to make deuterium fast enough to keep up with the fusion reactor and since they currently consume source water blocks, they create MASSIVE chunk update lag my suggestion is to either:

add to configuration file to toggle on/off water source block consumption ( which will help stop some lag and probly much easier to implement)

or

create a new mekanism machine that is specifically made to filter heavy water by just taking in water that is pumped into pipes from any pump from any mod and just filter the heavy water-- which is probly harder to implement but would work just as well and would allow the use of other pumps from other mods.

phantamanta44:
In real life, one part of heavy water will occur in 3600 parts of normal water, so it's expected that you would need a lot of pumps to get any decent output.

gp009:
true, but they point i was that when you got over 100 pumps consuming water source blocks, this creates a massive block update lag.  the suggestions was to help with lags and do the job. i could have 100 buildcraft pumps bringing alot of water and not consuming water source blocks causing alot less lag even though buildcraft pumps.  

since only mekanism pumps can make heavy water, having 100 of them will create alot of lag, the suggestions i made are to target the and eliminate as much lag as possible

the first suggestion was add an option in mekanism configuration file to toggle on/off the consumption of source blocks leading to fewer block updates and a lot less lag.  (most likely to happen which is fine)

the second suggestion was to remove pumps ability to filter heavy water and just pump normal water and create a machine that takes the water and filter the heavy water--this allow the use of pumps from other mods to pump water into the machine to make heavy water.

I hope im not confusing anyone here, my suggestions just target the block/chunk update lag mekanism pumps cause

speakerrob:
@gp009 I'm tracking what you're saying 100% of the way. I think this needs to be addressed at some point. 

I do love the fact that we now have a reason to use the Mekanism pump vs. other mods' options for this task; however, the lag that it creates is significant in some cases. I've also noticed that whenever the chunk reloads all of my water source blocks disappear and need to be replaced.

[...]

PR #3575, opened 2016-07-31
Mekanism: Secound way for Heavy Water and Deuterium Production

chaos234:
Refering to #3090 and #2988 I have created/implemented what I have had written.

@aidancbrady wanted this as a separat module but this wasn't possible so I included it in the main source.

THe only problem, which is currently still there is an invisbale Rotary Condensentrator Recipe, which will crash your client and I don't know why but I hope the error can be found.

Here is the Image how my solution works:

heawywatersnd

I hope that the written code is ok, I checked it more than two times and tested the hole process which works fine for me.


aidancbrady:
I'm very sorry, but this is too major a content addition to be added to Mekanism on its own, which is why I suggested creating a third-party module.  I'm happy to help if you're having problems.

chaos234:
Then let us have talk about it in Skype or don't force close it @aidancbrady :)

PR #3643, opened 2016-08-13
Added Occlusion Test for Pipes and Wires

### LINKED ISSUE #3640: Cables ignore other multiparts
The Cables and Pipes in Mekanism don't make a Occlusion Test so they ignore if there is a multipart that blocks the connection. 

EDIT: In this case the stone cover is made with Chisels & Bits

feldim2425:
Make it possible to block connections with Chisels & Bits and other multiparts.
Fix #3640


aidancbrady:
Have you tested this to work properly?

feldim2425:
With covers:

If I remove the covers without replacing the cables / pipes


I also tested it with energy and logistical pipes. They will not connect, automatically reconnect, and they don't transport anything until the cover gets removed.

aidancbrady:
Great work, thanks!

PR #3644, opened 2016-08-13
Allow metal blocks to be used for beacons

jmongeon:
Allow metal blocks, refined glowstone blocks, and refined obsidian blocks to be used for beacons.


aidancbrady:
Awesome PR, thanks!

Issue #3663, opened 2016-08-16
Thermal Evaporation Block Cannot Be Cast To Tile Entity Multiblock


Jengo13:
Hallo! My server has been experiencing this crash as of late. Before hand, the sever was running fine, and no major changes have been made to the server recently. Any help regarding this issue is greatly appreciated c:

---- Minecraft Crash Report ----
// This is a token for 1 free hug. Redeem at your nearest Mojangsta: [~~HUG~~]

Time: 17/08/16 00:36
Description: Exception ticking world

java.lang.ClassCastException: mekanism.common.tile.TileEntityThermalEvaporationBlock cannot be cast to mekanism.common.tile.TileEntityMultiblock
	at mekanism.common.block.BlockBasic.canCreatureSpawn(BlockBasic.java:430)
	at net.minecraft.world.SpawnerAnimals.func_77190_a(SpawnerAnimals.java:261)
	at net.minecraft.world.SpawnerAnimals.func_77192_a(SpawnerAnimals.java:168)
	at net.minecraft.world.WorldServer.func_72835_b(WorldServer.java:257)
	at net.minecraft.server.MinecraftServer.func_71190_q(MinecraftServer.java:963)
	at net.minecraft.server.dedicated.DedicatedServer.func_71190_q(DedicatedServer.java:432)
	at net.minecraft.server.MinecraftServer.func_71217_p(MinecraftServer.java:841)
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:693)
	at java.lang.Thread.run(Thread.java:745)

A detailed walkthrough of the error, its code path and all known details is as follows:
---------------------------------------------------------------------------------------

-- Head --
Stacktrace:
	at mekanism.common.block.BlockBasic.canCreatureSpawn(BlockBasic.java:430)
	at net.minecraft.world.SpawnerAnimals.func_77190_a(SpawnerAnimals.java:261)
	at net.minecraft.world.SpawnerAnimals.func_77192_a(SpawnerAnimals.java:168)
	at net.minecraft.world.WorldServer.func_72835_b(WorldServer.java:257)

-- Affected level --
Details:
	Level name: world
[...]

PR #3674, opened 2016-08-19
Fixed rare NPC spawning glitch

TitaniumSapphire:
This should fix issue #3663, which occurred when a creature tries to spawn on a Solar Evaporation block and the block is casted to a type it isn't.

### DIFF
--- a/src/main/java/mekanism/common/block/BlockBasic.java
+++ b/src/main/java/mekanism/common/block/BlockBasic.java
@@ -370,23 +370,28 @@ public abstract class BlockBasic extends Block implements ICTMBlock
 					case 2:
 					case 7:
 					case 8:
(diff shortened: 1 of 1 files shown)


src/main/java/mekanism/common/block/BlockBasic.java
```diff
@@ -370,23 +370,28 @@ public boolean canCreatureSpawn(IBlockState state, IBlockAccess world, BlockPos
...
+								} 
+								else {
+									if(tileEntity.clientHasStructure)
+									{
+										return false;
+									}
 								}
 							}
+						} catch(ClassCastException cce) {
+							System.err.println("Creature attempted to spawn in an invalid location");
```
ThatGamerBlue: Should be discarded silently

PR #3699, opened 2016-08-27
Re-implements https://github.com/aidancbrady/Mekanism/pull/3209

DraksterAU:
This was added previously, I believe, but then never put into Mekanism v9.1

Note, if you no longer support v1.7.10 then nevermind..... I will create a custom build for this and share it - because this only benefits Cauldron/Thermos users which use Bukkit based protections (Factions, Towny, WorldGuard etc).

We implemented the latest Mek and BC and the tweak to allow this the machines to behave, found BC working (and after some checking) I realised Mek doesn't have the fix :(


aidancbrady:
Very sorry, but 1.7.10 is no longer supported. Feel free to create a custom port if you wish.

DraksterAU:
Damn. Cauldron no longer gets Mekanism haha :(

A year from now sponge will a useful replacement for people running cauldron ( and will probably rely on this to) but hey... that's something for future ref.

PR #3788, opened 2016-09-11
More dynamic tank network optimization

tossaponk:
Currently, the dynamic tank will send update packet if stored fluid amount is changed even if the changed amount is very little and has no visual change(10000000->9999990) and wasting the server bandwidth. 
Instead of send change every time fluid amount changed. This edit will only send update packet when fluid level is noticeably changed saving a lot of bandwidth usage if the tank is actively getting filled and drained every tick.


aidancbrady:
Thanks!

Issue #3869, opened 2016-10-04
[1.10.2] Crash when looking at a Mekanism Robit


b0bst3r:
http://pastebin.com/kYNw0EpE

Mekanism-1.10.2-9.2.0.292
theoneprobe-1.10-1.1.2
topaddons-1.10.2-0.7.0

Also note that a similar thing happens with Waila see https://github.com/b0bst3r/Superstitions/issues/2

https://github.com/McJty/TheOneProbe/issues/60#event-812028269

PR #3890, opened 2016-10-15
Fix NPE when homeLocation hasn't yet been set when saving NBT (probab…

### LINKED ISSUE #3669: Robit Crash upon placing on chargepad
[crash-2016-08-18_10.34.20-client.txt](https://github.com/aidancbrady/Mekanism/files/425301/crash-2016-08-18_10.34.20-client.txt)

Not sure if you can see the crash report but I am getting this everytime I try to place a Robit on a Chargepad to Initialize it. Pretty sure the crash is caused by Entity's Vehicle: ~~ERROR~~ NullPointerException: null. I am using the latest release of Mekanism. (9.1.1.290 for 1.10.2)

thiakil:
…ly being called by waila before it is stored). Should fix #3669 & #3869


aidancbrady:
Done, thanks!

PR #3994, opened 2016-11-21
Remove deploader, depend on MCMP fixes #3991

### LINKED ISSUE #3991: Stop using the CCC DepLoader!
The CCC DepLoader is being phased out, You should never need to download any normal mod via these methods. MCMP is a normal mod, just depend on it via the @Mod annotation instead of downloading it. With ANY version of CCC for 1.10 it WILL Complain to users about your use of the dependencies.info file.

Btw, im not just some random user, im the new official maintainer of CB's suff.

thommy101:
As stated in issue #3991 support for deploader is being deprecated.
So, I removed deploader, all references to it, and added a required-after on MCMultiPart.


thommy101:
Tested it with and without MCMultiPart installed.

Without it gives the forge missing mod page on startup.
With starts and runs as supposed.

aidancbrady:
Thanks!

PR #4002, opened 2016-11-22
1.10 wrench

HyperBlueCore:
Can you please add wrenches for long breaking items or blocks such as universal cables because buildcraft is not yet updated and thermal expansion is most likley abandoned


PJB3005:
Wait, what?

is this a pull request with the purpose of an issue?

Mekanism already HAS a wrench. It's called the configurator.

PR #4024, opened 2016-11-29
Update ShapedMekanismRecipe.java

cookta2012:
This should fix a bug where jei shows some items require 2 items in one slot


aidancbrady:
This change would cause recipe parsing to fail.  Note that Forge's ShapedOreRecipe uses the same process.

PR #4035, opened 2016-12-03
Bin IItemHandler

thiakil:
Adds support for ItemHandler capability, exposing the stored items as a 1 slot inventory.

Like the DSU interface I left it as non-sided, and added simulate parameters to the add/remove functions, in order to not duplicate functionality.

Subclass was used because getStackInSlot works differently.

Benefits: things like AE2 & Refined Storage should be able to work with bins without any external dependencies.


PJB3005:
Oh my God thank you! 

I wanted to use bins to mass store ingots and such in my RS system but was disappointed that they don't support storage busses.

aidancbrady:
Lovely PR, thank you very much.

PR #4084, opened 2016-12-23
1.10 about lightwegh modpacks

HyperBlueCore:
this is to you adian 1.10 needs a lot of ram to run heavy mods like yours and take a really long time compared to 1.7.10 so its extremely hard to make a mod pack with this mod so can you please decrease the size of ram needed by decreasing non needed objects and clean the code especcialy in 1.10.2


ktole1999:
i have modpack with mek. list is 93mods and runs fine. yes it takes longer to load than a comparable 1.7 pack but it works fine. its not mek using the memory, its forge.

jburrows82:
Oh yay, another stupid pull request meant as a pointless issue. 

FTB has started to include Mekanism in it's packs... So, you must be doing something wrong Blue if your modpack can't handle Mekanism in it. FTB has no problems with Mekanism LOL

ThatGamerBlue:
Not a real PR /close @aidancbrady

PR #4103, opened 2016-12-30
Fix turbine not returning energy capability - Fixes #4097

### LINKED ISSUE #4097: Turbine multiblock crash - Universal cable Push on port
Setting Universal cable to Push configuration on active turbine port with another port in Pull on same cable [causes a crash](http://pastebin.com/x5rEZn3u), same on [relog](http://pastebin.com/jYYJUvMJ). Was going to set it to None for a test with RFTools Power monitor as it was showing odd RF values with 2 active Pull connections on ports.

thommy101:
Turbine declared it has the energy storage and cable outputter capability with hasCapability, but didn't provide them with getCapability.

Closes #4097

### DIFF
--- a/src/main/java/mekanism/generators/common/tile/turbine/TileEntityTurbineValve.java
+++ b/src/main/java/mekanism/generators/common/tile/turbine/TileEntityTurbineValve.java
@@ -440,6 +440,11 @@ public class TileEntityTurbineValve extends TileEntityTurbineCasing implements I
 	{
 		if((!worldObj.isRemote && structure != null) || (worldObj.isRemote && clientHasStructure))
 		{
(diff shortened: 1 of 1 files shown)


ThatGamerBlue:
a better fix for this would be whenever something requests a capability it should also check if said capability returned is null - patch one fix all

thommy101:
Get capability should never return null, you have to check with has capability is a tile entity has the capability you're looking for. In this case there was a mismatch between has and get capability.

ThatGamerBlue:
except when it doesnt because people cant code /shrug i dont really know

thommy101:
Was just updating my comment ;)

ThatGamerBlue:
i dont know much about this fancy capability system, other than its confusing the heck outta me :P

aidancbrady:
Thank you!

PR #4128, opened 2017-01-06
Fix wrong fence color - Fixes #3954

### LINKED ISSUE #3954: [9.2.0] Plastic Blocks: Colour issue
ModPack: HermitPack 1.2.1
Mekanism Version: 1.10.2-9.2.0.292

I have made several purple plastic blocks, and when I place these down, they are fine - they look like purple plastic blocks.

Both the One Probe (and helmet), and the WAILA mod, however, report them as being Lime Plastic Blocks when I hover over them, and when I break the purple plastic block, it drops a lime plastic block.

If I then place the Lime plastic blocks, the probe and WAILA say it's a purple plastic block, and breaking it gives me a purple one again.

The same happens with other colours, such as Red <=> Orange or Blue <=> Yellow.

thommy101:
Missed a color meta state with the plastic fence.

### DIFF
--- a/src/main/java/mekanism/common/block/BlockPlasticFence.java
+++ b/src/main/java/mekanism/common/block/BlockPlasticFence.java
@@ -38,7 +38,7 @@ public class BlockPlasticFence extends BlockFence
 	@Override
 	public IBlockState getStateFromMeta(int meta)
 	{
-		return this.getDefaultState().withProperty(colorProperty, EnumDyeColor.byMetadata(meta));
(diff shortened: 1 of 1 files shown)


thommy101:
Uhm.. It still has an issue I noticed, closing till I have time to push another fix.

thommy101:
Fixed the second missing state, forgot to push it. Fences do now have the correct colour on placement.

aidancbrady:
Thanks a bunch!

PR #4139, opened 2017-01-12
fix player kicked for flying with jetpack

ThatGamerBlue:
sorry about all the refactoring intellij likes to do it without me allowing it to

### DIFF
--- a/src/main/java/mekanism/common/CommonEventHandler.java
+++ b/src/main/java/mekanism/common/CommonEventHandler.java
@@ -10,23 +10,14 @@ import net.minecraftforge.fml.relauncher.Side;
  */
 public class CommonEventHandler {
 
-    private boolean canFly = false;
-
     @EventHandler
     public void onPlayerTick(TickEvent.PlayerTickEvent event){
         if(event.phase == TickEvent.Phase.START && event.side == Side.SERVER){
-            if(canFly){
-                if(!(event.player.inventory.armorInventory[1].getItem() instanceof ItemJetpack)){
-                    canFly = false;
-                    event.player.capabilities.allowFlying = canFly;
-                }
+            if(!(event.player.inventory.armorInventory[1].getItem() instanceof ItemJetpack)){
+                event.player.capabilities.allowFlying = false;
             } else {
-                if(event.player.inventory.armorInventory[1].getItem() instanceof ItemJetpack){
-                    canFly = true;
-                    event.player.capabilities.allowFlying = canFly;
-                }
+                event.player.capabilities.allowFlying = true;
             }
         }
     }
-
 }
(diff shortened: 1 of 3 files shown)


thommy101:
Tip: crtl + alt + s -> Editor -> Codestyle -> Java -> Imports -> Class count to use import with '*'. Change that one to a higher value to prevent auto merging of imports.

ThatGamerBlue:
@thommy101 omg thank you <3

src/main/java/mekanism/common/CommonEventHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ * @author ThatGamerBlue
+ */
+public class CommonEventHandler {
+
+    private boolean canFly = false;
+
+    @EventHandler
+    public void onPlayerTick(TickEvent.PlayerTickEvent event){
+        if(event.phase == TickEvent.Phase.START && event.side == Side.SERVER){
+            if(canFly){
```
Bluexin: Seems like you cache a value on the serverside.
One for all players, while this event is called for each player.
I'm highly doubtful about this working properly.

src/main/java/mekanism/common/CommonEventHandler.java
```diff
@@ -0,0 +1,32 @@
...
+                if(!(event.player.inventory.armorInventory[1].getItem() instanceof ItemJetpack)){
+                    canFly = false;
+                    event.player.capabilities.allowFlying = canFly;
```
Bluexin: Basing it solely on mek jetpack would probably break any other mod changing this wouldn't it?

ThatGamerBlue: thats their problem

Bluexin: I do not like that idea, because it'd go against good modding practices.

Tencao:
I second what has been said here. There needs to be stricter checks as to when to allow flying, rather than just checking if the item is in the inventory.  Right now, you're not even checking if the item has enough fuel to even fly, you're just saying if the player has the item. I have several mods in my modpack that control flying to some degree, and having the mentality of "its their problem" is not a fix at all, its laziness.

ThatGamerBlue:
the allowFlight player capability doesnt mean the player can fly, it means if the player is flying by whatever means the player wont get kicked

thommy101:
The way you've implemented it, when the player stops wearing the jetpack, you set the flying capability to false, but what if an other mod has set it to true? That's overwritten with your false at that moment.

Bluexin:
Yeah, but according to some people it's their problem ¯\_(ツ)_/¯

aidancbrady:
Not a big fan of modifying flight capabilities, sorry.

Issue #4143, opened 2017-01-14
Mekanism will not recharge EnderIO items (1.10.2)


Nuinethir:
I have come across a problem where Mekanism (what I use for my power systems) will not charge EnderIO items (such as their Electromagnet). I have tried direct insertion into Energy Cubes as well as standing on the Charging Pad. I have to use either their Capacitor Bank or the Wireless Charger.

Conversely, Mekanism items/blocks can be charged on EnderIO charging methods.

I thought this was strange and opened a ticket with them, though they said the problem was on your side (though Mekanism will charge any items from any mods so far except EnderIO). They said it had something to do with the Tesla API. I then installed the Tesla mod with no effect.

**Link to the issue on their Git:** https://github.com/SleepyTrousers/EnderIO/issues/3992
**Forge:** 1.10.2-forge1.10.2-12.18.3.2202
**Installed Mods:** http://pastebin.com/wZrYQXAx

ThatGamerBlue:
Duplicate of #4116

Nuinethir:
I will follow the other ticket. Thanks for the link. :+1:

Issue #4116, opened 2017-01-02
[1.10.2] Universal Cables


oryfoxer7410:
Issues with power transfer between Clockwork Engine from Forestry with universal cables. This issue shows up with other mods also.

roboderpy:
Confirmed. Universal Cables connects to RF based machinery, but they can't receive any energy from RF storages, only transmit energy to them.
![2017-01-03_10 37 54](https://cloud.githubusercontent.com/assets/6339511/21599548/d48e21da-d1a0-11e6-94a5-7e64e7c75a63.png)
![2017-01-03_10 38 10](https://cloud.githubusercontent.com/assets/6339511/21599547/d48e0434-d1a0-11e6-98e3-fef129ac2608.png)
![2017-01-03_10 38 43](https://cloud.githubusercontent.com/assets/6339511/21599549/d48e9d54-d1a0-11e6-9487-ea86de70c756.png)

ktole1999:
versions/pack used?

roboderpy:
Latest Mekanism version (9.2.1.295), Not a modpack.

roboderpy:
Additional info - Energy Cubes *can* transmit and receive RF energy directly from RF generators, capacitor banks, etc.

roboderpy:
Installed Tesla API. Now i got this:
![2017-01-04_20 27 03](https://cloud.githubusercontent.com/assets/6339511/21643674/35b6a060-d2bc-11e6-9d55-cbcbcbd0af70.png)
![2017-01-04_20 27 31](https://cloud.githubusercontent.com/assets/6339511/21643675/35bb96f6-d2bc-11e6-86fd-869f9980e24b.png)
Universal Cables can't transmit energy to RF machinery, but they now can receive energy from them.

thommy101:
Also took a look at this one and found the culprit. Mekanism doesn't follow the Tesla guidelines regarding to sending and receiving power.

From the standard page of the Tesla github:
Power sources are responsible for pushing power to nearby machines. Machines and wires should NOT pull power from other sources. 

Mekanism cables currently look for energy providers (TeslaProducer capability, IEnergyProvider RF, IEnergySource EU and own CableOutputter capability). It should look for energy consumers and push the energy towards them.
[...]

Issue #3918, opened 2016-10-24
Tesla implementation leads to infinite energy


utoc:
This has been a difficult issue to track down due to the lack of mods in my pack using Tesla, however, now that EnderIO supports it; we've narrowed it down further.

Preliminary case before EnderIO supported Tesla: https://github.com/utoc/BBQBugs/issues/18

The setup: Two tesla capable mods; one of which is Mekanism.  The other mod needs to have energy transport, ie, ActuallyAdditions or EnderIO are what we've tested with.

Using EnderIO item conduits, you effectively multiply energy by 2.5x when going from a Mek cube -> EIO Energy Conduit -> Mek Energy Cube.  Using ActuallyAdditions, you can use their energy relays and get some squirrely results.  EnderIO and ActuallyAdditions, between themselves, behave quite well.  This leads me to believe that something is misbehaving with Mek's Tesla implementation.

If I remove the Tesla jar, the issue goes away completely and Mek plays nicely.

Can you check into this;  I've been all over EnergyAcceptorWrapper.java and can't spot the issue

b0bst3r:
This maybe related to the issue we found:-

So made a quintuple AA battery, set to charge items in inventory/hotbar, is enabled.

Grab up an empty ultimate energy cube from Mekanism and it will fill 51.x million rf in about a second while using up very little energy from battery itself.

Tried with an EIO capacitor bank and didn't have any issues, same with draconic capacitor that was disabled to not charge anything.  It charged both of these items as expected. It's just mekanism energy cubes.

Saereth:
This is also occuring with Telsa energy blacklisted if the Tesla mod is installed

utoc:
This may have been addressed with https://github.com/aidancbrady/Mekanism/pull/4148

Bluexin:
I confirm my PR fixes it.
The issue was mekanism not converting back from Tesla to Joules.

PR #4148, opened 2017-01-17
Fixes various issues related to Tesla API

Bluexin:
See each commit message for details.
#4116 #4143 #3918 (maybe others I didn't find)

### DIFF
--- a/src/main/java/mekanism/common/util/CableUtils.java
+++ b/src/main/java/mekanism/common/util/CableUtils.java
@@ -228,7 +228,7 @@ public final class CableUtils
 		else if(CapabilityUtils.hasCapability(tileEntity, Capabilities.TESLA_CONSUMER_CAPABILITY, side.getOpposite()))
 		{
 			ITeslaConsumer consumer = CapabilityUtils.getCapability(tileEntity, Capabilities.TESLA_CONSUMER_CAPABILITY, side.getOpposite());
-			sent += consumer.givePower((long)Math.round(currentSending*general.TO_TESLA), false);
+			sent += consumer.givePower((long)Math.round(currentSending*general.TO_TESLA), false) * general.FROM_TESLA;
 		}
 		else if(MekanismUtils.useRF() && tileEntity instanceof IEnergyReceiver)
 		{
--- a/src/main/java/mekanism/common/util/CableUtils.java
+++ b/src/main/java/mekanism/common/util/CableUtils.java
@@ -1,9 +1,8 @@
 package mekanism.common.util;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
-
+import cofh.api.energy.IEnergyConnection;
+import cofh.api.energy.IEnergyProvider;
+import cofh.api.energy.IEnergyReceiver;
 import mekanism.api.Coord4D;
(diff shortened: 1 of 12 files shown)


src/main/java/mekanism/common/util/CableUtils.java
```diff
@@ -189,57 +189,30 @@ public static void emit(IEnergyWrapper emitter)
...
-
-					do {
-						double prev = sent;
-						sent += emit_do(emitter, outputtingSides, energyToSend-sent, tryAgain);
-
-						tryAgain = energyToSend-sent > 0 && sent-prev > 0 && i < 100;
-
-						i++;
-					} while(tryAgain);
+					double sent = emit_do(emitter, outputtingSides, energyToSend);
```
Bluexin: Not 100% sure why this was working the way it was, and I might have overlooked something (all of this is done at 5am).
In my tests, this made everything work as expected (and not as it used to), but I obviously can't check everything.
If you think this should stay (in an adapted way to still fix the issue), please leave your feedback.

Bluexin:
Forgot to mention, the cables do not seem capable of receiving energy from EnderIO capacitor banks and cables, because both of these don't reply to the Tesla outputter capability (this was tested both with and without the capacitor bank in output mode, as well as with and without the cable in PULL mode).
They receive energy from EnderIO generators just fine though (without setting the cable in a special mode).

thommy101:
If you're tackling power issues, there is another Tesla related one. I would fix it myself, but I'm currently busy with school and stuff. The current implementation totally ignores the simulate argument of the tesla capability.
Would be great if you could also tackle that.

Bluexin:
From what I've seen, it takes it into account. I didn't check everything though, only what was related to cables (and their network) as well as energy cubes.
Mind to link me specifically to an example?

PS: I'm also kinda busy, with exams and writing multiple mods for a modpack in addition to fixing existing mods :p

thommy101:
Some places I found it are at the takePower and givePower methods of TeslaIntergration. Changing that would require to add the simulated argument to the upstream call.

Bluexin:
I see now where these are.

@ThatGamerBlue I just saw your "rewrite" fork, but didn't investigate it much yet.
Are your plans to completely rewrite the energy parts (which have gotten damn messy over the years, that's for sure) ?
Do these issues apply to your version, and do you want me to see about porting these?
Finally, do you want me to see about helping to clean things up on your fork?

Bluexin:
While in theory these simulation parameters should work, I can not test it right now.
I made sure everything compiled right on this branch etc but did not test in-game.

ThatGamerBlue:
@Bluexin sorry I couldn't get back to you sooner, yes I plan to rewrite everything, currently there are no issues on my fork because it is very basic (just the metal storage blocks and plastics) at this point but I will probably need help in the future :)

roboderpy:
bump

aidancbrady:
Self-merged, thanks

thommy101:
@aidancbrady I'm still missing the simulated argument which causes unwanted extractions and insertions of energy. Some mods do test if other tiles accept energy with a simulated insertion.

roboderpy:
@aidanebrady why do you do this

PR #4239, opened 2017-02-25
Added ability to process Draconium Ore. Resolves #4223

### LINKED ISSUE #4223: [1.10.2-9.2.1.298] - No factory equipped to convert "Draconium Ore" or "Nether Draconium Ore" to dust.
By default, a pickaxe will turn draconium ore to draconium dust.  When using the digital miner with Silk Touch, you get the original ore, but the ores cannot be handled by any Mekanism factories, therefore it has to be taken out of the miner and manually mined into dust, before then being put back in the system to let the smelting factory convert that dust to ingots (which does work).

JaSpr:
Added enrichment chamber recipe for Draconium Ore (from Draconic Evolution) into Draconium Dust.


aidancbrady:
Done, thanks!

Issue #4233, opened 2017-02-22
[1.10.2-9.2.1.298] Recurring client crash while dismantling Thermal Evaporation Plant


JaSpr:
I was dismantling the top layer of my thermal tower (which is 12 or so blocks above the thermal controller) when the game crashed.  It specifically mentions rendering the thermal controller in the crash report.

[Crash report.](https://gist.github.com/JaSpr/440b48405a42f5ab65cda177eeb0793a)
[And another crash report.](https://gist.github.com/JaSpr/5574fa1281ef1b9ae480724d3753ddf3)
[#3](https://gist.github.com/JaSpr/a4d0125eab282bcad759c2082e026ef1)

JaSpr:
[And another crash report.](https://gist.github.com/JaSpr/5574fa1281ef1b9ae480724d3753ddf3)

JaSpr:
[#3](https://gist.github.com/JaSpr/a4d0125eab282bcad759c2082e026ef1)

JaSpr:
I stumbled across what was happening after about the fifth crash:  When you remove an entire level, removing the last block of that level causes the tower to be "complete" again (albeit 1 level shorter).  It was crashing each time that it tried to change the status of the tower from incomplete to complete again.  This *might* be because the contents of the tower (the water) would have exceeded the height of the tower itself. (The previous tower was at 100% capacity).

Each time it crashed, and I logged back in, the water was restored (to a proper level for the new height), and the first block I broke caused the water to disappear (as the tower was then incomplete).  It was only upon removing the last block from that level that it crashed, while trying to say the tower was complete.

JaSpr:
Error occurs on RenderThermalEvaporationController.java@48 when `tileEntity.inputTank.getFluidAmount()` is greater than `tileEntity.inputTank.getCapacity()`.

The result is that amount/capacity is greater than 1, and when multiplied times the length of `displayList`, it results in an index that is greater than the length of `displayList`.  Therefore, when trying to access the value of that index, it causes an out of bounds error.  This can be resolved by forcing a Math.min with (value) versus 1, meaning we'll never try to render the tower more full than 100%

JaSpr:
Fixed in #4240

PR #4240, opened 2017-02-25
Fix for #4233: out-of-bounds error in thermal evaporation render, which crashes client.

JaSpr:
Thermal Evaporation Plant will no longer attempt to render with a fluid-level that is higher than its max capacity.


aidancbrady:
Thanks!

PR #4278, opened 2017-03-28
Use Iterator.remove() instead of quirky workarounds + LinkedList usage

Lignum:
There are some instances of a `toRemove` pattern being used. That is, lists which need to remove elements while they're being iterated over queue up the items to be removed in a separate `toRemove` list to avoid a `ConcurrentModificationException`.

The exception occurs when a list is being manipulated directly instead of via its iterator, meaning we can delete elements by accessing the iterator instead, which is a lot nicer.

Additionally, I've changed some `ArrayList`s that should be `LinkedList`s but aren't `LinkedList`s to be `LinkedList`s. As well as that, I've simplified a couple of other places in the code by using `Collection.addAll`.


src/main/java/mekanism/common/util/CableUtils.java
```diff
@@ -1,8 +1,6 @@
 package mekanism.common.util;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
BrianEdmonds: I'd discourage wildcard imports, but I don't know what Mekanism's style is.

Lignum: I'm not a fan of it either. I've reinstalled my IDE recently and haven't configured it properly yet, so it does these things on its own 😛 . I'll change it if it's a problem.

PR #4282, opened 2017-03-30
Better OpenComputers compat.

marcin212:


src/main/java/mekanism/common/tile/TileEntityGasTank.java
```diff
@@ -421,4 +423,26 @@ public TileComponentSecurity getSecurity()
...
+		switch(method)
+		{
+			case 0:
+				return new Object[] {gasTank.getMaxGas()};
+			case 1:
+				return new Object[] {gasTank.getStored()};
+			case 2:
+				return new Object[] {gasTank.getGas()};
+			default:
+				throw new NoSuchMethodException();
```
thommy101: Maybe a ```return new Object[] {"Unknown command."};``` instead of throwing an error, just as TileEntityReactorLogicAdapter is currently using this?

marcin212: OC handles it correctly and here is similar: https://github.com/aidancbrady/Mekanism/blob/master/src/main/java/mekanism/common/tile/TileEntityLaserAmplifier.java#L364

thommy101: Looked a bit in the code, throwing an exception and returning the object with Unknown Command as text results in the same. I like the exception one better! :)

https://github.com/aidancbrady/Mekanism/blob/master/src/main/java/mekanism/common/integration/CCPeripheral.java#L47

BrianEdmonds: Consistency in code is important, so people don't wonder why it's one way one place, and different somewhere else. If the devs agree with you, I expect they'd be happy to accept a refactoring change after this is merged (as an object return) to change them all to exceptions.

Just my non-dev opinion.

aidancbrady:
Thanks!

Issue #4232, opened 2017-02-22
GasTanks


PandaPaw:
you are taking damage when you walk in to the gas tank if the tank are in head level

BrianEdmonds:
Confirmed in both 300 and 301. I routinely whack my head on these.

marcin212:
Confirmed in 300.

thommy101:
Fixed in 00a107c5990915233b030299580b576534e663f7

PR #4285, opened 2017-03-31
Change gastank to non full cube.

thommy101:
Resolves: #4232 

Currently the gas tank has a small bounding box, but is also a full cube which causes players to suffocate in them when walked into at head level and rubberbanding players when walked at feet level.

This commit changes:
- Gas tanks are non full blocks
- The bounding box aligned to 3/16th (0.1875) off a block

(Can be merged in 1.11 too)


aidancbrady:
Thanks!

PR #4305, opened 2017-04-08
More nullable changes, Fix IITemHandler capability

### LINKED ISSUE #4303: 1.11.2 capabilities crash
Link to my crashlog: 
https://pastebin.com/jTQU4anu

This happens when trying to place an ultimate cable next to another ultimate cable, but missing the grass block they are both on and hitting the previous ultimate cable instead.

I also got this error when placing a hopper on top of a biofuel power generating machine.

edit: I will report back if this happens without the 2.0.0.17 MCMP mod installed and see if the issue lies with that.

This does not seem to happen with MCMP removed.  However, still getting the hopper issue with the following pastebin crashlog:
https://pastebin.com/FqScPCw1

edit: I'll try again with foamfix removed and report back here, forgot that was installed.  Doubt its root cause, but I see that whole "don't report errors with this installed" line. 
edit2: yup same crash with foamfix removed.  Not sure the root cause though, as it doesn't link to a mekanism line.

thommy101:
Fixed some nullable issues with ItemStacks.
Also fixed the IItemHandler capability wrapper, which was using the wrong class (resulting in a constructor not found exception which led to a null capability being returned)

Fixes #4303

### DIFF
--- a/src/api/java/buildcraft/api/blueprints/SchematicBlock.java
+++ b/src/api/java/buildcraft/api/blueprints/SchematicBlock.java
@@ -166,7 +166,7 @@ public class SchematicBlock extends SchematicBlockBase {
             NBTTagList rq = new NBTTagList();
 
             for (ItemStack stack : storedRequirements) {
-                if (stack == null || stack.getItem() == null) throw new IllegalStateException("Found a null requirement! " + getClass());
+                if (stack == ItemStack.EMPTY || stack.getItem() == null) throw new IllegalStateException("Found a null requirement! " + getClass());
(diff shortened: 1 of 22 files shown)


kragnoth:
weird, I don't remember closing this.  Must have been a misclick. :)  Thanks a ton.

aidancbrady:
Self-merged, using isEmpty() instead of ItemStack.EMPTY equivalencies. Really appreciate your help!

thommy101:
:D No problem
Derp.. That method also exists

PR #4312, opened 2017-04-11
Fix cardboardbox tooltip

### LINKED ISSUE #4299: [1.11] Cardboard box tooltip
In 1.11 the cardboard box tooltip always shows "Air" as it's block content.

How to reproduce:
1. Give yourself a cardboard box
2. Pack an item (example obsidian)
3. Break it in survival
4. Look at tooltip
5. Place and unpack item
6. Original item is in the world

Expected output:
Show correct block name on the tooltip.

Possible solution:
Currently there's a ItemStack being made (L57 of ItemBlockCardboardBox) with  block id and meta as arguments. But the constructor of itemstack with accepts block and an int as argument, uses the int as amount.
Changing it to ItemStack(Block blockIn, int amount, int meta) should solve this issue. Thus adding an extra amount argument which is larger than 0, resulting in a valid ItemStack.

thommy101:
Fixes #4299

### DIFF
--- a/src/main/java/mekanism/common/item/ItemBlockCardboardBox.java
+++ b/src/main/java/mekanism/common/item/ItemBlockCardboardBox.java
@@ -54,7 +54,7 @@ public class ItemBlockCardboardBox extends ItemBlock
 		if(data != null)
 		{
 			try {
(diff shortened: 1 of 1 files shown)


thommy101:
Fixed in 4adbc332f960e8a1c82fa5b1d0749e3dbc29c0fd

PR #4313, opened 2017-04-11
Allow axe specific enchantments on axe

### LINKED ISSUE #4122: Mekanism axes can't be enchanted with Sharpness

### LINKED ISSUE #4211: [Feature Request] Processing raw emerald ore (similar to how diamond ore works)
Currently, we can send raw diamond ore through the enrichment chamber to get 2x returns. Emerald ore, however, can't be processed this way, and there's no way to get the emeralds out without taking the blocks, placing them in the world, and manually breaking them with a pickaxe. This can be extremely tedious when you have 4 full stacks of emerald ore from a trip with the digital miner (2 stacks of nether emerald ore through the smelter returns 4 stacks of plain emerald ore).

thommy101:
Fixes #4122

NOTE: target branch is 1.10, can also be cleanly merged in 1.11.

### DIFF
--- a/src/main/java/mekanism/tools/item/ItemMekanismAxe.java
+++ b/src/main/java/mekanism/tools/item/ItemMekanismAxe.java
@@ -1,34 +1,39 @@
 package mekanism.tools.item;
 
+import java.util.List;
 import java.util.Set;
 
+import mekanism.api.util.StackUtils;
+import mekanism.common.Mekanism;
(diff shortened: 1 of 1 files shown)


thommy101:
Whoops... I mentioned the wrong issue here.. #4122 is the right one

aidancbrady:
Thanks!

PR #4320, opened 2017-04-16
Fix IMC recipe removal

thommy101:
IMC messages aren't able to delete any recipe.
Fixed by changing the String to look at.


aidancbrady:
Brilliant

Issue #4342, opened 2017-04-24
[1.11.2] UUID conversion


thommy101:
(Opening an issue for visibility) 

There is an issue with your UUID commit.
I don't know exactly what happend, but my guess is an hover over an fluidtank.
Log: https://gist.github.com/thommy101/ea43c6df2fd4bba594664057a5c8ecbc

Playing on an old 1.11 world.

--Edit--
Shift hover over an fluidtank in my inventory triggered this bug.

aidancbrady:
Ah, looks like merely a conversion issue. Not worried as I consider 1.11 to be a hard update.

thommy101:
In my pull request was a little part about an username to uuid conversion which only runs once. Maybe an idea to use it?

thommy101:
I also think here are some issues with nbt tag conversions. There seems to be changes to the tag to get, but the check if those tags exists are not changed.

https://github.com/aidancbrady/Mekanism/blob/1.11/src/main/java/mekanism/common/tile/component/TileComponentSecurity.java#L156

ruyanve:
I also am having issue with NBT tags on 1.11.2. The mod LevelUp sets a no place nbt tag on ores which makes them not process in mek machines. Didn't know if I should open a new issue since it was mentioned here...

aidancbrady:
All done!

Issue #4345, opened 2017-04-25
[1.11.2] Crash when i want to place a Digital Miner


cyberpunkbln:
Hello,
when i want to place a digital miner its crash. I have the last commits. Ah yeah and all my upgrad modules where deleted in my world, i must cheat me back my machine-upgrades.

`---- Minecraft Crash Report ----

WARNING: coremods are present:
  RebornCoreASM (RebornCore-1.11.2-2.14.0.81-universal.jar)
  CCLCorePlugin (CodeChickenLib-1.11.2-2.6.0.254-universal.jar)
  LittlePatchingLoader (LittleTiles v1.4.8 mc1.11.2.jar)
  CreativePatchingLoader (CreativeCore v1.7.3 mc1.11.2.jar)
  DummyPlugin (ForgeEndertech-1.11.2-2.0.8.0-build.0025.jar)
  BCModPlugin (zz-backpacks 1.11.2 - 3.1.8.jar)
  ShetiPhian-ASM (shetiphiancore-1.11.0-3.4.4.jar)
  DynamicSurroundingsCore (DynamicSurroundings-1.11.2-3.3.3.1.jar)
Contact their authors BEFORE contacting forge

// Would you like a cupcake?

Time: 4/25/17 10:15 PM
Description: Unexpected error

java.lang.IllegalArgumentException: Invalid UUID string: ChaosKaiBln
	at java.util.UUID.fromString(Unknown Source)
	at mekanism.common.item.ItemBlockMachine.getOwnerUUID(ItemBlockMachine.java:876)
	at mekanism.common.item.ItemBlockMachine.placeBlockAt(ItemBlockMachine.java:337)
	at net.minecraft.item.ItemBlock.func_180614_a(ItemBlock.java:51)
	at mekanism.common.item.ItemBlockMachine.func_180614_a(ItemBlockMachine.java:286)
	at net.minecraft.item.ItemStack.func_179546_a(ItemStack.java:166)
	at net.minecraft.client.multiplayer.PlayerControllerMP.func_187099_a(PlayerControllerMP.java:462)
	at PlayerControllerOF.func_187099_a(PlayerControllerOF.java:87)
	at net.minecraft.client.Minecraft.func_147121_ag(Minecraft.java:1521)
	at net.minecraft.client.Minecraft.func_184117_aA(Minecraft.java:2179)
	at net.minecraft.client.Minecraft.func_184118_az(Minecraft.java:1956)
[...]

PR #4346, opened 2017-04-26
Fix more UUID changes. Fix Robit owner.

thommy101:
Fixes: #4342 and #4345.

### DIFF
--- a/src/main/java/mekanism/client/MekanismClient.java
+++ b/src/main/java/mekanism/client/MekanismClient.java
@@ -21,7 +21,7 @@ import net.minecraftforge.common.MinecraftForge;
 
 public class MekanismClient extends Mekanism
 {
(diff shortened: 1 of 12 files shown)


thommy101:
Changed all nbt references from owner to ownerUUID (to prevent crashes) and updated the Robit to have an owner UUID instead of an owner username.

aidancbrady:
The one issue I see with this is that UsernameCache is not synchronized on the client, hence clientUUIDMap in MekanismClient.  I'm gonna do some extra testing to make sure everything works properly.

thommy101:
Oh, didn't know that. Haven't tested the Robit to be honest. All other things are working fine.

aidancbrady:
Yeah, everything else looks great. I'll do some changes to make sure the Robit works properly on the client.

PR #4347, opened 2017-04-27
Add scroll wheel support for GuiScrollList

thiakil:
Allows to use the mouse wheel to scroll in GuiScrollList. Mouse event is accessible on all GuiElements.


aidancbrady:
Lovely PR, thanks a bunch.

thommy101:
Working flawless, awesome! It's much easier this way :) Thanks!

Issue #4336, opened 2017-04-22
[1.11.2] No null facing handling of inventories


thommy101:
Getting the inventory of a null side is a valid operation. The purpose of getting the inventory from the null facing is something like "get the internal inventory". This is used by mods who don't care about facings like The One Probe and RFTools.

~~~This also applies to fluid holders, althrough I didn't test if it is also affected by the same "bug".~~~
Fluid holders aren't affected.

Log:
https://gist.github.com/thommy101/0bd888d4768ceaf564d72789715ccaf5

aidancbrady:
Fixed!

thommy101:
Cannot confirm correct fix. It's still throwing a NPE with the same stacktrace.
Also the purpose of the null facing is showing all inventory, not nothing. Looking at the code of a vanilla furnace, it returns all three slots on the null facing and allowing other mods (like The One Probe) show all the three slots in the heads up display.

Issue #4279, opened 2017-03-28
feature request: better theoneprobe integration


BrianEdmonds:
When I look at a Quantum Entangloporter with the one probe (worn as a bauble), and shift for extra information I get "Error: theoneprobe:default" as the additional information. There are probably other blocks with similar issues, but this was the one that I hit when I was in a bug reporting mood.

Relevant versions: Minecraft 1.10.2, Mekanism 301, theoneprobe 1.4.4, Baubles 1.3.9 (if that matters).

Screenshot
![2017-03-28_14 00 12](https://cloud.githubusercontent.com/assets/18468613/24427180/e6908016-13be-11e7-82c9-1b12540a2069.png)

BrianEdmonds:
I take back the "other blocks" comment. I took a look around all my Mekanism equipment and only the QE displayed this issue. (Could of course affect something I've not built yet, as I'm only up to 4x ore processing, so there's a bunch of the tech tree I've yet to explore.)

thommy101:
Guess this one is solved now:

![TheOneProbe](https://cloud.githubusercontent.com/assets/1036527/26558308/f2cdface-44a7-11e7-8108-5272fe44d6bc.png)

PR #4357, opened 2017-05-06
Check for null side in TileComponentConfig

thiakil:
Trying to insert into null side can cause null pointer in TileComponentConfig via getBaseOrientation. This makes it so the component returns EMPTY if a side isn't given.

Example: place an RFTools builder with configured pump card, Quantum Entangleporter on top. Instant crash.

Crash log from before this fix.
[crash-2017-05-06_13.15.22-server.txt](https://github.com/aidancbrady/Mekanism/files/980725/crash-2017-05-06_13.15.22-server.txt)

### DIFF
--- a/src/main/java/mekanism/common/tile/component/TileComponentConfig.java
+++ b/src/main/java/mekanism/common/tile/component/TileComponentConfig.java
@@ -138,11 +138,20 @@ public class TileComponentConfig implements ITileComponent
 	
 	public SideData getOutput(TransmissionType type, EnumFacing side, EnumFacing facing)
 	{
+		if (side == null)
+		{
+			return EMPTY;
+		}
 		return getOutput(type, MekanismUtils.getBaseOrientation(side, facing));
 	}
 	
 	public SideData getOutput(TransmissionType type, EnumFacing side)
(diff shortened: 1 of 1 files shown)


thommy101:
Should fix #4336 and #4279 

Although, normally a null facing means internal and should show all inventory slots, instead of none.

thiakil:
Also could mean 'self' - or when side is not applicable, like an itemstack.

In my particular case it was a fluid interaction that caused it (where sidedness should definitely have applied), and with mekanism being strictly sided, I think empty is appropriate.

thommy101:
For itemstacks side is applicable, right? Input and output is separated. 
Things like a vanilla furnace had input at the top, fuel from the side and output at the bottom, yet requesting an item capability (which in Mekanism uses getBaseOriantation) with facing null returns the whole inventory. Requesting one with a side gives the appropriate capability for that side.

thiakil:
Applicable to itemstacks in the sense of `ItemStack.getCapability` - was not clear, sorry.

aidancbrady:
Self-merged in 1.11, thank you!

PR #4358, opened 2017-05-06
Register sounds with the GameRegistry

### LINKED ISSUE #3638: [1.10.2-9.1.1.] Logistical sorter sound bug
No idea why, no matter whether Mekanism sounds enabled/disabled, when Logical Sorter is working, it is not emitting 'Click' sound. Instead, it emits 'Eerie noise'.

thiakil:
I couldn't find anywhere where you register your sound effect with the GameRegistry, so I added it to MekanismSounds. This seems to fix #3638 on dedicated servers (no issue on client integrated server)


aidancbrady:
Thanks!

PR #4388, opened 2017-05-28
Updated Russian translation

raspopov:
Synchronized language IDs with en_US.lang.
Added missed translations.
Fixed wrong translations (for example "brine", "slurry" etc.) and long (doesn't fit to GUI) translation.


aidancbrady:
Appreciate it!

PR #4390, opened 2017-05-28
Dependency updates

thommy101:
Updated some (not all) the dependencies to their 1.11 versions and moved them from the API folder to a Gradle dependency.

Mods updated and moved to Gradle:
- Tesla
- Invtweaks
- OpenComputers

Updated WAILA to HWYLA and moved to Gradle. HWYLA is an up to date maintained version of WAILA of TehNut (https://github.com/TehNut/HWYLA)

Removed compat for MineFactoryReloaded and Metallurgy as both do not have a 1.11 version at this moment.
Computercraft itself doesn't have a 1.11 version, but since it's opensourced TheCodedOne has a maintained version with the same API (https://github.com/NoMansMinecraft/Computality)

Haven't looked at:
- IC2, but saw you've already updated it's API
- Buildcraft, but it's certainly available for 1.11
- Galacticraft, also, certainly available for 1.11

Added all changes as individual commits, for the case of wierd stuff a single mod can be reverted to the state it was.


thommy101:
Closed and reopened to trigger a new Travis build. Which is successful now

aidancbrady:
Thanks so much, been meaning to do this for a while.

Issue #4165, opened 2017-01-28
Basic Fluid Tanks + Higher


zerafox:
So being new to Mek, I found myself playing around Pipes and the Tanks (Not the dynamic).
I figured out that I could make the Basic tanks (and higher) "active" which allowed them to 'drain' to tanks below them. Unfornately they still act as seperate tanks which means if I setup a 'tube' of tanks for one liquid, lets say water. And another 'tube' for say lava, the lava can still pour into my 'water tube' because the connected Tanks still act as though they are separate, despite the liquids draining to the lower tank.

What I'm asking for, is making it so that stacked, active tanks can act as one complete tank.
The purpose for this, is to allow 'tubes' of liquids to setup in a manner of 'Water liquid tube' and 'Lava liquid tube' and 'Copper liquid tube' without having one of them wind up in some other tube they weren't meant to be in.

Sadly to achieve this right now I have 1 of 2 options:

1. complete separate pipe units
2. setup complex levers and redstone to disable pipes/tubes I don't want the liquid to enter, but allow into the one I do want to enter.

thommy101:
I've made a pull request implementing this feature, see linked pull request (4400)

PR #4400, opened 2017-05-29
Only allow same fluids in active fluidtank towers.

thommy101:
Implements feature #4165
If the current fluidtank is active and contains no fluid, check if the fluidtank underneath has the same fluid.
If it's the same as the one being inserted, accept the fluid. If they aren't the same, do not accept the fluid.
If the fluidtank underneath will execute the same checks if active.

Original behaviour is the same with non active fluidtanks.


thommy101:
Two pictures describing before and after filling the top tank.
Top tanks are empty, pipe underneath configured for pull on top, push on bottom.
The three top tanks in the middle are active (Shift click once with configurator).
Bottom once aren't outputting for demonstration purposes.
One of the bottom once is empty, one has lava, one has water.

Initial setup
After

aidancbrady:
Implemented perfectly, I love how the functionality only applies when the tank is in active state.  Thanks!

PR #4413, opened 2017-06-01
fix collision box of chargepad

marcin212:
This PR sponsored by the deaths of two players and a Robit.


aidancbrady:
Self-merged a cleaner fix, thanks!

PR #4419, opened 2017-06-02
Fixed IllegalArgumentException in OreDictCache class

### LINKED ISSUE #4418: Destroing sign with Atomic Disassembler throws issue and crashes on Server
When you destroy a Sign with Atomic Disassembler it throws IllegalArgumentException in OreDictCache.

**Mekanism Version**: Built from branch 1.11 yesterday after marcin212's PRs merges

Crash Log: https://hastebin.com/adeyutatit.sql

PlayerWithoutName:
As in 1.11 Forge no longer allows ItemStacks to be null, you can change this condition to check.isEmpty(), which also fixes #4418


aidancbrady:
Thanks!

PR #4422, opened 2017-06-02
Fixed reactor blocks dropping wrong items

### LINKED ISSUE #4420: Placing most of Reactor Blocks and destroing them doesn't return the same item
Ie. When you place and destroy Reactor Frame and destroy it, it returns block with its unlocalized name, when do the same with Reactor Port it gives you block with unlocalized name and when you place it again, it appears to be Reactor Controller

**Mekanism Version**: Built from branch 1.11 yesterday after marcin212's PRs merges

PlayerWithoutName:
Also, you should consider changing all of your classes to implement ITileEntityProvider instead of extending BlockContainer and switch from ISidedInventory to Capabilities

Fixes #4420


thommy101:
Looks like it fixed it!

The note on Capabilities vs ISidedInventory is false, Mekanism supports both of them at the same time for cross mod compatibility.

thommy101:
Also test build one to confirm functionality, everything works as supposed!

PR #4426, opened 2017-06-03
Fixed evaporator liquid rendering

### LINKED ISSUE #4425: Thermal Evaporation Plant causes massive lag while entering world
It is caused by instancing many display lists during recalculation of water model inside the plant.

Reproduction: Add one level to fully built and filled plant

PlayerWithoutName:
Simplified it a little and fixed it's lagginess, now it generates max 17 Display Instances per fluid ;)

Fixes #4425

### DIFF
--- a/src/main/java/mekanism/client/render/tileentity/RenderThermalEvaporationController.java
+++ b/src/main/java/mekanism/client/render/tileentity/RenderThermalEvaporationController.java
@@ -106,8 +106,6 @@ public class RenderThermalEvaporationController extends TileEntitySpecialRendere
 	}
 
 
-
-	@SuppressWarnings("incomplete-switch")
 	private DisplayInteger[] getListAndRender(Fluid fluid)
 	{
 		if(cachedCenterFluids.containsKey(fluid))
@@ -143,6 +141,7 @@ public class RenderThermalEvaporationController extends TileEntitySpecialRendere
 		}
 
 		MekanismRenderer.resetColor();
+		cachedCenterFluids.put(fluid, displays);
 		return displays;
 	}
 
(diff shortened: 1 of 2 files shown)


thommy101:
If this fixes what I think it fixes, it would be great! Adding layers causes so much lag right now (full stops for 3 seconds)

PlayerWithoutName:
Yeah, you will probably adapt a little but main idea was not to generate so much instances as on my processor it was taking about 10 seconds to unfreeze after joining and i had to disassemble Evaporation Plants because my friend's laptop was timing out before these instances could generate.

src/main/java/mekanism/client/render/tileentity/RenderThermalEvaporationController.java
```diff
@@ -74,118 +105,67 @@ private void push()
 		GL11.glBlendFunc(GL11.GL_SRC_ALPHA, GL11.GL_ONE_MINUS_SRC_ALPHA);
 	}
 
+
+
 	@SuppressWarnings("incomplete-switch")
```
thommy101: Since the switch has been removed, this is obsolete

Meller127:
Will this be pushed to 1.10 branch also?

PlayerWithoutName:
Fixed fluid caching as i've realized that i forgot to cache lists after generating

thiakil:
Can someone explain to me how the rendering is supposed to look? (or currently looks at least)

I know that looking down from above you see the top of the fluid at the relevant level, but how does it look when standing inside the tower?

thommy101:
Top, bottom looking up, half way looking down

<details>
  <summary>Click to expand</summary>
Top
Bottom looking up
Half way looking down
</details>

aidancbrady:
Thanks!

PR #4429, opened 2017-06-03
Fix integer overflows

thommy101:
Using Math.min to limit it to the max value of an integer.
Math.round(double a) returns a long and casting a long which is greater than the max int to an int will do all kind of things but the things you want it to do.


aidancbrady:
Perfect, thanks. I thought I caught all of these

PR #4432, opened 2017-06-03
Improve travis compile time

thommy101:
Sped up the check time of Travis. It only checked it, but somehow if defined explicitly together with a minimal CI workspace reduces the run time by 50%.

Ran it twice (once to create the build cache):
https://travis-ci.org/thommy101/Mekanism/builds/239091571


aidancbrady:
Thanks!

Issue #4449, opened 2017-06-06
Crashing on Loading World


PessimisticBastard:
Whenever I load up my save looking at various parts from Mekanism, my world immediately turns to black, with some white boxes showing up. 

However, so long as I let the world load without looking at anything from Mekanism, it loads just fine, and I continue working on Mekanism stuff with no issue. 

Crash log: https://pastebin.com/iEk4M8iw

Edit: I should add this is for 1.11.2

thiakil:
The crash itself looks like #4434 (seems to be basically a race condition)

As for the world turning to black (I'm assuming you mean it doesn't always crash), this could be a cross mod issue (you have quite a number of CoreMods), or perhaps an issue with the resource pack in use (perhaps shaders).

PR #4453, opened 2017-06-08
Attempt to resolve race condition with set/getDefault, Add sanity che…

thiakil:
Just a simple bail to super method if getDefaultForTile returns a null, and an attempt to prevent it from returning null in the first place.

Aimed at fixing #4449


thommy101:
Is this by any chance related to the fix Aiden already pushed out? 83fae924b813c805e8c1fc0adc552800b456f164

thiakil:
Same flow as far as I can tell - I didn't see that one so no idea if referenced issue submitter had that commit. So probably just means a couple more sanity checks ¯_(ツ)_/¯

thommy101:
The referenced commit isn't in a publicly available version as far as I know. So I don't think the issue submitter had that commit already in it's version

aidancbrady:
Yeah, 99% sure this PR accomplishes the same as my fix in the dev branch.  Thanks though!

Issue #4288, opened 2017-04-02
work with tinkers without registerAllItems=true


BrianEdmonds:
TL;DR: How much work would it be to have Mekanism advertise the metals it provides such that Tinkers Construct will automatically include support for them?

I'm working on a modpack that contains both Mek and TC, and I've found that without the TC config setting registerAllItems=true, TC does not recognize that Mek is providing Tin and Copper, so , for example, the tin and copper variants of the tool forge are not available to craft, among many other issues.

If it could also support Osmium as part of this integration work, that would be even better.

aidancbrady:
If someone wanted to submit a TC support pull request I'd consider merging it, but in the meantime I have too many other issues to look at.

BrianEdmonds:
@aidancbrady Out of curiosity, do you have something you use to track
planned work, or is all in your head?

laz2727:
It should automatically register them (TiC catches oredict event for that). Wonder what's the bug is.

Landmaster:
@aidancbrady @laz2727 @BrianEdmonds The reason why they are not automatically registered is that

> you are registering them in `init`, [whereas they are supposed to be registered in `preInit`.](https://mcforge.readthedocs.io/en/latest/utilities/oredictionary/) This is causing problems with other mods which depend on the oredicts, such as Tinkers Construct.
>
>The same problem occurs with your copper, tin, and bronze.
>
>(I discovered this problem early on when working on my addon mod [PlusTiC](https://github.com/Landmaster/PlusTiC) and adding Mekanism integration, and had to manually register the metals as a workaround.)

(from my comment on #4463)

thommy101:
Created pull request #4465 to fix this issue.

PR #4465, opened 2017-06-13
Fix oredict registration, moved to preinit

### LINKED ISSUE #4463: Mekanism Steel isn't recognized by Tinker's Construct
[I'm assuming this is a Mekanism problem because all other mods that have steel are recognized by Tinkers' as far as I can tell]

Mekanism Steel Ingots **are** registered to the ore dictionary, but oddly enough Tinkers' Construct doesn't recognize it.

Using latest (CurseForge releases, not GitHub) Tinkers' Construct and Mekanism for 1.11.2, but according to this thread it is also a problem in 1.10.2. http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-mods/mods-discussion/2800566-no-steel-in-tinkers-construct

thommy101:
Fixes cross mod integration of oredict metals. Dusts, ingots, blocks are now correctly registered with TinkersConstruct after moving registration of oredict entries to preinit instead of init.

Fixes #4463 and #4288

### DIFF
--- a/src/main/java/mekanism/common/Mekanism.java
+++ b/src/main/java/mekanism/common/Mekanism.java
@@ -1240,6 +1240,9 @@ public class Mekanism
 		MekanismItems.register();
 		MekanismBlocks.register();
 
+		//Integrate certain OreDictionary recipes
+		registerOreDict();
+
 		if(Loader.isModLoaded("mcmultipart")) 
 		{
 			//Set up multiparts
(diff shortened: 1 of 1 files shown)


thommy101:
Must mention, I wouldn't have found it this quick without your (@Landmaster) analysis.

aidancbrady:
Hope this doesn't break anything!

PR #4468, opened 2017-06-14
More Null ItemStack / Null Comparison

thiakil:
I found a way to get Idea to analyse for `null` being passed where `ItemStack` was expected and where `ItemStack` was compared with `null` and gave Mekanism a check over.

For some of them (especially the internal use ones) I've left the `!= null` there to ensure no null pointer errors - so you may wish to go over them to verify if they're still needed.

### DIFF
--- a/src/api/java/ic2/api/energy/prefab/BasicEnergyTile.java
+++ b/src/api/java/ic2/api/energy/prefab/BasicEnergyTile.java
@@ -220,7 +220,7 @@ abstract class BasicEnergyTile implements ILocatable, IEnergyTile {
 	 * @return true if energy was transferred
 	 */
 	public boolean charge(ItemStack stack) {
-		if (stack == null || !Info.isIc2Available() || getWorldObj().isRemote) return false;
+		if (stack == null || stack.isEmpty() || !Info.isIc2Available() || getWorldObj().isRemote) return false;
 
 		double amount = ElectricItem.manager.charge(stack, energyStored, Math.max(getSinkTier(), getSourceTier()), false, false);
 
@@ -237,7 +237,7 @@ abstract class BasicEnergyTile implements ILocatable, IEnergyTile {
 	 * @return true if energy was transferred
 	 */
 	public boolean discharge(ItemStack stack, double limit) {
-		if (stack == null || !Info.isIc2Available() || getWorldObj().isRemote) return false;
+		if (stack == null || stack.isEmpty() || !Info.isIc2Available() || getWorldObj().isRemote) return false;
 
 		double amount = capacity - energyStored;
 		if (amount <= 0) return false;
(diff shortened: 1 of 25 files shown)


src/main/java/mekanism/client/gui/GuiChemicalCrystallizer.java
```diff
@@ -48,7 +49,7 @@
 
 	public Gas prevGas;
 
-	public ItemStack renderStack;
+	public @Nullable ItemStack renderStack;
```
thommy101: ItemStacks should not be null

src/api/java/ic2/api/energy/prefab/BasicEnergyTile.java
```diff
@@ -220,7 +220,7 @@ public boolean useEnergy(double amount) {
 	 * @return true if energy was transferred
 	 */
 	public boolean charge(ItemStack stack) {
-		if (stack == null || !Info.isIc2Available() || getWorldObj().isRemote) return false;
+		if (stack == null || stack.isEmpty() || !Info.isIc2Available() || getWorldObj().isRemote) return false;
```
thommy101: In 1.11.2 it isn't necessary to check on null, you can remove those.

src/main/java/mekanism/client/gui/GuiChemicalCrystallizer.java
thiakil: As far as I could find, it is not used outside this class; so it should be unnecessary to force it to non-null and add the overhead of an extra method call & the contained checks.

src/api/java/ic2/api/energy/prefab/BasicEnergyTile.java
thiakil: Just noticed that's in IC2's api, and probably not used by Mekanism. Shouldn't really have changed it at all tbh

src/main/java/mekanism/client/gui/GuiChemicalCrystallizer.java
thommy101: It isn't used in this project, but it's a public value which could be changed by any other mod or Mekanism Addon.

src/api/java/ic2/api/energy/prefab/BasicEnergyTile.java
thommy101: This isn't the only occurence, I only commented on the first one ;)
Changing the API isn't needed as you said, it will be overwritten by the next update of the API of IC2

src/main/java/mekanism/client/gui/GuiChemicalCrystallizer.java
thiakil: It is unlikely being a GUI class, and that would be the point of annotating it (and also why I mentioned Aidan might want to review how his classes do it)

PR #4495, opened 2017-06-27
Add safetycheck to FluidHandlerWrapper.

### LINKED ISSUE #4491: 1.10.2 - Mekanism/IE interop - cannot get CCW of Down
Mekanism 301 on 1.10.2 interfacing with IE 0.10-60 pipes. I am aware of https://github.com/aidancbrady/Mekanism/issues/4039, but we are using the latest Mekanism version and the problem persists.

A full crash log (it's a clientside crash) can be found [here](https://pastebin.com/XNJqZ1RC).

thommy101:
Fixes #4491


thommy101:
Wrong fix.. This won't fix the problem of the issue.

PR #4498, opened 2017-06-28
Fix MCMP ClassCastException and usability

### LINKED ISSUE #4492: [1.11.2] Crash when using universal cables with Chisels and Bits
#### Issue description:
When trying to put a universal cable through a hollowed out block from Chisels and Bits (either by placing the hollow block over the cable or by placing the cable in the hollow block) a crash occurs. This occurs both with the latest versions of all mods from Curseforge as well as with the current (as of reporting) 1.11 branch.

#### Steps to reproduce:
1. Using the chisel from Chisels and Bits hollow out the center of a block
2. Place a universal cable inside the hollow part of the block

**Forge:** 13.20.0.2315 (also 13.20.0.2304 from Gradle)
**Mekanism:** 1.11.2-9.3.2.307 (also the current 1.11 branch)
**MCMultiPart:** 2.0.1 (also 2.0.0_19 from Gradle)
**Chisels and Bits:** 13.11

https://pastebin.com/XGVGJKGK

christofferson:
Fixes #4492 

Additionally fixes subsequent issues encountered with transmitters becoming permanently orphaned when being added to / removed from multipart containers.


src/main/java/mekanism/common/block/BlockTransmitter.java
```diff
@@ -424,4 +442,24 @@ public TileEntity createNewTileEntity(World worldIn, int meta)
 		
 		return null;
 	}
+	
+	private Optional<TileEntitySidedPipe> getTileEntitySidedPipe(TileEntity entity)
```
thiakil: Is it necessary to return `Optional` instead of `null` like `getTileEntity` does?

christofferson: It's certainly not necessary, I simply used that method because it's what the MCMultipart source does and I found it to feel a bit cleaner as it made null checks a bit more obvious for when they're necessary.

It'd be easy enough for me to refactor it out, if desired

src/main/java/mekanism/common/integration/multipart/MultipartMekanism.java
```diff
@@ -162,4 +165,23 @@ public static boolean hasConnectionWith(TileEntity tile, EnumFacing side)
...
+		return entity instanceof IMultipartContainer;
+	}
+	
+	public static Collection<TileEntity> getTileEntities(TileEntity container)
+	{
+		if(container instanceof IMultipartContainer)
+		{
+			return ((IMultipartContainer)container).getParts().values().stream()
+					.map(partInfo -> partInfo.getPart())
+					.filter(part -> part instanceof MultipartTile)
```
thiakil: Either change the method name to `getMekanismTileEntities` or remove the filter. `IPartInfo` has a `getTile` method.

Personally I'm not really feeling the `stream` usage; it's very unclear what things are returning and therefore what the block is doing. It also seems to require 3-4 more loops than would be necessary for a plain for loop.  Not great for performance.

src/main/java/mekanism/common/integration/multipart/MultipartTransmitter.java
```diff
@@ -28,6 +30,28 @@ public IPartSlot getSlotFromWorld(IBlockAccess world, BlockPos pos, IBlockState
 	}
 	
 	@Override
+	public void onAdded(IPartInfo part)
+	{
+		TileEntity tile = part.getTile().getTileEntity();
+		
+		if(tile instanceof TileEntitySidedPipe) 
+		{
+			((TileEntitySidedPipe)tile).onPartChanged(null);
```
thiakil: use `TileEntitySidedPipe.onAdded`

src/main/java/mekanism/common/integration/multipart/MultipartTransmitter.java
```diff
@@ -28,6 +30,28 @@ public IPartSlot getSlotFromWorld(IBlockAccess world, BlockPos pos, IBlockState
...
+	}
+	
+	@Override
+	public void onPartAdded(IPartInfo part, IPartInfo otherPart)
+	{
+		TileEntity tile = part.getTile().getTileEntity();
+		
+		if(tile instanceof TileEntitySidedPipe) 
+		{
+			tile.validate();
```
thiakil: use `onWorldJoin`

src/main/java/mekanism/common/tile/transmitter/TileEntityTransmitter.java
```diff
@@ -50,6 +52,10 @@ public void onWorldJoin()
 		{
 			TransmitterNetworkRegistry.registerOrphanTransmitter(getTransmitter());
 		}
+		else if(lastClientNetwork != null)
```
thiakil: What exactly does this solve? The network should be synced to the client via packet at the end of the server tick.

src/main/java/mekanism/common/integration/multipart/MultipartMekanism.java
```diff
@@ -162,4 +165,23 @@ public static boolean hasConnectionWith(TileEntity tile, EnumFacing side)
 		
 		return Collections.emptyList();
 	}
+	
+	public static boolean isMultiPartContainer(TileEntity entity)
```
thiakil: This method is only used in one place.

src/main/java/mekanism/common/block/BlockTransmitter.java
```diff
@@ -226,16 +227,24 @@ public boolean onBlockActivated(World world, BlockPos pos, IBlockState state, En
...
-		TileEntitySidedPipe tileEntity = (TileEntitySidedPipe)world.getTileEntity(pos);
-		ItemStack itemStack = new ItemStack(MekanismBlocks.Transmitter, 1, tileEntity.getTransmitterType().ordinal());
-		
-		if(!itemStack.hasTagCompound())
+		ItemStack itemStack;
+		Optional<TileEntitySidedPipe> tileEntity = getTileEntitySidedPipe(world.getTileEntity(pos));
+		if(!tileEntity.isPresent())
 		{
-			itemStack.setTagCompound(new NBTTagCompound());
+			itemStack = ItemStack.EMPTY;
```
thiakil: set this in the variable declaration instead of an `if` block

src/main/java/mekanism/common/integration/multipart/MultipartTransmitter.java
christofferson: `validate` is used here not to trigger `onWorldJoin` but instead to make sure the tile entity remains valid when a multipart is added around an existing transmitter. If `validate` is not called at this point then the transmitter gets removed as invalid but since the TileEntity remains invalid the `OrphanPathFinder` doesn't iterate over it and it remains permanently orphaned.

src/main/java/mekanism/common/tile/transmitter/TileEntityTransmitter.java
christofferson: This solved an issue that may be unique to Chisels and Bits. When previewing placement of a chiseled block it shows a ghost block that only affects the client side. This preview caused `onWorldSeparate` to be triggered on the client side which invalidated the transmitter network but if the block was not placed then no server change occurs and, thus, no sync between server and client. When the ghost block is moved away from the transmitter `onWorldJoin` is called on the client side which I used to correct this.

It's not game-breaking but since the client side thinks the transmitter has no network it appears as if nothing is going through the transmitter despite everything else updating correctly.

src/main/java/mekanism/common/integration/multipart/MultipartTransmitter.java
thiakil: MCMP calls `validate` as far as I could find..?

src/main/java/mekanism/common/integration/multipart/MultipartMekanism.java
christofferson: Correct. This was created simply to keep direct references to the `mcmultipart.api` packages restricted to the `mekanism.common.integration.multipart` package.

src/main/java/mekanism/common/integration/multipart/MultipartTransmitter.java
christofferson: In this case MCMP validates the new part being added (in this case the chiseled block) but before it does that it ends up changing the block state at the position of the transmitter which causes the block to break thereby invalidating the transmitter. Since the transmitter TileEntity should still be valid (it just exists as part of the new multipart block) this works around that.

christofferson:
Retrying Travis build

src/main/java/mekanism/common/integration/multipart/MultipartMekanism.java
```diff
@@ -162,4 +164,28 @@ public static boolean hasConnectionWith(TileEntity tile, EnumFacing side)
...
+	{
+		return entity instanceof IMultipartContainer;
+	}
+	
+	public static Collection<TileEntity> getMekanismTileEntities(TileEntity container)
+	{
+		ArrayList<TileEntity> entities = new ArrayList<>();
```
thiakil: I don't see why this needs to return a list; there will only ever be one TileEntitySidedPipe in the block as far as I know

src/main/java/mekanism/common/integration/multipart/MultipartMekanism.java
```diff
@@ -162,4 +164,28 @@ public static boolean hasConnectionWith(TileEntity tile, EnumFacing side)
...
+		
+		if(container instanceof IMultipartContainer)
+		{
+			for(IPartInfo partInfo : ((IMultipartContainer)container).getParts().values())
+			{
+				IMultipartTile partTile = partInfo.getTile();
+				if(partTile instanceof MultipartTile)
+				{
+					entities.add(((MultipartTile)partTile).getTileEntity());
```
thiakil: this cast is also redundant

src/main/java/mekanism/common/integration/multipart/MultipartTransmitter.java
thiakil: In that case, does `onWorldJoin` need to be called? If not, would it be better to call only `TileEnttity.validate` (via another method)?

src/main/java/mekanism/common/integration/multipart/MultipartMekanism.java
christofferson: Thanks for catching this, it was a bit of Eclipse auto-complete that I missed

src/main/java/mekanism/common/integration/multipart/MultipartMekanism.java
christofferson: You are correct, there will only ever be one `TileEntitySidedPipe` but there may be other Mekanism tile entities in the multipart (namely glow panels).

That being said after investigating this I've found a much better way of getting the appropriate tile entities by taking advantage of MCMP's own world wrapping so I'll be replacing this method with that.

src/main/java/mekanism/common/integration/multipart/MultipartTransmitter.java
christofferson: Unfortunately it does because the override of `invalidate` calls `onWorldSeparate`

src/main/java/mekanism/common/tile/TileEntityGlowPanel.java
```diff
@@ -37,6 +37,11 @@ public void setOrientation(EnumFacing newSide)
 	@Override
 	public void handlePacketData(ByteBuf dataStream) throws Exception 
 	{
+		if(Mekanism.hooks.MCMPLoaded)
+		{
+			dataStream.readByte();
```
thiakil: what are all these unused bytes for?

src/main/java/mekanism/common/integration/multipart/MultipartTileNetworkJoiner.java
```diff
@@ -0,0 +1,83 @@
...
+import io.netty.buffer.ByteBuf;
+import mcmultipart.api.container.IMultipartContainer;
+import mcmultipart.api.slot.EnumFaceSlot;
+import mcmultipart.api.slot.IPartSlot;
+import mcmultipart.api.world.IMultipartBlockAccess;
+import mekanism.common.base.ITileNetwork;
+import net.minecraft.tileentity.TileEntity;
+import net.minecraft.world.IBlockAccess;
+
+public class MultipartTileNetworkJoiner implements ITileNetwork
```
thiakil: I think this class should have some javadoc explaining
- what it is
- why it is necessary
- how it interacts with the existing capability

src/main/java/mekanism/common/integration/multipart/MultipartMekanism.java
thiakil: IJ Idea highlights it ;)

src/main/java/mekanism/common/tile/TileEntityGlowPanel.java
christofferson: These are the single byte ordinal headers mentioned in the commit description. `TileEntityGlowPanel` is the only possible multipart that attaches to a face so it adds its `EnumFacing` ordinal as the header while the transmitters all use `6` (one higher than the highest possible `EnumFacing` ordinal) to indicate the center slot.

These are used by the new `MultipartTileNetworkJoiner`.

src/main/java/mekanism/common/integration/multipart/MultipartTileNetworkJoiner.java
christofferson: When a capability is requested from a multipart container that request is passed on to the parts it contains. When multiple parts have the same capability a `Joiner` (as defined by MCMP) is required, or else only the first part the container finds is used.

Since it's possible for multiple parts implementing the `Capability<ITileNetwork>` capability to exist in the same container (both a cable and a glow panel) a joiner is necessary to ensure that packets arrive at their appropriate destination.

From `mekanism.common.network.PacketTileEntity.java`: 
```
PacketHandler.handlePacket(new Runnable() {
@Override
public void run()
{
    TileEntity tileEntity = message.coord4D.getTileEntity(player.world);
    
    if(CapabilityUtils.hasCapability(tileEntity, Capabilities.TILE_NETWORK_CAPABILITY, null))
    {
        ITileNetwork network = CapabilityUtils.getCapability(tileEntity, Capabilities.TILE_NETWORK_CAPABILITY, null);
        
        try {
            network.handlePacketData(message.storedBuffer);
        } catch(Exception e) {
            e.printStackTrace();
        }
        
        message.storedBuffer.release();
    }
}
}, player);
```

When the `TileEntity` at the message location is a multipart container with 2 or more parts that have the `ITileNetwork` capability MCMP requests the joiner that was registered for this capability. This creates a new `MultipartTileNetworkJoiner` which builds a map of which side (or center) each part in the container is attached to. It uses this map combined with the ordinal headers added to the possible multipart tile entities to route packets correctly.

I'm just heading out the door right now but I'll add some JavaDoc detailing this when I'm back.

thiakil: Good explanation, however the issue I have with this is that you add extra bytes in the individual classes, where it would be better to do so in this class.

christofferson: That is what I tried, initially however it causes issues when the tile entities themselves send packets instead of a capability being requested from a `TileEntity` in the world.

For example, in `mekanism.common.tile.transmitter.TileEntityLogisticalTransporter.java`:
```
@Override
protected EnumActionResult onConfigure(EntityPlayer player, int part, EnumFacing side)
{
    TransporterUtils.incrementColor(getTransmitter());
    onPartChanged(null);
    PathfinderCache.onChanged(new Coord4D(getPos(), getWorld()));
    Mekanism.packetHandler.sendToReceivers(new TileEntityMessage(new Coord4D(getPos(), getWorld()), getNetworkedData(new ArrayList())), new Range4D(new Coord4D(getPos(), getWorld())));
    player.sendMessage(new TextComponentString(EnumColor.DARK_BLUE + "[Mekanism]" + EnumColor.GREY + " " + LangUtils.localize("tooltip.configurator.toggleColor") + ": " + (getTransmitter().getColor() != null ? getTransmitter().getColor().getColoredName() : EnumColor.BLACK + LangUtils.localize("gui.none"))));

    return EnumActionResult.SUCCESS;
}
```

In this case, since there is no need to request a capability from the `TileEntity` at the world coordinates, the joiner on the server side never gets created and the ordinal headers would not be added to the packet. On the client side, when the packet is received, the packet handler would request the `TileEntity` from the world and that packet would be passed to `MultipartTileNetworkJoiner` but since the ordinal header is missing that packet would either be dropped or improperly routed.

thiakil: Would it not be possible to detect when the TE is inside a MP container inside `getNetworkedData`, and then pass to a `static` method in `MultipartTileNetworkJoiner` that wraps (i.e pushes the needed byte to the first index) the data?

christofferson: It's certainly possible, but I'm not sure what the end goal would be. Performance wise, considering we're talking about a single byte, I don't think we'd gain anything (and are more likely lose a little) by first checking if the TE is inside a MP container.

If we're looking for code style then it would mean that `MultipartTileNetworkJoiner` could remove the header byte itself and I suppose a static method (or methods, one to unwrap if we don't first check) would make the code a bit more self-documenting and eliminate the "magic number" usage in the entities themselves.

In the latter case, please let me know which route you'd rather go (check + static wrap method or static wrap/unwrap methods) and I'll make that change.

thiakil: sure, byte wise it's not much of a difference, but at the absolute least it needs a comment in each file about why its reading a byte and doing nothing with it - so it would look even tidier if you can manage to put it in the Cap. How exactly you go about that is mostly up to you.

How many other places bypass the cap to send messages?

christofferson: Quite a few, but the only places I can see that impacting multiparts are:
- `TileEntitySidedPipe.update()`
- `TileEntityLogisticalTransporter.onConfigure(EntityPlayer, int, EnumFacing)`
- `TileEntityThermodynamicConductor.sendTemp()`

thiakil: it looks like those three could be handled with some kind of check in `TileEntitySidedPipe#getNetworkedData/handlePacketData`  ?

christofferson: `getNetworkedData` certainly, since it simply uses an `ArrayList` and it would be easy to insert a byte as the first object. Unfortunately, for the case of `handlePacketData` since it's using a stream and the last override is called first (that, of course, being the ones in the individual transmitter implementations) trying to clear the header inside `TileEntitySidedPipe` wouldn't work.

What I think I'll do is, in the morning, I'll write up the code to check whether or not the TE is part of a container and call the static wrapper method in `TileEntitySidedPipe.getNetworkedData` / `TileEntityGlowPanel.getNetworkedData` and then test it having `MultipartTileNetworkJoiner` remove those headers before they're passed on to `handlePacketData`. That could work and clean up a lot of the code.

thiakil: Hmm I wonder why it doesn't call super right away, you'd think they'd do that normally

christofferson: I think it might be because of how the `TileEntityLogisticalTransporter` is set up. The first thing it does is `readInt()` to get the type of message and only calls `super.handlePacketData` if that type is 0 and the rest were just set up to call super afterwards.

Although the thermodynamic conductor seems to have data both before and after super so I really don't know.

aidancbrady:
Is this tested and ready to be merged?

christofferson:
I can't speak for any of the contributors but I've been testing quite a bit on my end.

Unfortunately there's also a bug in 1.11 MCMP during the ray trace events that can cause a crash (see amadornes/MCMultiPart#83) so unless the tester also gets the bug fix for that testing could prove to be difficult.

src/main/java/mekanism/common/integration/multipart/MultipartTileNetworkJoiner.java
```diff
@@ -88,11 +88,46 @@ public void handlePacketData(ByteBuf dataStream) throws Exception
...
+	 * @param data The network data list
+	 * @param facing The side this part is attached to or <code>null</code> for the center slot
+	 */
+	public static void addMultipartHeader(TileEntity entity, ArrayList<Object> data, EnumFacing facing)
+	{
+		int tileNetworkParts = 0;
+		IMultipartContainer container = MultipartMekanism.getContainer(entity.getWorld(), entity.getPos());
+		if(container != null)
+		{
+			for(IPartSlot slot : container.getParts().keySet())
```
thiakil: I'm not sure we need to count, I'd be fine with it if it adds the header just because it's inside a container. Thoughts?

christofferson: The joiner only gets created if 2 or more parts implement `ITileNetwork` so we do need to count in order to determine if the header is actually necessary, otherwise the header doesn't always get removed properly.

thiakil: alright, perhaps after incrementing `tileNetworkParts` you could insert a break `if > 1` ?

thiakil:
It's almost ready; I haven't tested the latest networking changes, but it certainly tested fine for the other parts

aidancbrady:
@thiakil, just let me know when I should merge. I'm going to try and release v9.3.3 tonight.

src/main/java/mekanism/common/integration/multipart/MultipartTileNetworkJoiner.java
christofferson: I mean, worst case scenario it reaches 7 and that's if, for some reason, someone attaches a glow panel to every single face of a transmitter but breaking `if > 1` certainly couldn't hurt. I'll add that in quickly here

thiakil: Yeah it's a minor performance tweak, but every bit could help. Also saves testing e.g C&B parts for a mek interface

thiakil:
Sure thing, though I won't have a chance to test this until well after your tonight haha

I've tested cd8cbf43b1fdfbc794613896926177731c6ef564 in case you want to try a partial merge (though IDK how GH PR handles that) - but of course if you have time to test the network stuff I'm fine to defer to your expertise

aidancbrady:
The code added looks good- I'll do a full merge and see if anything comes up. If we need to we can push a ninja-fix for any issues that arise.

@christofferson, I appreciate your help on this- after making the transmitter changes for the initial 1.11 release I knew this would be necessary but I was dreading taking it on myself. Let me know if you think of any other changes we'll need to make.

christofferson:
@aidancbrady I'm happy to help. I'm a big fan of the mod and, honestly, I'm thrilled I was able to add to it.

@thiakil Thanks a ton for helping me get this up to snuff.

thiakil:
You're welcome. I hope I didn't come off as too abrasive!

christofferson:
Not at all. I've been going through sources for quite some time and have been coding in Java for years but this is my first bit of real MC mod work so the directness was appreciated; it helped me look at my own work critically and come up with better ways to solve issues I never would have thought of otherwise.

PR #4504, opened 2017-06-30
1.10 bugfix backport

thommy101:
As mentioned at some issues, I'm backporting the fixes of 1.11 to 1.10. The idea is to backport a lot of fixes to make one last and final 1.10 version with the latest fixes to make that version a lot more stable for the users of it.

Huge thanks and shout out to @thiakil for the largest part of back porting!

Things added:
- ForgeEnergy support
- IItemHandler support
- Username to UUID conversion to make it non-breaking
- Formulatic assembler stock controll

Bugfixes:
- Silktouch drops of miner
- Better gas management
- Fluid issues
- IMC Recipe removal
- Runoff worldgen
- Better computer support
- Atomic assembler offhand
- Client desync causing wrong facings and nullpointers
- Lot more of little changes

Reason for WIP: There are some issues with the backported item handling of logistical transporter pipes.


thommy101:
There's still one known bug remaining in this regarding item handling with logistical pipes.
When a machine on auto output pushes something in a pipe which is connected to a chest the resulting itemstack has a size of 0.
Still debugging that fault, for the rest it looks stable and doesn't crash.

thiakil:
@thommy101 I'd suggest looking at `getPredictedInsert` if you haven't already. I recall having to fudge that method a lot because the API used didn't exist in 1.10

thommy101:
That does sound like a probable cause by the name of it, thanks!

thommy101:
Finally got some time to check the last issue here. Looks like getPredictedInsert is functioning correctly, the correct `TransporterStack` is given to `MultipartTransporter`'s `transit` HashList. But somewhere when transporting "in the pipe" the stack amount gets changed to 0. Still looking where that would/could happen.

thommy101:
Found the missing commit! ba3c37b6c3eb2d9cde0a681463725afc4610a7ff

Branch is ready for merge! 😃

thiakil:
O_O

Where I never would have thought to look (because of the commit message)

Jolly good work!

thommy101:
Commit messages might be slightly misleading... Learned it the hard way

I didn't look at the commits, I debugged for hell until I found one thing messing with itemstacks which could cause the bug, launching my 1.12 workspace I saw some differences and thanks to git annotate I found the corresponding commit :)

aidancbrady:
This is great, thanks!

Issue #4510, opened 2017-07-03
Factories not Retaining Energy


guicogo:
#### Issue description:
Factories are not retaining stored energy after breaking it. The non-factory version retain just fine.
This also happens on the recommended version 9.3.2.307.

#### Steps to reproduce:
1. Self-explanatory

#### Version (make sure you are on the latest version before reporting):
**Forge:** 13.20.1.2388
**Mekanism:** 9.3.2.309
**Other relevant version:** Minecraft 1.11.2

thommy101:
There's a bit more wrong than only after breaking.
Currently no factory retain the energy after a unload. If I place a factory, fill it with energy, leave world, join world all energy is lost. Looks like there's some information not being saved.

thommy101:
Narrowed it down to a client desync, the server has the correct data, but the client doesn't know about it. I'm not familiar with the data sync between them so I think i could not solve it. Maybe someone else can :)

thiakil:
Whereabouts does the server know about it, but the client doesn't?

thommy101:
One example I can give without checking my computer right now is setEnergy(). The max energy is unset. 

The packet sending and receiving is another one, server sends it, client doesn't save it

thommy101:
Did some more debugging, it are in fact two bugs, one is also happening in 1.10.

Bug 1: Clientside smelting factory `BASE_MAX_ENERGY` is 0 after reload.
Cause is TileEntityFactory `handlePacketData(ByteBuf)` checks if the `recipeType` received is equal to the one already known (`oldRecipe`). The default value of the `recipeType` is, as you can guess by now, `RecipeType.SMELTING` causing it to skip setting the values.

Bug 2: Not retaining energy after breaking/when using the pick block feature of creative.
Cause: not yet discovered, will update this when I found more for this bug.

PR #4530, opened 2017-07-14
[1.11] Different factory issues.

thommy101:
Fix smelting factories default max energy value.
Adresses #4510 

The fix for getMaxEnergy isn't the nicest, but I couldn't think of an other way to fix it for all usages.


aidancbrady:
Great stuff, your cleanup is exactly how I would have solved this.

PR #4534, opened 2017-07-16
[1.11] Fix logistical sorter handling with less than full stacks.

### LINKED ISSUE #4527: Logistical Sorters Delete Overflowing Items
#### Issue description:
   When using a logistical sorter to send items into a container, if the storage container being outputted into doesn't have enough space for a stack of an item that is in its input storage container, but does have space for a partial stack, it will send an entire stack through anyway, deleting what won't fit.

#### Steps to reproduce:
1. Place down two storage containers/machines with inventories, and connect them with a logistical sorter.
2. Fill the second container(the one on the sorters output) with any item, but leave one slot only partially full
3. Put enough of said item into the first container (input end) so that it would be MORE than enough to fill the second container.
4. the sorter will pull a/the stack of items out of the first container, put what it can in the second, and delete the rest of that stack

#### Current workaround:
   Make sure your logistical sorters always have enough space to output into, such as a double chest that then outputs directly into a machine using regular pipes, rather than using the sorter to go straight into the machine. This will prevent any overflow from happening and thus, prevent item deletion.

#### Version:
**Forge: 13.20.0.2228** 
**Mekanism: 9.3.2** 
**Minecraft: 1.11.2**

thommy101:
Fixes #4527

### DIFF
--- a/src/main/java/mekanism/common/content/transporter/InvStack.java
+++ b/src/main/java/mekanism/common/content/transporter/InvStack.java
@@ -77,7 +77,7 @@ public final class InvStack
 				}
 				else {
 					ItemStack ret = stack.copy();
-					ret.setCount(inventory.getStackInSlot(slotIDs.get(i)).getCount() - stack.getCount());
+					ret.setCount(inventory.getStackInSlot(slotIDs.get(i)).getCount() - amount);
 					inventory.setInventorySlotContents(slotIDs.get(i), ret);
 					amount -= stack.getCount();
 				}


aidancbrady:
Great job fixing this. Transporter code is some of the most complicated stuff I've written, which tends to make bugs like this hard to track down.

thommy101:
It indeed looka like something realy complicated, but a 100% reproduction and a debugger make bugs like these easy to solve :)

PR #4535, opened 2017-07-17
Use FMLInterModComms.IMCEvent to process IMC messages

thiakil:
I noticed while working on Applied Llamagistics that IMC added recipes (e.g. crusher, enrichment) don't show in JEI - I assume this is because Mek's `LoadComplete` handler fires after JEI's.

So, two options: use the dedicated event or process in postInit.

`IMCHandler` could probably also be static since it's not directly an event handler - as far as Idea can tell it's only ever used in the one place.

The `LoadComplete` event seems to also be frowned upon judging from the Javadoc.

### DIFF
--- a/src/main/java/mekanism/common/Mekanism.java
+++ b/src/main/java/mekanism/common/Mekanism.java
@@ -1212,9 +1212,9 @@ public class Mekanism
 	}
 	
 	@EventHandler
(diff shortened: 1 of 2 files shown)


aidancbrady:
Lovely, thanks.

PR #4536, opened 2017-07-17
Suggested Integration Changes for AE2

thiakil:
NB: this is mainly based on my for of AE2 for 1.11+, Applied Llamagistics.

1. Universal cables should be registered to be a ForgeEnergy tunnel attunement. You're free to register it conditionally as IC2 if you want though

2. Commented out recipes for Certus and Fluix processes, as I put back the integration on the AL side with its custom recipe system. I've left sand and nether quartz in because that's not exclusive to AL.
NB: this commit should **not** be merged to the 1.10 branch as such integration does not exist there.

### DIFF
--- a/src/main/java/mekanism/common/integration/MekanismHooks.java
+++ b/src/main/java/mekanism/common/integration/MekanismHooks.java
@@ -156,13 +156,6 @@ public final class MekanismHooks
 	@Method(modid = "appliedenergistics2")
 	public void registerAE2P2P() 
 	{
-		String energyP2P = "add-p2p-attunement-rf-power";
-		
-		if(IC2Loaded)
-		{
-			energyP2P = "add-p2p-attunement-ic2-power";
-		}
-		
 		for(TransmitterType type : TransmitterType.values())
 		{
 			if(type.getTransmission().equals(TransmissionType.ITEM))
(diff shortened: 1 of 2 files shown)


aidancbrady:
Thanks!

asiekierka:
Do note that the second commit (which comments out the crusher/enrichment recipes) only keeps compatibility with Applied Llamagistics. Ideally, one would modify this to load the recipes in if AE2 and not Applied Llamagistics is detected. Not sure if they have separate mod IDs or not, though.

thiakil:
Nope, same modid. Mekanism does have IMC methods for them to add the recipes if they desire though (doesn't even need the API jar)

asiekierka:
@thiakil Would restoring the IMC calls affect Applied Llamagistics in any way, though? I mean, you removed them for a reason, I hope?

thiakil:
AL adds slightly different recipes (numbers differ slightly IIRC). Short answer is that I feel it better to leave it to the mod adding the items to decide what is fair exchange ("balance" to some)

thommy101:
So, if I understand @thiakil correctly, ideally we should/could submit a PR to AE2 (not AL2) which adds the necessary IMC calls to add the in this PR removed recipes?

thiakil:
That is a solution, yes

PR #4543, opened 2017-07-19
[1.11] Make empty containers stackable.

### LINKED ISSUE #3802: [feature/enhancement] stackable empty "containers"
make cardboxes, bins etc. stackable as long as they are empty, similar to the way buckets work (empty stackable, non-empty not)

right now (1.10.2) stacking of empty bins is very buggy, sometimes dropping an empty bin after i tried to stack them
trying to place cardboxes into a bin doesn't work in the world, and in the inventory it again is somehwat buggy, resulting in 0-sized stacks and doubled icons (though no doubled items)

thommy101:
Makes Cardboardboxes, gastanks and fluidtanks stackable. Bins are already stackable.
Fixes #3802


aidancbrady:
Great!

Issue #3808, opened 2016-09-17
[enhancement] provide a way to disable the "auto-ascend" on free runners


rapus95:
The sense is that one often uses some sort of 1-high barrier to prevent falling into a lava lake etc. (especially when mining)

Btw. is the amount of consumption reduced when wearing an armored jetpack?

SuperFanatic:
This. Oh *please* this!

thommy101:
Created a pull request to add a disabled feature to free runners :)

thommy101:
Implemented in https://github.com/aidancbrady/Mekanism/commit/6f78c19cb202b131ec729483678c1ce3e3904146

PR #4545, opened 2017-07-19
[1.11] Add disable toggle to freerunners.

thommy101:
Adds feature #3808.

Also changed the default armor switch key binding from F to G as F is currently a default key binding for switching held item with offhand.


src/main/java/mekanism/client/MekanismKeyHandler.java
```diff
@@ -46,17 +48,19 @@
 {
 	public static final String keybindCategory = "Mekanism";
 	public static KeyBinding modeSwitchKey = new KeyBinding("Mekanism " + LangUtils.localize("key.mode"), Keyboard.KEY_M, keybindCategory);
-	public static KeyBinding armorModeSwitchKey = new KeyBinding("Mekanism " + LangUtils.localize("key.armorMode"), Keyboard.KEY_F, keybindCategory);
+	public static KeyBinding armorModeSwitchKey = new KeyBinding("Mekanism " + LangUtils.localize("key.armorMode"), Keyboard.KEY_G, keybindCategory);
```
thiakil: Why the change to the existing keybind default? Too many clashes?

thommy101: See original comment ;) Minecraft uses F for switching items with offhand

thiakil: -facepalm- my bad! It's late xD

thommy101: Don't worry, we all do make comments like that :)

aidancbrady:
Wow! Impressive, you followed the jetpack/gas tank implementation and added full client-server sync for other users. This would make it easy to add some kind of graphic change to the free runners in the future when the mode is changed. Thanks!

Issue #3807, opened 2016-09-17
[enhancement] place drops of the bins directly into the inventory


rapus95:
when one clicks a bin to get items, a stack is fired into the world. That has 2 negative side effects:
a) if 2 or more people are standing at close range, the wrong guys gets the stack way too often,
b) (less critical) when dropping the item into the world instead of the inventory, an EntityItem has to be created for that fraction of a second until the item is collected and the entityItem has to be gc'ed

PR #4547, opened 2017-07-19
[1.11] Add bin drops directly to player inventory.

thommy101:
If there's no space, drop them in front of the bin.
Adds feature #3807

### DIFF
--- a/src/main/java/mekanism/common/block/BlockBasic.java
+++ b/src/main/java/mekanism/common/block/BlockBasic.java
@@ -51,6 +51,7 @@ import net.minecraft.entity.EntityLivingBase;
 import net.minecraft.entity.item.EntityItem;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
+import net.minecraft.init.SoundEvents;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.tileentity.TileEntity;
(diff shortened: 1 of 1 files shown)


src/main/java/mekanism/common/block/BlockBasic.java
```diff
@@ -412,12 +414,27 @@ public void onBlockClicked(World world, BlockPos pos, EntityPlayer player)
...
-					else {
-						world.spawnEntity(new EntityItem(world, player.posX, player.posY, player.posZ, bin.remove(1).copy()));
+					else
+					{
+						stack = bin.removeStack().copy();
+					}
+
+					if(!player.inventory.addItemStackToInventory(stack))
+					{
+						BlockPos dropPos = pos.offset(bin.facing);
```
thiakil: Not at the player's feet?

thommy101: Thought it was nicer to drop them out of the bin.
They currently drop at the players feet, this drops it (without any velocity) at the block in front of the bin.

Open for suggestions which one is nicer :)

thiakil: hmm I guess it won't make a huge difference since you have to be within reaching distance of the face anyway.

Minor difference of 'oops I can't pick that up' vs 'oops my hands & inventory is full (but I can pick it up for a split second?)' lol

Also I can't remember if you can activate this in creative, but if you can the `addItemStackToInventory` will never fail, even if it is full

thommy101: I think (can't test this version right now as I'm currently murdering the code) that interacting with the bin in creative breaks the bin instead of removing items from it.

thiakil: Yeah that was my recollection too

PR #4549, opened 2017-07-21
[1.11] Fix TE access from ChunkCache.

### LINKED ISSUE #3534: [Issue] Tesselating block model
Hi,

Random client crash when i'm in my base, it seems to begin to happen since i've built a boiler.

Crash log : http://pastebin.com/qQqD8yf6

thommy101:
Fixes #3534


thiakil:
What about adding a `getTileEntitySafe` (name others have used in the past) or `getTileEntityNoCreate` to one of the Utils classes? Saves writing the `instanceof` everywhere

thommy101:
Sounds like a good idea, will rewrite it ;)

thommy101:
Does indeed look better! Thanks!

PR #4550, opened 2017-07-21
ISidedInventory sanity & UsernameCache safety

### LINKED ISSUE #4484: [1.11.2] Server Crash
#### Issue description:
 Hello! A few days ago, our official pack server started crashing whenever someone is near a Mekanism machine (tested with a Metallurgic Infuser and Enrichment Chamber). At first we thought that it was all due to Mekanism, but today we figured out that it doesn't happen when the crafting recipes in JEI bug out, as it sometimes happens with the 1.11.2 version. Then we tried changing the JEI version on the client back to the version where all was fine, but it didn't help. Either way, by the crash log we can tell that it has something to do with Mekanism.

**Forge:** 13.20.0.2315
**Mekanism:** 9.3.2.307
**JEI (if it actually has to do with it)** 4.5.0.289

https://paste.ee/p/a1wzk

thiakil:
Fixes #4484 

Returns `<???>` if neither Forge's cache nor Yggdrasil knows about it. Could easily be changed to a localised string, but this saves a lookup.


thommy101:
Maybe add a log (with stacktrace) when <???> is returned so it can be traced where it could possibly go wrong?

thiakil:
Yeah we could add that, but I don't want to cause too much logspam, so needs a list of failed lookups I guess

thommy101:
True, it could get quite spammy. Logging each failed username once should be enough I think.

aidancbrady:
Perfect, thanks!

Issue #4540, opened 2017-07-18
Crash on startup with Mek 1.10


warplord:
*Please use the search functionality before reporting an issue. Also take a look at the closed issues!*

#### Issue description:
Crash on startup when I updated from Mekanism v9.2.2.301 to final

#### Steps to reproduce:
1. Start the game
2.
3.

#### Version (make sure you are on the latest version before reporting):
**Forge:** latest for 1.10
**Mekanism:** final
**Other relevant version:** 

#### If a (crash)log is relevant for this issue, link it here: (_It's almost always relevant_)

http://paste.feed-the-beast.com/view/269205cd

thommy101:
Note: Reportedly happening after updating Mekanism. Also reported on the reddit post about this update. 
Linking to https://github.com/SonarSonic/Flux-Networks/issues/124

warplord:
So the final version is unusable? OR is there anything that can be done?

thommy101:
Ik not sure what causes the crash, I'll look into it a bit more tomorrow (currently on mobile)

bytetrip:
This should probably be closed. FluxNetworks is using the old API and crashing. It's already reported there: https://github.com/SonarSonic/Flux-Networks/issues/124

thommy101:
I'm happy it's reported here. Gives us insight at what happened and we might be able to solve it. 

The reason it's breaking isn't their fault, we released a version without backwards compatibility.

thommy101:
Created a pull request for Flux-Networks to fix this issue.

thiakil:
This should be fixed in this Mekanism version (FN should load, but any specific Mek functions will require them to update)
https://minecraft.curseforge.com/projects/mekanism/files/2455227

PR #4553, opened 2017-07-23
[1.10 - compat] add back removed interfaces (and the capability just in case) 

thiakil:
Marked as deprecated, should be excluded from the MDK zip.

Tested with the Flux Networks from #4540 and it loads fine.

The capability is a complete dud, but in the unlikely event anyone does use the field in `Capabilities` they likely won't check it's non-null, which could cause issues.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -128,6 +128,10 @@ task MDKZip(type: Zip) {
     baseName = 'MDK'
     from sourceSets.main.java.srcDirs
     include 'mekanism/api/**'
+    exclude 'mekanism/api/energy/ICableOutputter.java'
+    exclude 'mekanism/api/gas/IGasTransmitter.java'
+    exclude 'mekanism/api/ItemRetriever.java'
+    exclude 'mekanism/api/reactor/**'
 }
 
 task releaseJars(type: Copy) {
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -128,6 +128,10 @@ task MDKZip(type: Zip) {
     baseName = 'MDK'
     from sourceSets.main.java.srcDirs
     include 'mekanism/api/**'
+    exclude 'mekanism/api/energy/ICableOutputter.java'
+    exclude 'mekanism/api/gas/IGasTransmitter.java'
+    exclude 'mekanism/api/ItemRetriever.java'
+    exclude 'mekanism/api/reactor/**'
```
thommy101: Nice doing this!

thommy101:
Just some explanation why this PR is opened:
The latest version has some breaking API changes which resulted in a couple mods crashing with a class not found exception. This restores binary compatability, but still has breaking features as Mekanism itself doesn't use these classes anymore.

The result after this: Other mods who use the removed classes will continue to work, although not making use of the feature they think they want. They must update to restore broken functionality.

At this moment we've identified two mods (FluxNetworks and Galacticraft) using one of the removed API classes (ICableOutputter) and opened a PR to both of them to update to the latest version of the 1.10 API (not accepted/no comment on both).

aidancbrady:
Perfect, thanks!

Issue #4507, opened 2017-07-03
v9.309 Lose all pipe/power connections when outside 'chunk' area even if area chunkloaded


FSFan:
*Please use the search functionality before reporting an issue. Also take a look at the closed issues!*

#### Issue description:
Whether or not you are running a chunk loader (FTBUtils for example) .309 still loses all connections
and all my machines shut down.  Including my reactor.  

#### Steps to reproduce:
1.  Chunk load using FTBUtils (or any other chunk loader
2.  Make sure area is showing chunkloaded
3.  Go outside what would be the normal distance for the chunk to be unloaded

#### Version (make sure you are on the latest version before reporting):
**Forge:** 8.0.99.99
**Mekanism:** 9.3.2.309
**Other relevant version:**  Minecraft 1.11.2

#### If a (crash)log is relevant for this issue, link it here: (_It's almost always relevant_)

[gist/pastebin/etc link here]

FSFan:
Sorry.  Forge 13.20.1.2393.  My bad

thiakil:
Did you verify it still happens with another chunkloader? I could be completely mistaken but I remember reading something about FTBLib lets the chunk unload and then reloads it chunkloaded.

FSFan:
Yes.  I tried using one called "achunkloader" and it does the same thing.  But thanx for the suggestion!

-Rick

kbcoder:
hopefully this will give some additional insight to this problem.
MC 1.11.2
Forge Version 2393
Mek Versions 311

I wasn't using a chunkloader, logged in to find all mek. pipes not working. All pipes had to be busted and replaced. This includes pressurized, Mechanical and universal. RF readers would show universals running at anywhere from -21 to 4RF. Any universal that was already in place had to be broken and replaced, not even connecting them  to newly placed working universals would fix them. The RF flow would stop  at the ones that have not yet been replaced.  

Things to note:
I was running the digital miner a few chunks away using an entangelporter. So maybe something went wrong with that interaction while not logged in (??). But, I have done something similar a day or so ago with no problems either.

aidancbrady:
Can anyone provide a strict reproduction scenario which will lead to this issue? I'd really like to get this solved soon but I've been unable to reproduce anything similar.

rmonk:
Forge: 13.20.0.2310
Mekanism: Mekanism-1.11.2-9.3.2.309.jar

[...]

PR #4555, opened 2017-07-24
Performance improvements, pass 1

Ricket:
My "All The Mods 2" server was being very slow. A user made a 5x ore processing flow and used Thermal Dynamics pipes for all the energy/fluid/item transfers, since the Mekanism pipes are currently really broken (#4507 and another issue where you have to place each pipe piece twice). The server's TPS was having severe issues, I profiled to see what's going on. Methods like "receiveEnergy" and "fill" were taking the vast majority of the time. So I dug in.

The result of today's investigation is these commits. I drastically increased the performance of these methods. For example receiveEnergy was not accounting for if we were already full of energy. It was always checking if the side was valid and marking the chunk dirty. You can look at each commit for its improvements.

The observer thing is something I haven't followed through with yet. My general thinking is like this: getConfig seems to be a bit slow (according to the profiler), and for these purposes it stores which sides are set to input, output, etc. That's something that the user generally changes shortly after placing the machine, and then doesn't touch again. So most of the effort of going and getting the config is wasted; we should be able to pretty aggressively cache which side is which. I figured if we add an observable and then trigger it when a side data is changed, then we can cache/memoize some methods and just update the cached values when the observer triggers an event.

### DIFF
--- a/src/main/java/mekanism/common/tile/component/SideConfig.java
+++ b/src/main/java/mekanism/common/tile/component/SideConfig.java
@@ -1,0 +1,56 @@
+package mekanism.common.tile.component;
+
+import net.minecraft.util.EnumFacing;
+
+import java.util.Observable;
+import java.util.Observer;
+
+public class SideConfig {
+    private byte[] directions = new byte[EnumFacing.VALUES.length];
+    private Observable observable = new Observable();
+
+    public SideConfig()
+    {
+    }
+
+    public SideConfig(byte[] b) {
+        this(b[0], b[1], b[2], b[3], b[4], b[5]);
+        assert b.length == EnumFacing.VALUES.length;
+    }
+
+    public SideConfig(byte d, byte u, byte n, byte s, byte w, byte e)
+    {
+        setDirections(d, u, n, s, w, e);
+    }
+
+    public void setDirections(byte d, byte u, byte n, byte s, byte w, byte e) {
+        directions[0] = d;
+        directions[1] = u;
+        directions[2] = n;
+        directions[3] = s;
+        directions[4] = w;
+        directions[5] = e;
+        observable.notifyObservers(this);
+    }
+
+    public byte get(EnumFacing f) {
+        return directions[f.ordinal()];
+    }
+
+    public byte[] asByteArray() {
+        return directions;
+    }
+
+    public void set(EnumFacing f, byte value) {
+        directions[f.ordinal()] = value;
+        observable.notifyObservers(this);
+    }
+
+    public void addObserver(Observer o) {
+        observable.addObserver(o);
+    }
+
+    public void deleteObserver(Observer o) {
+        observable.deleteObserver(o);
+    }
+}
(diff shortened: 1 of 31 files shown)


aidancbrady:
Great, appreciate the changes. Once we get the transmitter issues sorted out Mekanism is gonna be more stable than ever.

src/main/java/mekanism/common/tile/component/SideConfig.java
```diff
@@ -0,0 +1,56 @@
+package mekanism.common.tile.component;
+
+import net.minecraft.util.EnumFacing;
+
+import java.util.Observable;
+import java.util.Observer;
+
+public class SideConfig {
+    private byte[] directions = new byte[EnumFacing.VALUES.length];
+    private Observable observable = new Observable();
```
thiakil: Shouldn't `SideConfig` extend `Observable`?

Also appears `Observable#setChanged` should be called before `notifyObservers`

Ricket: It should not extend Observable. SideConfig is not a specific type of Observable; if Observable were an interface then yes it should implement the interface, stating that one of its features is that it is observable. But to extend Observable means that it is a class which takes Observable and extends it; whereas in this case, we want to use it. (this is the meaning of "prefer composition over inheritance")

A made-up example of extending Observable would be if you wanted a "SingleObservable", like an observable that could only have one observer.

Yes it does appear setChanged needs to be called, good catch.

I'm not stuck on the use of Observable, I just reached for it for prototyping. It might be better to just implement it from scratch so that we could strong-type the Observer and the data that is passed with the change event.

thiakil: setChanged is protected, hence why I ask if it should extend (and that's what Google seemed to indicate also) 

It's still kinda a specific type of observable tho, no? I imagine the reason it's not an interface though is that it existed pre Java 8 and default methods?

Ricket: Hm, right again. I disagree with extending it personally. I don't disagree with this object having those set of methods, I just disagree with Observable being the superclass. But it seems that is actually the intent from the author of the class. "It can be subclassed to represent an object that the application wants to have observed." I get that the reason `setChanged` is protected is probably so that the observer itself is the only one to decide when it has changed, but I think my explanation from before is valid.

I think the easy thing is to just extend Observable. I think the correct thing is to either have an inner class which subclasses Observable and gives the ability to set changed to the SideConfig class; or to implement the observable pattern from scratch. I'd probably do the latter, personally.

Though at the moment, I'd probably sooner just delete the Observable from the class, since it's not being used.

thiakil: Well now that I think about it there'd be no way to implement the observer list in an interface, so you either have to have a superclass or an internal object with proxy methods (which is rather annoying) 

I'll look into redoing it in a more sensible way with generics.

PR #4562, opened 2017-07-25
Fix issue with side configuration not respecting block orientation

### LINKED ISSUE #4560: [1.11.2] Energy side configuration doesn't respect block orientation
#### Issue description:
Regardless of which way a block is oriented the side configuration for energy always treats the faces as follows:
- Front is North
- Back is South
- Left is West
- Right is East
- Bottom is Down
- Top is Up

Also tested with items but this seems to be functioning as expected.

#### Steps to reproduce:
1. Face north
2. Place an energy cube / machine / quantum entangloporter
3. Configure the block to accept energy on every side
3. Surround the block with universal energy cables
4. Toggle the energy acceptance for each side on/off and make note of which cables detach
5. Repeat the process facing South, East, West, Up, Down as desired

#### Version:
**Forge:** 13.20.1.2393
**Mekanism:** 9.3.3.311-fix

christofferson:
Fixes #4560 

I also noticed during my investigation that `IEnergyWrapper#getConsumingSides()` is no longer used anywhere in the project and `getOutputtingSides()` is only called in `CableUtils#emit(IEnergyWrapper)` so some code cleanup may be needed

### DIFF
--- a/src/main/java/mekanism/common/tile/TileEntityFactory.java
+++ b/src/main/java/mekanism/common/tile/TileEntityFactory.java
@@ -394,6 +394,12 @@ public class TileEntityFactory extends TileEntityMachine implements IComputerInt
 	{
 		return configComponent.getSidesForData(TransmissionType.ENERGY, facing, 1);
 	}
+	
+	@Override
+	public boolean sideIsConsumer(EnumFacing side)
+	{
+		return configComponent.hasSideForData(TransmissionType.ENERGY, facing, 1, side);
+	}
 
 	public void sortInventory()
 	{
--- a/src/main/java/mekanism/common/tile/component/TileComponentConfig.java
+++ b/src/main/java/mekanism/common/tile/component/TileComponentConfig.java
@@ -70,11 +70,11 @@ public class TileComponentConfig implements ITileComponent
 		SideConfig config = getConfig(type);
 		EnumFacing[] translatedFacings = MekanismUtils.getBaseOrientations(facing);
 		
-		for(int i = 0; i < EnumFacing.VALUES.length; i++)
+		for(EnumFacing sideToCheck : EnumFacing.values())
 		{
-			if(config.get(translatedFacings[i]) == dataIndex)
+			if(config.get(translatedFacings[sideToCheck.ordinal()]) == dataIndex)
 			{
-				ret.add(translatedFacings[i]);
+				ret.add(sideToCheck);
 			}
 		}
 		
--- a/src/main/java/mekanism/common/tile/component/TileComponentConfig.java
+++ b/src/main/java/mekanism/common/tile/component/TileComponentConfig.java
@@ -83,7 +83,8 @@ public class TileComponentConfig implements ITileComponent
 
 	public boolean hasSideForData(TransmissionType type, EnumFacing facing, int dataIndex, EnumFacing sideToTest)
 	{
-		return getConfig(type).get(sideToTest) == dataIndex;
+		EnumFacing[] translatedFacings = MekanismUtils.getBaseOrientations(facing);
+		return getConfig(type).get(translatedFacings[sideToTest.ordinal()]) == dataIndex;
 	}
 	
 	public void setCanEject(TransmissionType type, boolean eject)


Ricket:
Thanks for carrying the torch :) and sorry for introducing this issue. I had convinced myself that a base orientation translation wasn't necessary, somehow. Did you test that it works? I haven't quite wrapped my brain around it yet.

christofferson:
I've tested it locally and does appear to have fixed the issue. Essentially whenever the sides are queried, since the configuration only keeps track of the `EnumFacing` relative to how the block was oriented when it was placed (with the center facing the player), that has to be translated into real-world values or else it simply applies the configuration to the cardinal directions.

It took a little searching because I'm not intimately familiar with the code base yet but at least it was a simple fix. :)

Ricket:
I suspect something's still not right. Need to do more testing but here is an energy cube that doesn't seem to be exporting its contents.

image

image

image

Ricket:
Ah I totally broke getSidesForData. My stupidity goes even deeper than the hasSideForData implementation 😢 

It used to loop over each direction in order (D U N S W E), translate it relative to the block's facing direction, then if the translated direction had the desired dataIndex, add the absolute direction to the EnumSet. And then the caller would call contains(sideToTest).

I totally just removed that translation by accident.

christofferson:
Ah, good catch. I was only testing with empty machines / cubes so that may be why I missed it. I'll have a new push ready in a second here.

Ricket:
Thanks. Just change `ret.add(translatedFacings[i]);` to `ret.add(EnumFacing.VALUES[i]);`, right?

christofferson:
I actually switched it to a for-each over `EnumFacing.values()` and replaced `i` with `sideToCheck.ordinal()`

Ricket:
Thanks. Looks good to me 👍 I'll test it shortly and make sure the energy cube is doing its thing.

Ricket:
Energy cube is happy now. LGTM

thommy101:
Will test within a couple hours ;)

thommy101:
Thought it was perfect, still found a small issue. Don't know yet if it happened with the new side change, gonna look into that now.

The bug:
If I disable all energy sides on a factory (smelting and crushing tested) it still accepts energy from any side.

Things that work good:
Energy cubes placed in all kind of directions, setting only the wanted face (front) to on, only connects to the front (way placed). Quantum entangloporters have a correct idea of that is it's front.
Pipes only connect to the corresponding faces.

christofferson:
Nice catch @thommy101! It seems factory tiles were missing the `sideIsConsumer` override and, as such, defaulted to `TileEntityElectricBlock` which always returns `true`

thommy101:
How did I oversee it.. Looked at the code to see if I could fix it :)
*testing*

aidancbrady:
Thanks for your help @christofferson!

Ricket:
Hm. Not sure how I missed that. I used intellij to find all implementations of getConsumingSides (sorry if name is wrong I'm on phone) and then added sideIsConsumer to all the results. I double  checked it, also for outputting sides. Again sorry all. With this being the case, I wonder if there are others I missed then. Maybe we should do a plain grep just to be certain. I can check it later when I'm back at my computer.

thommy101:
So every getConsumingSides should have a corresponding sideIsConsumer? That's easy enough to filter, commandline is almost always open here :)

christofferson:
@thommy101 Tested before and after.

Before all 3 entities definitely had issues accepting energy from any side.

After your patch the pump and plenisher are fixed and the miner seems to be accepting energy from any of the 9 bottom faces or any of the 6 faces to the left and right, is that the expected behavior?

thommy101:
Currently not at my computer, that diff was generated before shutting it down. 
I've copied the way the other methods are working. If someone could test it with https://github.com/aidancbrady/Mekanism/commit/8d55f5996b4f18f579d18cfbf5103817c0ab0218 (commit before the side config change) it would be great.

christofferson:
Tested with the version before any of the performance changes and cables still connect to any face on the left, right or bottom so if we're calling it a bug it's a pre-existing one. Patch looks good to me.

thommy101:
As long as the functionality is the same as before, then I'm happy :)

thommy101:
Thanks for testing @chistofferson!

thommy101:
Created PR with the previous mentioned diff #4567

PR #4565, opened 2017-07-25
Update Transporter Pathfinder

### LINKED ISSUE #4437: [1.11.2] Logistical Transporters try to insert items into bins from bottom (and fail)
Hey there,
I have been building a room with bins as a storage room, connected with logistical transporters to an input-chest. See picture below for setup :D
The lower bins have their inputs connected from the top (as it should be), the upper bins have been disconnected from the logistical transporters that lead to the lower bins with the configurator (set to "NONE") as they are not supposed to output any items via pipe. For the upper bins there is another pipe, leading alongside the top sides of those bins (as intended).
The bug now is, that the pipe tries to fill the bins from the bottom and fail. The bins won't accept items from that side but the pipe is stubborn and tries endless. New items that are sent into the system, will also go the wrong way. For the bins in the bottom row everything works fine. Is it a bug in the pipe-system or in the bins?
Disassembling the corresponding piece of pipe results in itemloss. I've lost 1000 redstones that way because i thought it'd drop into my face when disassembling the buggy piece of pipe.
I could of course add a block in between the upper bins and the pipe for the lower bins to achieve the upper ones not beeing connected to the lower bins' pipes, but the setup I've built worked in older versions, so it should still do?
The chest you can see at the end of the setup normally is not there. Build it for debugging attempts. It does not affect the outcome, still happens without the chest attached.

Kind regards
pflonk

Mekanism-1.11.2-9.3.1.306
forge1.11.2-13.20.0.2312

christofferson:
Fixes #4437

Add checks to TransporterPathfinder to make sure the current node is able to emit to the side(s) we're about to check.

### DIFF
--- a/src/main/java/mekanism/common/content/transporter/TransporterPathfinder.java
+++ b/src/main/java/mekanism/common/content/transporter/TransporterPathfinder.java
@@ -446,6 +446,9 @@ public final class TransporterPathfinder
 			}
 
 			double maxSearchDistance = start.distanceTo(finalNode) * 2;
+			ArrayList<EnumFacing> directionsToCheck = new ArrayList<>();
+			Coord4D[] neighbors = new Coord4D[EnumFacing.VALUES.length];
+			TileEntity[] neighborEntities = new TileEntity[neighbors.length];
 
 			while(!openSet.isEmpty())
 			{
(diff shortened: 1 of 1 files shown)


aidancbrady:
Well done, thanks!

Issue #4503, opened 2017-06-29
[Feature] Quantum Entangloporter - Keep frequency when broken


Lukas45:
What title says.
When you break a Quantum Entangloporter, it loses the Frequency it's set to.
I'd like to request that it keeps the Frequency it's set to, as you're very likely to use the same frequency (usually power) when putting it down, and if not, you have to open it anyways and nothing changes.
Also since side config is saved on the QE, different item/gas/fluid channels can very likely not be used without changing them anyways, so it would make sense to keep using the same Frequency.
Just my 2 cents of why it's a worthy QOL update ;)

Maybe this is already the case in 1.11.2 though (I'm still at 1.10.2)

Issue #3953, opened 2016-11-07
Quantum Entangleporter Max output is 256mb/t


Saereth:
MC 1.10.2 
FTB Lite 1.10.2 pack
forge 2121

Currently going from a drum of water->ultimate mechanical pipes->entangleporter and then outputting to another drum from a shared frequency entangleporter only provides a throughput of 256mb/t. I thought this was meant to have unlimited transfer rate? 

For whatever reason it looks like the value is defaulting to this min here:
https://github.com/aidancbrady/Mekanism/blob/81e810794581cabba8925634f93f1323400014cb/src/main/java/mekanism/common/tile/component/TileComponentEjector.java#L142

This also seems to be the same for the energy transfer rate (internal buffer capping at 400krf which makes each frequency cap out at 400krf/t transfer rate).

Issue #3846, opened 2016-09-29
[feature] allow anchor upgrades in quantum entangloporters


rapus95:
the best example would be a quantum entangloporter & a pump in the nether that shall bring lava into the overworld. Problem: the nether station usually is unloaded except i place a teleporter or a DM next to it.
Sure, a teleporter might even make sense but you don't want to be forced to use one :)

SuperFanatic:
You could always make a basic machine, put an anchor upgrade in it, and place it next to the entangloporter. it doesn't need to be running to be anchoring.

thiakil:
BTW: I'm fairly certain you don't need the full teleporter structure. Obviously it would only be inbound from the portable tele, but it is a solution.

JustinHuPrime:
This has been fixed for a while - shouldn't it be taken off the feature requests project?

PR #4601, opened 2017-08-06
Entangleporter Improvements

thiakil:
Implements #4503, #3846, and half of what #3953 wants.

As for #3953 I've not changed the amount by which tiles emit to pipes, but have increased the buffer capacity to the same as the ultimate tiers of cable/pipe/tube can handle, so in theory transmitters can *pull* at their full rate, on one side of the porter at least.

Frequency storage in the itemstack stores only the name & public/private flag, so private frequencies should attach themselves to the placing player (e.g. should it be different to the frequency's).

Anchor upgrade is copied from the teleporter - loads just the tile's chunk.


aidancbrady:
Great work, thanks!

PR #4628, opened 2017-08-14
Use access transformer instead of reflection

thiakil:
Replaces the usages of `ReflectionUtils` with an access transformer.

Mostly cosmetic with a possible tiny performance improvement.

This will probably conflict trying to merge with 1.12, and will need the srgNames double checked


thommy101:
Nice, this was also a thing on my todo list! nice to see it :)

thommy101:
1.12 should have the same srg names, otherwise it would have errored already ;)

thiakil:
well, only if there was no null handling. I know IDEA warned about unboxing potentially causing it in a couple of places. This way we'll know for sure if they're correct or not

PR #4641, opened 2017-08-18
Add CraftTweaker integration

thommy101:
Optional dependency on MTLib to make CraftTweaker integration easier

Thanks to @jaredlll08 for letting me use the "old" ModTweaker code (which was almost 100% compatible)

### DIFF
--- a/src/main/java/mekanism/common/integration/crafttweaker/CrafttweakerIntegration.java
+++ b/src/main/java/mekanism/common/integration/crafttweaker/CrafttweakerIntegration.java
@@ -1,0 +1,26 @@
+package mekanism.common.integration.crafttweaker;
+
+import crafttweaker.CraftTweakerAPI;
+import crafttweaker.IAction;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class CrafttweakerIntegration
+{
+    public static final List<IAction> LATE_REMOVALS = new LinkedList<>();
+    public static final List<IAction> LATE_ADDITIONS = new LinkedList<>();
+
+    /**
+     * Apply after (machine)recipes have been applied, but before the FMLLoadCompleteEvent is fired.
+     * Preferbly in (post)init.
+     *
+     * Applying to early causes remove to mallfunction as no recipes have been registered.
+     * Applying to late causes JEI to not pickup the changes.
+     */
+    public static void apply()
+    {
+        LATE_REMOVALS.forEach(CraftTweakerAPI::apply);
+        LATE_ADDITIONS.forEach(CraftTweakerAPI::apply);
+    }
+}
--- a/src/main/java/mekanism/common/integration/crafttweaker/CrafttweakerIntegration.java
+++ b/src/main/java/mekanism/common/integration/crafttweaker/CrafttweakerIntegration.java
@@ -18,7 +18,7 @@ public class CrafttweakerIntegration
      * Applying to early causes remove to mallfunction as no recipes have been registered.
      * Applying to late causes JEI to not pickup the changes.
      */
-    public static void apply()
+    public static void applyRecipeChanges()
     {
         LATE_REMOVALS.forEach(CraftTweakerAPI::apply);
         LATE_ADDITIONS.forEach(CraftTweakerAPI::apply);
--- a/build.gradle
+++ b/build.gradle
@@ -73,6 +73,10 @@ repositories {
         name 'ic2'
         url 'http://maven.ic2.player.to'
     }
+    maven {
+        name 'BlameJared - Crafttweaker'
+        url 'http://maven.blamejared.com/'
+    }
 }
 
 // Define properties file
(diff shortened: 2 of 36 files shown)


thommy101:
Tested and everything seems to be working!

Tested with the test/example file from the ModTweaker wiki. (Attached as .zs.txt as Github doesn't allow .zs files)
[mekanism.zs.txt](https://github.com/aidancbrady/Mekanism/files/1236485/mekanism.zs.txt)

src/main/java/mekanism/common/integration/crafttweaker/CrafttweakerIntegration.java
```diff
@@ -0,0 +1,26 @@
...
+    public static final List<IAction> LATE_ADDITIONS = new LinkedList<>();
+
+    /**
+     * Apply after (machine)recipes have been applied, but before the FMLLoadCompleteEvent is fired.
+     * Preferbly in (post)init.
+     *
+     * Applying to early causes remove to mallfunction as no recipes have been registered.
+     * Applying to late causes JEI to not pickup the changes.
+     */
+    public static void apply()
```
thiakil: Perhaps it would be better to call it `applyRecipeChanges` for now? Mostly just for some clarity in MekanismHooks without having to look at this file

thommy101: That does indeed look better :) Changed

josephcsible:
Does this not support the ore dictionary? This line:
```
mods.mekanism.combiner.addRecipe(<ore:dustDiamond> * 4, <gas:liquidStone>, <minecraft:diamond_ore>);
```
results in this error:
```
test.zs:1 > 2 methods available but none matches the parameters (crafttweaker.item.IIngredient, mod.mekanism.gas.IGasStack, crafttweaker.item.IItemStack)
This is usually an error in your script, not in the mod
addRecipe(ZenTypeNative, ZenTypeNative )
addRecipe(ZenTypeNative, ZenTypeNative, ZenTypeNative )]
```
Here's what I have to use instead:
```
mods.mekanism.combiner.addRecipe(<mekanism:otherdust:0> * 4, <gas:liquidStone>, <minecraft:diamond_ore>);
```

thommy101:
You might be right with that one.. I didn't think of that. I don't even know if the underlying code supports OreDict. 

I'll take a look at it when I'm back in front of my computer.

josephcsible:
Also, using `:*` to match all variants of an item doesn't seem to work. For example:
```
mods.mekanism.reaction.addRecipe(<minecraft:sand:*>, <liquid:liquidsodium>, <gas:water>, <minecraft:quartz>, <gas:hydrogen>, 50, 20);
```
I don't get any errors from that, and the recipe shows up fine in JEI, but when I actually try it in a machine, it doesn't do anything.

thommy101:
Thank you very much for testing this @josephcsible! It helps a lot.

thommy101:
ignore my last commit... doesn't work yet

thommy101:
Added wildcard support to all recipes (looks like it normal itemstack inputs already had it).
Always good to take a second look some time later, this fix is way easier than what I saw earlier.

Oredict support will be postponed as it would require a rewrite of the current recipe system. I'll open a new feature request so interested parties could subscribe to that one.

@josephcsible Your example is working perfectly now, it accepts both sand & red sand. If you find any more issues (not oredict related ofcourse), feel free to share them :)

josephcsible:
Thanks! As an interim solution to the oredict issue, would it be possible for you to make it give an error such as ”Mekanism recipes don't yet support the ore dictionary” rather than the cryptic one that I got?

thommy101:
The error you've received is an error from CraftTweaker notifying you about what you've entered isn't possible..
Only way to give a warning/error would be creating methods that do accept oredict and logging a big warning that it's not yet supported. I can look into it, but if it's a lot of work than I would rather invest my time in a bit of documentation about current CraftTweaker integration and a warning on top of that that oredict isn't supported yet.

I've opened a issue for the full support of oredict at #4646

thiakil:
> creating methods that do accept oredict and logging a big warning that it's not yet supported.

At what point in the cycle is that called? The simple solution for now would be to bake the list by querying the OreDict in init or postinit

thommy101:
That would be a huge and ugly hack, but very plausible. Methods are queried at post init after all (or at least most) OreDict should be registered. 

The apply method, which does the actual addition and removal of recipes is called from MekanismHooks post init ;)

thiakil:
Well I wouldn't call it a *hack* personally, but I certainly agree it's a *suboptimal* workaround :P

thommy101:
I've looked at providing an warning when a oredict recipe is being used, and decided to certainly not go that way. It does provide a nice warning, but you'll loose line numbers, which is in my opinion more important.

So it will be either go the way it's now and wait for a nice oredict implementation of all recipes, or as @thiakil said, go for the *suboptimal* workaround.

WTFNoNamesLeftoO:
Seeing that mekanism doesn't have MT integration in 1.12 yet - when will we be able to lay hands on a version that does have support for MT?

Thanks for working on it! Cheers!

thommy101:
Once this pull request is accepted, I think within the next days or so, then I've got tome to complete this :)

Thinking of combining #4648 into this one

josephcsible:
@WTFNoNamesLeftoO Whenever this gets merged, the next release will have it. If you're impatient, you can compile it yourself with this PR's changes included (that's what I did to test it).

PR #4666, opened 2017-09-02
Fixed gas fluid not being gaseous.

halvors:
Hi!

I ran into some compatibility issues with some other mods that expects gas fluids to be gaseous.
I suspect that this is a detail that was overlooked when migrating to the new forge fluid system.

Fluids that have the same name is usually not added twice, but instead mods use the first one registered, when other mods expect gases to be set to gaseous issues might occur.

Please backport this as well if possible to 1.10 and 1.11 as many modders is still one these versions.
I wasn't sure how to submit this PR to all the branches at once, but as this is a 1 line fix it should be fairly easy to backport. Code works for all 1.10+ versions of minecraft. 

Thanks.


thiakil:
Those fluids are not gasses as far as I can see. They are the Rotary Condensentratored liquids.

See `mekanism.common.MekanismFluids#register`

halvors:
@thiakil Isn't steam a gas? Strange that the class is called Gas otherwise.

thiakil:
Steam does not use the `Gas` class in Mekanism, and is already set to `Gaseous`

halvors:
Hmm. Ok i understand. My bad. Just confusing that mekanism is using the gas system for liquids?...

thiakil:
Not sure what you mean? Mekanism gasses can be converted to a liquid form using a Rotary Condensentrator. These liquids are standard MC Fluids.

PR #4709, opened 2017-09-24
Fix several generics type-safety issues

josephcsible:


src/api/java/dan200/computercraft/api/ComputerCraftAPI.java
```diff
@@ -117,7 +117,7 @@ public static IWritableMount createSaveDirMount( World world, String subPath, lo
 	 * @see dan200.computercraft.api.peripheral.IComputerAccess#mountWritable(String, IWritableMount)
 	 * @see dan200.computercraft.api.filesystem.IMount
 	 */
-	public static IMount createResourceMount( Class modClass, String domain, String subPath )
+	public static IMount createResourceMount( Class<?> modClass, String domain, String subPath )
```
thiakil: Please don't change things in another mod's api

src/main/java/mekanism/common/PacketHandler.java
```diff
@@ -207,7 +207,7 @@ else if(data instanceof byte[])
 				}
 				else if(data instanceof ArrayList)
 				{
-					encode(((ArrayList)data).toArray(), output);
+					encode(((ArrayList<?>)data).toArray(), output);
```
thiakil: I don't see what benefit all these `<?>` calls add...

src/main/java/mekanism/common/block/property/PropertyConnection.java
```diff
@@ -5,7 +5,7 @@
 import mekanism.common.tile.transmitter.TileEntitySidedPipe.ConnectionType;
 import net.minecraftforge.common.property.IUnlistedProperty;
 
-public class PropertyConnection implements IUnlistedProperty<PropertyConnection>
+public final class PropertyConnection implements IUnlistedProperty<PropertyConnection>
```
thiakil: Why `final`?

src/api/java/dan200/computercraft/api/ComputerCraftAPI.java
josephcsible: Oops, I didn't catch that this wasn't our API. I'll change this one back.

src/main/java/mekanism/common/PacketHandler.java
josephcsible: It lets the compiler know that we explicitly don't know what the type parameter is, so it knows we're using the type safely and not doing anything with it that might unexpectedly throw a ClassCastException. (If you enable warnings for raw types and unchecked casts, you'll see them go down a lot after this change.)

src/main/java/mekanism/common/block/property/PropertyConnection.java
josephcsible: Looking more closely, I guess it's not strictly necessary. I can take it out if you want.

src/main/java/mekanism/common/PacketHandler.java
thiakil: In this particular case though, IDEA at least doesn't seem to care - we're not even doing it 'safely' tbh

src/main/java/mekanism/common/block/property/PropertyConnection.java
thiakil: If you have a god reason for it we can do it, but finality for finality's sake is just no.

josephcsible: My reasoning is that PropertyConnection is an IUnlistedProperty of itself, which is unusual. It's more common to see, for example, PropertyFloat being an IUnlistedProperty of Float. If someone were to subclass this, they wouldn't be able to make getType() return the right thing anymore, since it's defined to return a `Class<V>`, not a `Class<? extends V>`.

src/main/java/mekanism/common/PacketHandler.java
josephcsible: Eclipse and javac both do care when the right set of warnings are on, and technically this way is safe (at least locally).

src/main/java/mekanism/common/block/property/PropertyConnection.java
thiakil: Not really unusual, why have 2 classes, when one will do.

`getType` should return the subclass' `Class`, and generics don't matter at runtime

josephcsible: getType can't return the subclass's Class, though, because the subclass would still be an IUnlistedProperty<PropertyConnection>, so it would still have to return a Class<PropertyConnection>, which a subclass's class isn't. And although generics are erased at runtime, doing unsafe things could still result in a ClassCastException at runtime.

src/api/java/dan200/computercraft/api/ComputerCraftAPI.java
josephcsible: I notice that this is already fixed upstream. Can we update our copy of their API?

src/main/java/mekanism/common/block/property/PropertyConnection.java
thiakil: well, actually having looked further, were this only a container class (ie not a property of itself) that would make sense.

It currently has no methods, just properties, but you could subclass it to have some and still put it in the same `IUnlistedProperty` storage in the blockstate, no?

src/api/java/dan200/computercraft/api/ComputerCraftAPI.java
thiakil: could, but not in this PR

thiakil: Ideally they should have a maven, or we add an api jar as a lib

thommy101:
Thanks @josephcsible for the work on this! Much appreciated! Help is always welcome

Will look at this today ;)

thommy101:
You've also got your other PR in this one? (Execution bit set on gradlew)

josephcsible:
> You've also got your other PR in this one? (Execution bit set on gradlew)

Oops, I needed that to even try running this one, and I forgot to remove it.

src/api/java/dan200/computercraft/api/ComputerCraftAPI.java
josephcsible: Okay, I'll make a separate PR for this.

src/main/java/mekanism/common/block/property/PropertyConnection.java
josephcsible: I guess so. I'll remove `final` for now.

josephcsible:
> Most changes look good, I do however agree with @thiakil with the <?>. Don't know if that would solve warnings as I don't get them in IDEA and Code analysis tools.

Are you sure you have warnings turned on for raw types and unchecked casts?

thommy101:
Pretty sure I've got that enabled ;) 34e81bb65d49e11d228b1335996d63f2912e9d62

josephcsible:
Odd that you're not seeing them then. Eclipse shows them to me.

Mekanism_Client.launch
```diff
@@ -1 +1 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?><launchConfiguration type="org.eclipse.jdt.launching.localJavaApplication"><stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value="GradleStart"/><stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="Mekanism"/><stringAttribute key="org.eclipse.jdt.launching.WORKING_DIRECTORY" value="/Users/aidancbrady/Desktop/Minecraft Development/Projects/Mekanism/run"/><stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value=" -DFORGE_FORCE_FRAME_RECALC=true"/></launchConfiguration>
\ No newline at end of file
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><launchConfiguration type="org.eclipse.jdt.launching.localJavaApplication"><stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value="GradleStart"/><stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="Mekanism"/><stringAttribute key="org.eclipse.jdt.launching.WORKING_DIRECTORY" value="/home/joe/MinecraftMods/Mekanism/run"/><stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value=" -DFORGE_FORCE_FRAME_RECALC=true"/></launchConfiguration>
```
thiakil: can you please undo these `.launch` changes? The files really shouldn't be in git in the first place (I'll remove them on the main branch)

josephcsible: Oops, done. (And +1 to removing them)

PR #4711, opened 2017-09-24
Delete old recipe code

josephcsible:
This is all unused in 1.12 in favor of the new .json system.


thiakil:
Nope, this is there for a reason. The JSON is created by a conversion 'script'

josephcsible:
Wait, the script wasn't a one-shot deal? Why would we re-run it now that we already have the JSONs?

thiakil:
Some of them are created in loops etc. Not everyone understands the new format. Creates less issues if we backport recipe changes.

Tom's call really.

thommy101:
I do prefer leaving it in. It is indeed unused but it makes it a lot easier if we want to add some recipes. It's easier to add one rule and let it generate a bunch of json files than to write that file by hand, which would probably contain some faults. This way we have all recipes defined in one place instead of a couple hundred files.

PR #4713, opened 2017-09-24
Nuggets!

josephcsible:
- Add a nugget corresponding to each ingot that we add
- Add the usual crafting recipes between ingots and nuggets
- Make MekanismTools items smelt into nuggets like vanilla items do


thommy101:
What I see from mobile it looks damn good! Thank you!

thiakil:
I'm not sure I like the nuggets & smelting for non-metals, but that's not my call.

aidancbrady:
Love this

PR #4716, opened 2017-09-25
Fix #4710: remove the removeAll function

### LINKED ISSUE #4710: FrequencyManager.removeAll doesn't work as intended
https://github.com/aidancbrady/Mekanism/blob/89e10b1d9d0399ca821c30db3f416638522ee723/src/main/java/mekanism/common/frequency/FrequencyManager.java#L121

This is seeing if a `UUID` equals a `String`, which will always be false, since they're different types. That leads to this whole method effectively becoming a no-op. (I'm not sure whether `user` is supposed to be a UUID, or if it should be checking `iterFreq.name` instead of `iterFreq.ownerUUID`, and I can't find the code that calls this method to check there.)

josephcsible:
This function is unused, and it never worked right anyway due to a bug.

### DIFF
--- a/src/main/java/mekanism/common/frequency/FrequencyManager.java
+++ b/src/main/java/mekanism/common/frequency/FrequencyManager.java
@@ -110,25 +110,6 @@ public class FrequencyManager
 		}
 	}
 	
-	public int removeAll(String user)
-	{
-		int amount = 0;
-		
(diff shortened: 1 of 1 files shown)


thiakil:
> and it never worked right anyway
It did, but wasn't converted when we switched to UUIDs as the command that used it got removed by Aidan ;)

PR #4719, opened 2017-09-26
Update to the latest vesion of the ComputerCraft API (19e4c03)

josephcsible:


thiakil:
As mentioned in the other PR, I'd really prefer this was added as a gradle dep, or as a compiled jar in a libs folder at minimum (we've moved everything else we can to gradle). Likely requires work on CC's end or a custom build.

josephcsible:
> Likely requires work on CC's end or a custom build.

Yeah, they don't offer a repository or even a jar right now. I opened dan200/ComputerCraft#446 about it. Can we do this for now, then switch to a gradle dep or a jar once they get one added?

thiakil:
Does the integration work without this PR?

aidancbrady:
Any updates on this?

thiakil:
doesn't look like we got any progress on CC's side. I can generate an API jar for us this weekend though if we wanna go that way

aidancbrady:
@thiakil, that sounds like a good option for now. Let me know once it's ready and one of us can add it in as a maven dep.

thiakil:
added on my maven server and our 1.12 branch, changes available on https://github.com/thiakil/ComputerCraft/tree/api-jar if wanted to host elsewhere

PR #4733, opened 2017-10-03
Fix #4648: Let the combiner combine with things other than cobblestone

### LINKED ISSUE #4648: Feature request: Allow non-cobblestone in the combiner
Currently, the combiner is hardcoded to only combine things with cobblestone, so it's not possible, for example, to add a recipe for it to combine dusts with netherrack to make nether ores. Can its recipe system be updated to support things other than cobblestone, so that other mods (and the upcoming CraftTweaker integration) can add these sorts of recipes?

josephcsible:
This lets us combine netherrack with quartz to make nether quartz ore now, and can be extended to combine any two arbitrary items. Note the following two (rather minor) breaking changes:
1. Cobblestone used to be consumed at the beginning of work. It's now consumed at the end. Any combiners that were running prior to this update will lose the single piece of cobblestone that they had consumed but not yet finished using.
2. Any CraftTweaker scripts that used the three-argument form of `mods.mekanism.combiner.addRecipe` (passing `<gas:liquidStone>` as the second argument) will break.

### DIFF
--- a/src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
+++ b/src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
@@ -1,20 +1,26 @@
 package mekanism.common.recipe.machines;
 
-import mekanism.common.MekanismFluids;
-import mekanism.common.recipe.inputs.AdvancedMachineInput;
+import mekanism.common.recipe.inputs.DoubleMachineInput;
 import mekanism.common.recipe.outputs.ItemStackOutput;
+import net.minecraft.init.Blocks;
 import net.minecraft.item.ItemStack;
 
-public class CombinerRecipe extends AdvancedMachineRecipe<CombinerRecipe>
+public class CombinerRecipe extends DoubleMachineRecipe<CombinerRecipe>
 {
-	public CombinerRecipe(AdvancedMachineInput input, ItemStackOutput output)
+	public CombinerRecipe(DoubleMachineInput input, ItemStackOutput output)
 	{
 		super(input, output);
 	}
 
+	public CombinerRecipe(ItemStack input, ItemStack extra, ItemStack output)
+	{
+		super(input, extra, output);
+	}
+
+	@Deprecated
 	public CombinerRecipe(ItemStack input, ItemStack output)
 	{
-		super(input, MekanismFluids.LiquidStone, output);
+		super(input, new ItemStack(Blocks.COBBLESTONE), output);
 	}
 
 	@Override
--- a/src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
+++ b/src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
@@ -17,7 +17,11 @@ public class CombinerRecipe extends DoubleMachineRecipe<CombinerRecipe>
 		super(input, extra, output);
 	}
 
-	@Deprecated // TODO remove this method in MC 1.13
+	/**
+	 * @deprecated Replaced by {@link #CombinerRecipe(ItemStack, ItemStack, ItemStack)}.
+	 * May be removed with Minecraft 1.13.
+	 */
+	@Deprecated
 	public CombinerRecipe(ItemStack input, ItemStack output)
 	{
 		super(input, new ItemStack(Blocks.COBBLESTONE), output);
--- a/src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
+++ b/src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
@@ -17,7 +17,7 @@ public class CombinerRecipe extends DoubleMachineRecipe<CombinerRecipe>
 		super(input, extra, output);
 	}
 
-	@Deprecated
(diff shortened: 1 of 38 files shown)


src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
```diff
@@ -1,20 +1,26 @@
...
 	{
 		super(input, output);
 	}
 
+	public CombinerRecipe(ItemStack input, ItemStack extra, ItemStack output)
+	{
+		super(input, extra, output);
+	}
+
+	@Deprecated // TODO remove this method in MC 1.13
```
thiakil: Why remove it? That would be a convenience method for the default.

thiakil:
Iis there a particular reason behind a new recipe/TE/container/gui type that exists to only be extended by the combiner? The combiner already has its own `RecipeType`

src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
josephcsible: It seems rather arbitrary for cobblestone to be the default, and there's no way we'd add this method if we were building the combiner like this from scratch today.

josephcsible:
I imagine it could be useful in future machines, and there's already precedent for it: ChanceMachineRecipe is only extended by SawmillRecipe.

src/main/java/mekanism/common/tile/prefab/TileEntityDoubleElectricMachine.java
```diff
@@ -0,0 +1,287 @@
...
+	 *
+	 * @param soundPath - location of the sound effect
+	 * @param name - full name of this machine
+	 * @param maxEnergy - maximum amount of energy this machine can hold.
+	 * @param baseEnergyUsage - how much energy this machine uses per tick.
+	 * @param ticksRequired - how many ticks it takes to smelt an item.
+	 */
+	public TileEntityDoubleElectricMachine(String soundPath, String name, double maxEnergy, double baseEnergyUsage, int ticksRequired)
+	{
+		super(soundPath, name, maxEnergy, baseEnergyUsage, 4, ticksRequired, MekanismUtils.getResource(ResourceType.GUI, "guibasicmachine.png"));
```
josephcsible: On the combiner today, this is [guiadvancedmachine.png](https://github.com/aidancbrady/Mekanism/blob/89e10b1d9d0399ca821c30db3f416638522ee723/src/main/resources/assets/mekanism/gui/guiadvancedmachine.png), which has a bar between the cobblestone and dust slots that showed the liquid stone level. I'm not sure whether it'd be better to change that to [guibasicmachine.png](https://github.com/aidancbrady/Mekanism/blob/89e10b1d9d0399ca821c30db3f416638522ee723/src/main/resources/assets/mekanism/gui/guibasicmachine.png) as I did here, which replaces the bar with an arrow pointing up from the cobblestone to the dust, or to [guiblank.png](https://github.com/aidancbrady/Mekanism/blob/89e10b1d9d0399ca821c30db3f416638522ee723/src/main/resources/assets/mekanism/gui/guiblank.png), which would just remove the bar.

josephcsible:
I also wonder whether we should make each ”lane” of the factories have their own cobblestone/netherrack/whatever slot, instead of all sharing one, so the same factory could do more than one kind of combining at once (although I don't think the infusion factory supports this either, so I'm not overly concerned right now).

src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
thiakil: > arbitrary 

I disagree, given that 99% of ores are made from cobble/stone

src/main/java/mekanism/common/tile/prefab/TileEntityDoubleElectricMachine.java
thiakil: yeah `basicmachine` should be fine

thiakil:
It would probably get too messy if you added a slot per lane. You will need to account for that specially if you're consuming the item upon completion of course

src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
josephcsible: We don't have convenience methods for any other recipes, though. It just doesn't seem natural to have one here. (But if this is really a big deal, I am willing to un-deprecate them to get this merged though.)

josephcsible:
Okay, I think it's good for now with just the one slot. (And we can always change that later if we really want.)

src/main/java/mekanism/common/recipe/machines/CombinerRecipe.java
thiakil: I'm fine with you suggesting it, but putting it in without mentioning it comes across as "we must do this"

To me this constructor defines what the default is

josephcsible: Fair point; I'll reword the comment a bit.

Issue #4743, opened 2017-10-06
Seismic Reader - add a description (SHIFT + M)


EliogabalusPL:
*Please use the search functionality before reporting an issue. Also take a look at the closed issues!*

#### Issue description:
The Seismic Vibrator is based on the SHIFT + M keypad, but unfortunately most users try to use Seismic Reader first, and they have no idea how this device works. It's easy for English-speaking gamers to use the module's Web site. It would be sufficient to add a description of the Seismic Reader when you press SHIFT + M

#### Steps to reproduce:
1. Add a description of Seismic Reader by pressing SHIFT + M on it
2.
3.

#### Version (make sure you are on the latest version before reporting):
**Forge: 1.12.1
**Mekanism:* 1.12.1
**Other relevant version: additional modules Mekanism

#### If a (crash)log is relevant for this issue, link it here: (_It's almost always relevant_)

[gist/pastebin/etc link here]

thommy101:
Created a PR to add some information to the tooltip ;)

PR #4748, opened 2017-10-07
Add tooltip information to Seismic Reader

thommy101:
Added some information about usage to the Seismic Reader, almost the same as the Seismic Vibrator.
Adds feature #4743

Opened as PR to make sure the used wording for the tooltip is right ;)

### DIFF
--- a/src/main/resources/assets/mekanism/lang/en_us.lang
+++ b/src/main/resources/assets/mekanism/lang/en_us.lang
@@ -805,6 +805,7 @@ tooltip.flamethrower.modeBump=Flamethrower mode bumped to
 
 tooltip.seismicReader.needsEnergy=Not enough energy to interpret vibration
 tooltip.seismicReader.noVibrations=Unable to discover any vibrations
+tooltip.mekanism.SeismicReader=A portable machine that uses seismic vibrations to provide information on differing layers of the world.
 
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/mekanism/lang/en_us.lang
```diff
@@ -805,6 +805,7 @@ tooltip.flamethrower.modeBump=Flamethrower mode bumped to
 
 tooltip.seismicReader.needsEnergy=Not enough energy to interpret vibration
 tooltip.seismicReader.noVibrations=Unable to discover any vibrations
+tooltip.SeismicReader=A portable machine that uses seismic vibrations to provide information on differing layers of the world.
```
thiakil: English is fine, but while we're at it, shouldn't we make it `tooltip.mekanism.SeismicReader`?

thommy101: I've got a separate branch which will update all translations at once ;) Will update that later today

thiakil: I know, but would it not be easier to add the new one with the new schema already?

thommy101: Just looked, I apparently have updated tooltips already.. Will change this here ;)

PR #4801, opened 2017-11-07
Wrench compat

thiakil:
Adds back COFH hammer to the Configurator & adds generic (and extensible) 3rd party wrench handling.


aidancbrady:
Awesome!

PR #4803, opened 2017-11-08
Api update

thiakil:
Some API updates that the MMD guys wanted.

Seeking feedback on the recipe stuff; IMC is good, but seems undocumented and people don't like it. Implemented a recipe helper which does a basic sanity check.


thommy101:
It's certainly not undocumented ;)
https://github.com/aidancbrady/Mekanism/wiki/IMC-based-recipe-integration

thiakil:
Huh, even I didn't know that existed :P

Could use with listing the exact recipe keys though, nobody wants to go code diving or guess those xD

PR #4804, opened 2017-11-08
Nonnull network data

thiakil:
Rather large refactor that essentially changes `ITileNetwork.getNetworkedData(ArrayList<Object>)` to `ITileNetwork.getNetworkedData(TileNetworkList)` which is a `NonNullList` subclass.

This will allow us to easily find when & where nulls are being written to the data list, before it hits the encode phase.


aidancbrady:
Love it!

PR #4833, opened 2017-11-28
This is never this simple, what am I missing?

drunkripper:
I'm hoping this fixes it, but when are these things actually this easy?

### DIFF
--- a/src/main/java/mekanism/common/tile/prefab/TileEntityElectricBlock.java
+++ b/src/main/java/mekanism/common/tile/prefab/TileEntityElectricBlock.java
@@ -81,13 +81,14 @@ public abstract class TileEntityElectricBlock extends TileEntityContainerBlock i
 	@Method(modid = "IC2")
 	public void deregister()
 	{
-		if(!worldObj.isRemote)
+		if(!worldObj.isRemote && ic2Registered)
 		{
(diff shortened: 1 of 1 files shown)


thiakil:
That is indeed all I changed in the 1.12 branch. Make sure you actually test it though.

drunkripper:
How do I build it though with Gradle?

tenten8401:
99% of the time it's just a `./gradlew build` as I recall.
drunkripper:patch-1 built successfully, I'll see if I can test it out.

tenten8401:
I just tested this on a server affected with the issue and it does not seem to fix the issue.
image
The CPU spikes can still be associated with the IC2 EnergyNet removal:
image

thiakil:
I don't see how the CPU usage is relevant... that's not something we can control.

tenten8401:
It's causing chunks to be loaded that shouldn't be loaded and therefore
increasing the CPU utilization to levels that really interfere with
gameplay.

thiakil:
Then you should be reporting that to IC2. Registration with ENET is just an event firing.

tenten8401:
However I do not believe it's being registered properly with IC2 in the
first place though.

thiakil:
the boolean is only set after the event has been sent :man_shrugging:

PR #4868, opened 2017-12-16
Make electrolytic separator not so efficient.

### LINKED ISSUE #4852: [1.12] Electrolytic Separator + Energy Upgrades = Hydrogen Power
#### Issue description:
The Electrolytic Separator accepts Energy Upgrades.  In the past, this only increased the energy capacity.  The Electrolytic Separator was the only machine to not receive a boost in efficiency in order to prevent an infinite energy loop using Hydrogen Gas.

The Electrolytic Separator currently does get the energy boost and can produce an infinite energy loop using Hydrogen Gas.

#### Steps to reproduce:
1. Stick a full set of Energy Upgrades in the Electrolytic Separator.
2. Supply Water
3. Burn Hydrogen in a Gas-Burning Generator for 10x the power you put in.  A full set of Speed Upgrades will net you nearly 18k J/t.

**Forge: 14.23.1.2561**
**Mekanism: 9.4.2.327**

Test pack included Mek Generators and Tools, and JEI 4.8.5.134.  No other mods.

thommy101:
Force the energy usage to be a constant and not modified by any electrical upgrades.
Fixes #4852

### DIFF
--- a/src/main/java/mekanism/common/tile/TileEntityElectrolyticSeparator.java
+++ b/src/main/java/mekanism/common/tile/TileEntityElectrolyticSeparator.java
@@ -646,4 +646,21 @@ public class TileEntityElectrolyticSeparator extends TileEntityMachine implement
 	{
 		return new Object[] {fluidTank, leftTank, rightTank};
 	}
+
+	@Override
+	public void recalculateUpgradables(Upgrade upgrade)
+	{
+		super.recalculateUpgradables(upgrade);
+
+		switch(upgrade)
+		{
+			case ENERGY:
+				maxEnergy = MekanismUtils.getMaxEnergy(this, BASE_MAX_ENERGY);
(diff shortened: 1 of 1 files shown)


src/main/java/mekanism/common/tile/TileEntityElectrolyticSeparator.java
```diff
@@ -646,4 +646,21 @@ public boolean hasCapability(Capability<?> capability, EnumFacing side)
 	{
 		return new Object[] {fluidTank, leftTank, rightTank};
 	}
+
+	@Override
+	public void recalculateUpgradables(Upgrade upgrade)
+	{
+		super.recalculateUpgradables(upgrade);
```
thiakil: Wouldnt it be better to have this in the `default` block, to save doing any useless calcs of energy use?

Issue #4814, opened 2017-11-15
[1.12.2] Coal/charcoal blocks don’t work in the metallurgic infuser


ghost:
Redstone blocks work, so why not coal blocks (and charcoal blocks, which some mods add).

**Mekanism version**: 9.4.1.326

PR #4869, opened 2017-12-16
Add (char)coal blocks to infuser.

thommy101:
Adds feature #4814.

Opened as PR to discuss used values.


thiakil:
What basis did you use for the values?

Coal block should be 9x coal item

Charcoal, I'm unsure. Half coal values? Same?

thommy101:
Based on coal & charcoal, which are in the two lines above them ;)

thiakil:
Ah late night comprehension fail of the coal item meta lol

I would have thought it should be the other way round in terms of coal/charcoal carbon levels... But i'm no chemist xD

PR #4876, opened 2017-12-20
Add Jenkinsfile

Mastergalen:
Allows the project to use Jenkin's [declarative pipeline syntax](https://jenkins.io/doc/book/pipeline/syntax/).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,33 @@
+pipeline {
+  agent any
+
+  tools {
+    gradle "Gradle 4.4"
+  }
+
+  stages {
+    stage('Build') {
+      steps {
+        sh "gradle"
+      }
+    }
+    stage('Deploy') {
+      when {
+        branch 'master'
+      }
+      steps {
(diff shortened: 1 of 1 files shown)


thiakil:
First of all, why? AFAIK auto builds are already setup behind the scenes...

Second of all, where is there an api key in the file?!

thommy101:
I know a bit of S3 storage, and that isn't an api key ;)
It's a Jenkins user which contains the 2 secrets needed.
https://github.com/jenkinsci/pipeline-aws-plugin#withaws

This is either something @aidancbrady requested, or something that's unwanted.

Also, this should target development, not master

aidancbrady:
We're moving site-specific deployment to AWS S3 instead of the raw webserver. Definitely intentional!

thiakil:
@Mastergalen the jenkins server seems to be using a still old version of gradle, despite 4.4 being specified in the jenkinsfile

It could also potentially use the gradle wrapper script (`./gradlew`)

PR #4889, opened 2018-01-01
Respect NONE connection between pipes.

### LINKED ISSUE #4781: [Suggestion] Separate wires without covers
Like with (almost) every other mod that brings a item transfer/logistic cable, make them disconnect from each other with a wrench.
You can already set the connections to do PULL/PUSH/NONE, but those don't work on wire-to-wire connections, so why not have them disconnect when either side is set to NONE? Maybe even have them respect PULL/PUSH as well when connected to other wires. Would give a lot of options when making the wires.

The reason I'm asking for this is due to a current bug with MCMP (or C&B?) causing crashes (https://github.com/AlgorithmX2/Chisels-and-Bits/issues/310). While it is going to get fixed eventually (hopefully), it could happen again, and always nice with options.

thommy101:
Fixes #4781 

Quick and easy solution to make pipes not connect when one of both sides is configured with NONE.

Only drawback to this solution (and the reason I opened it as a PR) is that once set to NONE, they do not connect, and there's no way to reconfigure it back to normal/push/pull other than breaking the pipe and placing it again.

Confirmed that switching one pipe to none, the networks get separated and do not connect visual and "technical".

### DIFF
--- a/src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
+++ b/src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
@@ -329,6 +329,11 @@ public abstract class TileEntitySidedPipe extends TileEntity implements ITileNet
 	@Override
 	public boolean canConnect(EnumFacing side)
 	{
+		if(connectionTypes[side.ordinal()] == ConnectionType.NONE)
+		{
+			return false;
+		}
+
 		if(!redstoneSet)
 		{
 			if(redstoneReactive)
@@ -611,6 +616,10 @@ public abstract class TileEntitySidedPipe extends TileEntity implements ITileNet
 					sendDesc = true;
 
 					onModeChange(EnumFacing.getFront(hitSide.ordinal()));
+
+					refreshConnections();
+					notifyTileChange();
+
 					player.sendMessage(new TextComponentGroup().translation("tooltip.configurator.modeChange").string(" ").translation(connectionTypes[hitSide.ordinal()].translationKey()));
 
 					return EnumActionResult.SUCCESS;
--- a/src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
+++ b/src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
@@ -482,6 +482,10 @@ public abstract class TileEntitySidedPipe extends TileEntity implements ITileNet
 	protected void onModeChange(EnumFacing side)
 	{
 		markDirtyAcceptor(side);
+		if (getPossibleTransmitterConnections() != currentTransmitterConnections)
+		{
+			markDirtyTransmitters();
+		}
 	}
 
 	protected void markDirtyTransmitters()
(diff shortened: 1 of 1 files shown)


thiakil:
> and "technical".

Are you sure? For me it doesn't according to the network tool unless I add this:
```diff
--- a/src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
+++ b/src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
@@ -482,6 +482,9 @@ public abstract class TileEntitySidedPipe extends TileEntity implements ITileNet
        protected void onModeChange(EnumFacing side)
        {
                markDirtyAcceptor(side);
+               if (getPossibleTransmitterConnections() != currentTransmitterConnections){
+                       markDirtyTransmitters();
+               }
        }
```

As for the 'cant change from NONE' issue (which seems to affect connections to other blocks too), this lets the side set to none be reconfigured (`onSneakRightClick` method):
```diff
@@ -609,6 +612,9 @@ public abstract class TileEntitySidedPipe extends TileEntity implements ITileNet
                        }
                        else {
                                EnumFacing hitSide = sideHit(hit.subHit + 1);
+
+                               if (hitSide == null && connectionTypes[side.ordinal()] == ConnectionType.NONE)
+                                       hitSide = side;

                                if(hitSide != null)
                                {
```

thommy101:
... Weird, in my tests yesterday they correctly split the networks and gas couldn't flow from one to the other. The network probe correctly showed the amount of pipes in the network.
Will test tonight.

The fix for the hitside is also nice, will look into that too :)

thiakil:
i was testing with a water filled pipe, in case it makes any difference. Universal cables, logisticals also showed the network probe stuff. Don't think I tested gas tubes

thommy101:
Thank you @thiakil, your changes where exactly what was needed.
Don't know how my first tests succeeded.. Tested again and it's exactly as you said. Applied your changes :)

thiakil:
Who knows, my dev env seemed to be doing some weird things with bounding boxes not always showing too so idk lol

thiakil:
Minor trouble, items in transit seem to get stuck, bouncing back and forth (not going back to source) when the destination is cut off mid transit

thiakil:
Also when the game was restarted and the world reloaded the none connections reconnected :(

thiakil:
Never mind that last comment, the connection modes weren't saving at all

pupnewfster:
I manually pulled these changes into a local branch to see how things worked and a couple comments:
1. The bug @thiakil described with the pipes getting stuck and not going back into their home spot, was fixed by #5335 
2. It seems decent to me, though one thing I think people may find confusing is that it *properly* removes the bounding boxes of the connection between the two things when it is set to "NONE" so the actual side has to be right clicked against, rather than the outline that previously remained when hovering the section the connection would be.

pupnewfster:
Actually upon more testing I found one sort of critical bug introduced by this PR. When for example connecting two pressurized tubes of different gases one overwrites the other. Though maybe this is a bug even without this PR which if so we should fix ASAP.

thiakil:
Closing in lieu of #5387

PR #4903, opened 2018-01-07
Updated Russian translation

raspopov:
Added all missed translations (nuggets, mekanism.gui.buffer*, mekanism.pipe.*, tooltip.freerunner.* etc)
Changed some long words to similar short ones to fit GUI (digital miner, item eject mode)
Changed incorrect Personal Chest name and description (long time ago it was correct).

### DIFF
--- a/src/main/resources/assets/mekanism/lang/ru_ru.lang
+++ b/src/main/resources/assets/mekanism/lang/ru_ru.lang
@@ -37,7 +37,7 @@ item.Salt.name=Соль
 item.BrineBucket.name=Ведро рассола
 item.LithiumBucket.name=Ведро лития
 item.HeavyWaterBucket.name=Ведро тяжелой воды
-item.FreeRunners.name=Свободные бегунки
+item.FreeRunners.name=Бегунки
 item.ArmoredJetpack.name=Бронированный реактивный ранец
 item.ConfigurationCard.name=Карта настройки
 item.SeismicReader.name=Сейсмический считыватель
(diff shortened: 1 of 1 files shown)


Foxgreat:
Good work!

src/main/resources/assets/mekanism/lang/ru_ru.lang
```diff
@@ -439,13 +449,13 @@ gui.start=Запуск
...
-gui.eject=Извлечение
+gui.eject=Выброс
 gui.input=Вход
 gui.strictInput=Строгий вход
 gui.slots=Слоты
 gui.state=Состояние
-gui.on=Вкл.
-gui.off=Выкл.
+gui.on=Да
+gui.off=Нет
```
thiakil: Why change these to yes/no?

raspopov: Same meanings and fits any tiny button on GUI (old ones does not!).

raspopov: (for example "filter: on/off" can be clearly understand in form "filter [active]: yes/no")

thiakil: Hmmm I see. My issue with that is we can't guarantee it will always make sense as the english does not mean the same

raspopov: Make a buttons with dynamic fit to text size.

Issue #4901, opened 2018-01-05
[Suggestion] Oredict sawdust recipes


SHIuSH:
mekanism sawdust is incompatible with thermalfoundation's sawdust in some recipes like making paper, cardboard

SHIuSH:
Yeah apparently mekanism's recipes don't use oredict sawdust yet mekanism's sawdust is present in oredict

ProsperCraft:
Yes please, I ran into the same issue.

thiakil:
fixed in d8a4972

thommy101:
Released in in version 9.4.4

PR #4918, opened 2018-01-23
Fix for Paper-Recipecompability

GityUpNow:
I fixed one small Issues:

#4901
I replaced Mekanism-Sawdust in the Paper-Recipe with generic Sawdust from the OreDictionary, which should fix the problem


GityUpNow:
Sorry, forgot to regenerate ^^'
Yeah, I tested it, everything seems to work to as planned. The Paper-Recipe is now also accepting Thermalexpansion-Sawdust.

Concerning the second part:
It seems so, there are no typos, both used names are the most commonly used for Sawdust and the call to the OreDictionary seems to be correct as well.

thommy101:
Thanks! Will review soon... I think there might be a slight issue with my recipe generation code. The recipe type is not defined correctly.
Don't worry, it's something I will look at ;)

GityUpNow:
Glad to help ^^

thommy101:
Aah.. Already see what's wrong.. Could you please apply these changes on the 1.12 branch and open a PR to that branch? That contains a fix from me (since last week) to properly define recipe types :)

GityUpNow:
Okay, learned a ton of stuff about git.

Closing this PR now.

PR #4919, opened 2018-01-23
Fix for Sawdust-Recipecompability

GityUpNow:
(Continuation of https://github.com/aidancbrady/Mekanism/pull/4918#pullrequestreview-90953620)


thiakil:
IMO this should be done with a custom ingredient that matches both oredicts, not duplicating recipes

thiakil:
solved this another way

PR #4952, opened 2018-02-18
Item drop compatibility

danielzgtg:
Fix machines disappearing after mining/wrenching.
Fix RFTools builder losing machines.
Simplify and use methods in Block instead of calculating drops.
Make drops behave the same as mined blocks.

### DIFF
--- a/src/main/java/mekanism/common/block/BlockBasic.java
+++ b/src/main/java/mekanism/common/block/BlockBasic.java
@@ -38,6 +38,7 @@ import mekanism.common.util.MekanismUtils;
 import mekanism.common.util.SecurityUtils;
 import mekanism.common.util.StackUtils;
 import net.minecraft.block.Block;
+import net.minecraft.block.BlockFlowerPot;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.properties.PropertyEnum;
 import net.minecraft.block.state.BlockStateContainer;
@@ -449,7 +450,7 @@ public abstract class BlockBasic extends Block
 
 						if(entityplayer.isSneaking())
 						{
-							dismantleBlock(state, world, pos, false);
+							MekanismUtils.dismantleBlock(this, state, world, pos);
 							return true;
 						}
 
@@ -817,8 +818,7 @@ public abstract class BlockBasic extends Block
 		super.breakBlock(world, pos, state);
 	}
 
-	@Override
-	public ItemStack getPickBlock(IBlockState state, RayTraceResult target, World world, BlockPos pos, EntityPlayer player)
+	private ItemStack getDropItem(IBlockState state, IBlockAccess world, BlockPos pos)
 	{
 		BasicBlockType type = BasicBlockType.get(state);
 		ItemStack ret = new ItemStack(this, 1, state.getBlock().getMetaFromState(state));
@@ -859,48 +859,60 @@ public abstract class BlockBasic extends Block
 	}
 
 	@Override
-	public Item getItemDropped(IBlockState state, Random random, int fortune)
+	public ItemStack getPickBlock(IBlockState state, RayTraceResult target, World world, BlockPos pos, EntityPlayer player)
 	{
-		return null;
+		return getDropItem(state, world, pos);
 	}
 
 	@Override
-	public boolean removedByPlayer(IBlockState state, World world, BlockPos pos, EntityPlayer player, boolean willHarvest)
+	public void getDrops(NonNullList<ItemStack> drops, IBlockAccess world, BlockPos pos, IBlockState state, int fortune)
 	{
-		if(!player.capabilities.isCreativeMode && !world.isRemote && willHarvest)
-		{
-			float motion = 0.7F;
-			double motionX = (world.rand.nextFloat() * motion) + (1.0F - motion) * 0.5D;
-			double motionY = (world.rand.nextFloat() * motion) + (1.0F - motion) * 0.5D;
-			double motionZ = (world.rand.nextFloat() * motion) + (1.0F - motion) * 0.5D;
-
-			EntityItem entityItem = new EntityItem(world, pos.getX() + motionX, pos.getY() + motionY, pos.getZ() + motionZ, getPickBlock(state, null, world, pos, player));
-
-			world.spawnEntity(entityItem);
-		}
-
-		return world.setBlockToAir(pos);
+		drops.add(getDropItem(state, world, pos));
 	}
 
-	public ItemStack dismantleBlock(IBlockState state, World world, BlockPos pos, boolean returnBlock)
+	/**
+	 * {@inheritDoc}
+	 * Keep tile entity in world until after
+	 * {@link Block#getDrops(NonNullList, IBlockAccess, BlockPos, IBlockState, int)}.
+	 * Used together with {@link Block#harvestBlock(World, EntityPlayer, BlockPos, IBlockState, TileEntity, ItemStack)}.
+	 *
+	 * @author Forge
+	 * @see BlockFlowerPot#removedByPlayer(IBlockState, World, BlockPos, EntityPlayer, boolean)
+	 */
+	@Override
+	public boolean removedByPlayer(IBlockState state, World world, BlockPos pos, EntityPlayer player,
+	                               boolean willHarvest)
 	{
-		ItemStack itemStack = getPickBlock(state, null, world, pos, null);
+		return willHarvest || super.removedByPlayer(state, world, pos, player, willHarvest);
+	}
 
+	/**
+	 * {@inheritDoc}
+	 * Used together with {@link Block#removedByPlayer(IBlockState, World, BlockPos, EntityPlayer, boolean)}.
+	 *
+	 * @author Forge
+	 * @see BlockFlowerPot#harvestBlock(World, EntityPlayer, BlockPos, IBlockState, TileEntity, ItemStack)
+	 */
+	@Override
+	public void harvestBlock(World world, EntityPlayer player, BlockPos pos,
+	                         IBlockState state, TileEntity te, ItemStack stack)
+	{
+		super.harvestBlock(world, player, pos, state, te, stack);
 		world.setBlockToAir(pos);
+	}
 
-		if(!returnBlock)
-		{
-			float motion = 0.7F;
-			double motionX = (world.rand.nextFloat() * motion) + (1.0F - motion) * 0.5D;
-			double motionY = (world.rand.nextFloat() * motion) + (1.0F - motion) * 0.5D;
-			double motionZ = (world.rand.nextFloat() * motion) + (1.0F - motion) * 0.5D;
-
-			EntityItem entityItem = new EntityItem(world, pos.getX() + motionX, pos.getY() + motionY, pos.getZ() + motionZ, itemStack);
-
-			world.spawnEntity(entityItem);
-		}
-
-		return itemStack;
+	/**
+	 * Returns that this "cannot" be silk touched.
+	 * This is so that {@link Block#getSilkTouchDrop(IBlockState)} is not called, because only
+	 * {@link Block#getDrops(NonNullList, IBlockAccess, BlockPos, IBlockState, int)} supports tile entities.
+	 * Our blocks keep their inventory and other behave like they are being silk touched by default anyway.
+	 *
+	 * @return false
+	 */
+	@Override
+	protected boolean canSilkHarvest()
+	{
+		return false;
 	}
 
 	@Override
--- a/src/main/java/mekanism/common/block/BlockBounding.java
+++ b/src/main/java/mekanism/common/block/BlockBounding.java
@@ -6,6 +6,7 @@ import mekanism.common.block.states.BlockStateBounding;
 import mekanism.common.tile.TileEntityAdvancedBoundingBlock;
 import mekanism.common.tile.TileEntityBoundingBlock;
 import net.minecraft.block.Block;
+import net.minecraft.block.BlockFlowerPot;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.state.BlockStateContainer;
 import net.minecraft.block.state.IBlockState;
@@ -16,9 +17,13 @@ import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.EnumBlockRenderType;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.EnumHand;
+import net.minecraft.util.NonNullList;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.RayTraceResult;
+import net.minecraft.world.IBlockAccess;
 import net.minecraft.world.World;
+import net.minecraftforge.event.world.BlockEvent;
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent;
 
 public class BlockBounding extends Block
 {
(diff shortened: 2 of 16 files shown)


src/main/java/mekanism/common/block/BlockBasic.java
```diff
@@ -449,7 +447,7 @@ else if(tile instanceof TileEntityBin)
 
 						if(entityplayer.isSneaking())
 						{
-							dismantleBlock(state, world, pos, false);
+							dismantleBlock(state, world, pos);
```
thiakil: why is there a new `dismantleBlock` method?

src/main/java/mekanism/common/block/BlockBasic.java
```diff
@@ -859,48 +862,28 @@ else if(type == BasicBlockType.INDUCTION_PROVIDER)
...
-
-		if(!returnBlock)
-		{
-			float motion = 0.7F;
-			double motionX = (world.rand.nextFloat() * motion) + (1.0F - motion) * 0.5D;
-			double motionY = (world.rand.nextFloat() * motion) + (1.0F - motion) * 0.5D;
-			double motionZ = (world.rand.nextFloat() * motion) + (1.0F - motion) * 0.5D;
-
-			EntityItem entityItem = new EntityItem(world, pos.getX() + motionX, pos.getY() + motionY, pos.getZ() + motionZ, itemStack);
+	@Override
```
thiakil: this is standard behaviour, why is it here?

src/main/java/mekanism/common/block/BlockBasic.java
```diff
@@ -859,48 +862,28 @@ else if(type == BasicBlockType.INDUCTION_PROVIDER)
...
+	public void harvestBlock(World world, EntityPlayer player, BlockPos pos, IBlockState state, TileEntity te, ItemStack stack) {
+		player.addStat(StatList.getBlockStats(this));
+		player.addExhaustion(0.005F);
+	}
 
-			world.spawnEntity(entityItem);
+	@Override
+	public void onBlockHarvested(World world, BlockPos pos, IBlockState state, EntityPlayer player) {
```
thiakil: again, standard behaviour

src/main/java/mekanism/common/block/BlockBounding.java
```diff
@@ -81,14 +81,27 @@ public ItemStack getPickBlock(IBlockState state, RayTraceResult target, World wo
 	}
 
 	@Override
-	public boolean removedByPlayer(IBlockState state, World world, BlockPos pos, EntityPlayer player, boolean willHarvest)
```
thiakil: This will cause bounding boxes being broken to not break their parent.

src/main/java/mekanism/common/block/BlockBounding.java
```diff
@@ -81,14 +81,27 @@ public ItemStack getPickBlock(IBlockState state, RayTraceResult target, World wo
...
-			return state1.getBlock().removedByPlayer(state1, world, tileEntity.mainPos, player, willHarvest);
+			state1.getBlock().getDrops(drops, world, pos, state1, fortune);
 		} catch(Exception e) {
-			return false;
+			e.printStackTrace();
+		}
+	}
+
+	@Override
+	public void harvestBlock(World world, EntityPlayer player, BlockPos pos, IBlockState state, TileEntity te, ItemStack stack) {
```
thiakil: Standard behaviour copy paste

src/main/java/mekanism/common/block/BlockBounding.java
```diff
@@ -115,18 +128,6 @@ public float getPlayerRelativeBlockHardness(IBlockState state, EntityPlayer play
 		}
 	}
 
-	@Override
```
thiakil: Why

src/main/java/mekanism/common/block/BlockEnergyCube.java
```diff
@@ -284,35 +255,25 @@ public EnumBlockRenderType getRenderType(IBlockState state)
...
+	private ItemStack getDropItem(IBlockAccess world, BlockPos pos) {
 		TileEntityEnergyCube tileEntity = (TileEntityEnergyCube)world.getTileEntity(pos);
+
 		ItemStack itemStack = new ItemStack(MekanismBlocks.EnergyCube);
-		
-		if(!itemStack.hasTagCompound())
+		if (tileEntity == null)
 		{
-			itemStack.setTagCompound(new NBTTagCompound());
+			throw new NullPointerException();
```
thiakil: You must give it a message if you insist on throwing. Though I don't see much point.

src/main/java/mekanism/common/block/BlockEnergyCube.java
```diff
@@ -284,35 +255,25 @@ public EnumBlockRenderType getRenderType(IBlockState state)
...
 		}
-		
-		if(tileEntity instanceof ISecurityTile)
```
thiakil: why are you removing instance checks?

src/main/java/mekanism/common/block/BlockTransmitter.java
```diff
@@ -231,49 +230,54 @@ public boolean onBlockActivated(World world, BlockPos pos, IBlockState state, En
...
+		world.setBlockToAir(pos);
+	}
+
+	private ItemStack getDropItem(IBlockAccess world, BlockPos pos) {
 		TileEntitySidedPipe tileEntity = getTileEntitySidedPipe(world, pos);
 		if(tileEntity != null)
 		{
-			itemStack = new ItemStack(MekanismBlocks.Transmitter, 1, tileEntity.getTransmitterType().ordinal());
-			
-			if(!itemStack.hasTagCompound())
```
thiakil: This will wipe any data a capability may have added.

src/main/java/mekanism/common/block/BlockBasic.java
danielzgtg: We don't need neither the ```returnBlock``` boolean anymore nor reinventing the wheel while dropping blocks with the other changes.

src/main/java/mekanism/common/block/BlockBasic.java
danielzgtg: This was used to still add stats, but not drop the block because the tile entity is missing by now.
However, I found a better solution...

src/main/java/mekanism/common/block/BlockBasic.java
danielzgtg: The block is dropped here instead of ```harvestBlock```.

src/main/java/mekanism/common/block/BlockBounding.java
danielzgtg: This is not needed because ```getDrops``` is overriden.

src/main/java/mekanism/common/block/BlockEnergyCube.java
danielzgtg: Removed as you asked.

src/main/java/mekanism/common/block/BlockEnergyCube.java
danielzgtg: There was a warning that ```TileEntityEnergyCube``` is always ```instanceof ISecurityTile```.

src/main/java/mekanism/common/block/BlockBasic.java
thiakil: javadoc states `Called regardless of if the player's tool can actually collect this block`

src/main/java/mekanism/common/block/BlockTransmitter.java
danielzgtg: The constructor ```ItemStack(Block, int, int)``` (calls ```ItemStack(Item, int, int)```) is being used, so the ```NBTTagCompound``` should always be null.

danielzgtg:
Summary of what is being done:
Previously, Mekanism used ```removedByPlayer``` to drop the block.
This was a problem because RFTools' builder uses ```getDrops```, which was not returning any items because ```quantityDropped``` was returning 0.

The original pull request made ```getDrops``` return the proper amount of items. It did not work until drops were moved ```onBlockHarvested``` since the tile entity was removed before ``harvestDrops```.

The second set of changes did not copy and paste vanilla code and used ```removedByPlayer``` to delay removal of the tile entity. Unfortunately, I just found out the new solution caused inventory loss, so I reverted it, and I am still working on it.

src/main/java/mekanism/common/block/BlockTransmitter.java
thiakil: It does not matter what constructor you use.

thiakil:
`onBlockHarvested` is called by `removedByPlayer` (and only by it) regardless of whether the player has the correct tool or not. The `willHarvest` flag needs to be checked.

src/main/java/mekanism/common/block/BlockBounding.java
danielzgtg: Oops. Fixed with new ```removedMainBlock``` method.

src/main/java/mekanism/common/block/BlockBasic.java
danielzgtg: Removed

src/main/java/mekanism/common/block/BlockTransmitter.java
danielzgtg: I must have overlooked ```forgeInit```

danielzgtg:
# New changes #
1. Removed duplicated vanilla code in normal ```harvestBlock```
2. Added new edited vanilla code to ```harvestBlock``` for ```BlockContainer``` to fix inventory loss caused by change 1.
3. Fixed boundary blocks broken by original pull request.
4. Extracted ```dismantleBlock``` into utils.
5. Restored ```hasTagCompound``` code.

Now waiting on McJty/RFTools#1688. This pull request will cause a DigiMiner duplication glitch with RFTools builder otherwise.

# Some testing #
Test environment: Singleplayer, survival & creative, normal diamond pickaxe, silk diamond pickaxe, crescent hammer (thermal expansion), RFTools builder.

Harvesting these seems to work:
1. Digiminer (empty)
2. Digiminer with 1 grass
2. Osmium & obsidian storage blocks
3. Harvesting enrichment chamber with obsidian storage block
4. Harvesting enrichment chamber with the item from test 4

src/main/java/mekanism/common/block/BlockBounding.java
```diff
@@ -76,20 +86,82 @@ public ItemStack getPickBlock(IBlockState state, RayTraceResult target, World wo
 			IBlockState state1 = world.getBlockState(tileEntity.mainPos);
 			return state1.getBlock().getPickBlock(state1, target, world, tileEntity.mainPos, player);
 		} catch(Exception e) {
+			e.printStackTrace();
```
thiakil: no

src/main/java/mekanism/common/block/BlockBounding.java
```diff
@@ -76,20 +86,82 @@ public ItemStack getPickBlock(IBlockState state, RayTraceResult target, World wo
...
 		try {
 			TileEntityBoundingBlock tileEntity = (TileEntityBoundingBlock)world.getTileEntity(pos);
-			IBlockState state1 = world.getBlockState(tileEntity.mainPos);
-			return state1.getBlock().removedByPlayer(state1, world, tileEntity.mainPos, player, willHarvest);
+			BlockPos mainPos = tileEntity.mainPos;
+			IBlockState state1 = world.getBlockState(mainPos);
+			state1.getBlock().getDrops(drops, world, mainPos, state1, fortune);
 		} catch(Exception e) {
-			return false;
+			e.printStackTrace();
```
thiakil: log it properly or rewrite it so it doesn't need the try catch (the better option)

src/main/java/mekanism/common/block/BlockBounding.java
```diff
@@ -76,20 +86,82 @@ public ItemStack getPickBlock(IBlockState state, RayTraceResult target, World wo
...
+			return true;
 		}
+
+		removeMainBlock(world, pos);
+		return super.removedByPlayer(state, world, pos, player, willHarvest);
+	}
+
+	private void removeMainBlock(World world, BlockPos pos) {
+		TileEntityBoundingBlock tileEntity = (TileEntityBoundingBlock)world.getTileEntity(pos);
+		world.setBlockToAir(tileEntity.mainPos);
```
thiakil: Setting the main block to air doesnt seem right to me...

src/main/java/mekanism/common/block/BlockTransmitter.java
```diff
@@ -231,49 +230,84 @@ public boolean onBlockActivated(World world, BlockPos pos, IBlockState state, En
...
+			ItemStack itemStack = new ItemStack(MekanismBlocks.Transmitter, 1,
+					tileEntity.getTransmitterType().ordinal());
```
thiakil: this line does not need to be split

src/main/java/mekanism/common/block/BlockTransmitter.java
```diff
@@ -231,49 +230,84 @@ public boolean onBlockActivated(World world, BlockPos pos, IBlockState state, En
...
+			if (!itemStack.hasTagCompound())
```
thiakil: needless format change

src/main/java/mekanism/common/block/BlockBounding.java
danielzgtg: What are we trying to catch anyway? If it's just a ```NullPointerException```, this can be simply replaced with a null check.

src/main/java/mekanism/common/block/BlockBounding.java
danielzgtg: It works fine, because we drop the items in ```getDrops``` on behalf of the main block.

The main block shouldn't mind not being told about being removed, because all the ```removedByPlayer``` and ```harvestBlock``` hooks were just added.
All of the other bounding blocks are removed with block updates.

src/main/java/mekanism/common/block/BlockBounding.java
thiakil: That's my best guess as to what it was for, so yeah, just null check it

src/main/java/mekanism/common/block/BlockBounding.java
thiakil: > It works fine, because we drop the items in getDrops on behalf of the main block.

That is bad practice. The main block should be handling its own breakage.

danielzgtg: It's not the best but if the main block were to handle its own harvesting, it would drop itself as an itemstack into the world.
That means I can't pack up my base with RFTools builder anymore (or probably even Buildcraft quarry), which is the purpose of this pull request.
Another way is to add a method to the main block, to ask it to give us the items, but isn't that just ```getDrops```?

src/main/java/mekanism/common/block/BlockTransmitter.java
danielzgtg: Removed most IDE and non-essential changes.

src/main/java/mekanism/common/block/BlockBounding.java
danielzgtg: Doing a ```getDrops``` and ```setBlockToAir``` seems to be standard, and will be done to us by other mods:
1. https://github.com/McJty/RFTools/blob/496527a1faa37ad16b60d2ea10751074c3d7d2e8/src/main/java/mcjty/rftools/blocks/builder/BuilderTileEntity.java#L1136-L1144
2. https://findusages.com/search/net.minecraft.block.Block/getDrops$4
3. https://findusages.com/search/net.minecraft.block.Block/getDrops$5

josephcsible:
> Now waiting on McJty/RFTools#1688. This pull request will cause a DigiMiner duplication glitch with RFTools builder otherwise.

Done in RFTools 7.32.

josephcsible:
Why was this closed?

dizzyd:
Hi @josephcsible - this PR got closed as we are in the process of merging Mekanica (a short-lived fork) back into Mekanism. As you can see, there are now some conflicts, so if you're still interested in having this merged, we'll need to get those resolved. Thanks! :)

pupnewfster:
Closing this as I ported the changes to be able to merge cleanly in #5384 once we have a chance to further review and test this.

PR #5066, opened 2018-05-29
work on buildcraft

fandig:
want the Universal Cable to work whit buildcraft when they are moving back to MJ and don't get a simple way to keep all MJ you get over without a convertor and it's not simple to get a cnvertor mod to work whit 1.12.2


thommy101:
This looks more like a feature request. Please open an ticket, not a pull request with a totally unrelated commit from someone else.

PR #5137, opened 2018-08-09
Separated mob loading from TE loading and moved to PreInit

Saereth:
resolves https://github.com/aidancbrady/Mekanism/issues/5136


Saereth:
not sure I follow, it's called in the preInit evenhandler similarly to how it was previously called from init, I just separated it from addEntities, made a nearly identical function called addMobEntites that only does the mob registering and called that in pre-init instead of init to match the coding style already in place.

thiakil:
there is a specific registry event for mobs

Saereth:
yeah I'm not sure when I'll have time to work on this again I'll just close for now, maybe someone else will have time to fix it for Vyraal's request, if not I'll revisit once I have some time from work

PR #5172, opened 2018-10-01
Fixed armor spawning checks firing when they shouldn't.

darkhax:
The event listener used for spawning zombies and skeletons with swords and armor from this mod is using the non-specific parent event. This means that the logic for spawning mobs with these items is fired for all of the events that inherit from LivingSpawnEvent. This causes the logic to run in many situations where it shouldn't. Under dev environment conditions, there are two situations where this event is firing where it probably shouldn't be. 

1. CheckSpawn - This is fired when the game tries to see if a location is valid for a mob to spawn. This method is fired a lot, and doesn't always, or even usually result in a mob spawning. You don't need to fire this early to achieve the same effect. This also results in a lot of wasted operations since the mob doesn't always spawn. This would also mean that every mob that does spawn has two attempts at receiving the items, since the actual spawn event also fires.
2. AllowDespawn - This is fired when the game tries to despawn an entity This has been fired after the entity has been spawned, and can be fired a virtually unlimited amount of times for the same entity. This is also a waste of logic, and in theory would allow players to create systems which try to despawn mobs that can't despawn, and forcing this code to run a lot, and give that entity lots of items. Combining this with some other mods which allow you to take items off of mobs would result in an infinite source of these items. 

You can confirm this issue by doing the following steps. 

1. Add some debug code to [this](https://github.com/aidancbrady/Mekanism/blob/3e469fb40a2a086527cf14dbe9ba709f708a9bca/src/main/java/mekanism/tools/common/MekanismTools.java#L371) method. Something like `System.out.println(event.getEntityLiving().getPersistentID().toString() + " - " + event.getClass().getName());` should be fine.
2. Run the game.
3. Create a new flat world with creative mode game type.
4. Set difficulty to not be peaceful.
5. Set time to night.
6. Fly around in the world and let some mobs spawn. Do this for about 30 seconds.
7. Look in the console and see the various debug messages.

### DIFF
--- a/src/main/java/mekanism/tools/common/MekanismTools.java
+++ b/src/main/java/mekanism/tools/common/MekanismTools.java
@@ -368,7 +368,7 @@ public class MekanismTools implements IModule
 	}
 
 	@SubscribeEvent
-	public void onLivingSpecialSpawn(LivingSpawnEvent event)
+	public void onLivingSpecialSpawn(LivingSpawnEvent.SpecialSpawn event)
 	{
 		double chance = event.getWorld().rand.nextDouble();
 		int armorType = event.getWorld().rand.nextInt(4);


thiakil:
thanks!

Issue #5146, opened 2018-08-26
Mining a Wizardry Wisdom Tree with an Atomic Disassembler in vein mode causes crash


TigerWalts:
#### Issue description:
When breaking **Wisdom Trees** with the **Atomic Disassembler** in Vein mode, an Exception is thrown when the loop finding all the blocks reaches a Wisdom Leaf Block.

The **LibrarianLib** `getPickBlock()` method requires the `player` argument to be non-null. When breaking multiple blocks, **Mekanism** calls this method with the argument set to `null` - [ItemAtomicDisassembler.java:414](https://github.com/aidancbrady/Mekanism/blob/4a877de4b2dbd74c17a89155a2700fe6e024dc81/src/main/java/mekanism/common/item/ItemAtomicDisassembler.java#L414).

This crash is similar to issue #5034 only this time the `player` argument is the cause and the call is being made from the `Finder` class' `loop` method.

#### Steps to reproduce:
1. Place multiple **Wisdom Wool Log**s next to each other
2. Place some **Wisdom Leaves** touching the **Wisdom Wool Log**s
3. Set your **Atomic Disassembler** to **Vein** mode
4. Mine the **Wisdom Wood Log**s

On SSP, the error will appear in the log, but the game will continue to run. On a Dedicated server, the Exception will cause the server to restart.

#### Version (make sure you are on the latest version before reporting):
**Forge:** 14.23.4.2715
**Mekanism:** 1.12.2-9.4.13.349
**Wizardry:** 0.8.5
**LibrarianLib:** 1.12.2-4.9
**Modpack - All the Mods 3:** 5.9.8d

#### If a (crash)log is relevant for this issue, link it here: (_It's almost always relevant_)

```
java.lang.IllegalArgumentException: Parameter specified as non-null is null: method com.teamwizardry.librarianlib.features.base.block.BlockModLeaves.getPickBlock, parameter player
	at com.teamwizardry.librarianlib.features.base.block.BlockModLeaves.getPickBlock(BlockModLeaves.kt)
	at mekanism.common.item.ItemAtomicDisassembler$Finder.loop(ItemAtomicDisassembler.java:414)
	at mekanism.common.item.ItemAtomicDisassembler$Finder.loop(ItemAtomicDisassembler.java:418)
	at mekanism.common.item.ItemAtomicDisassembler$Finder.calc(ItemAtomicDisassembler.java:425)
[...]

PR #5178, opened 2018-10-12
fix root cause of #5146 (mining wizardry wisdom wood logs with atomic disassembler)

tcheinen:
As mentioned in #5146, the getPickBlock method used in the finder class for the atomic disassembler passes a null argument for the player.  Although this normally works, it is possible for the block to override the method in a way that breaks it.  This was the case in Wizardry and may still be a problem with other mods.  I've modified the code so that it passes the player object of the player who is using the disassembler.


tcheinen:
im going to go ahead and close this because it got fixed later

PR #5201, opened 2018-11-22
Add combiner recipes for Cyclic's nether/end ores

josephcsible:
The cost for redstone and lapis is lower than for the normal ores because Cyclic's versions of these have much lower yields, averaging only 2.2 redstone and 4.4 lapis per ore, even with fortune 3.

### DIFF
--- a/src/main/java/mekanism/common/integration/MekanismHooks.java
+++ b/src/main/java/mekanism/common/integration/MekanismHooks.java
@@ -186,7 +186,6 @@ public final class MekanismHooks
 		} catch(Exception e) {}
 	}
 
-	@Method(modid = CYCLIC_MOD_ID)
 	public void registerCyclicCombinerOreRecipe(String ore, int quantity, Item input2, String outputName)
 	{
 		Item outputItem = ForgeRegistries.ITEMS.getValue(new ResourceLocation(CYCLIC_MOD_ID, outputName));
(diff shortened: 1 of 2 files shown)


src/main/java/mekanism/common/integration/MekanismHooks.java
```diff
@@ -173,6 +186,49 @@ public void loadOCDrivers()
 		} catch(Exception e) {}
 	}
 
+	@Method(modid = CYCLIC_MOD_ID)
+	public void registerCyclicCombinerOreRecipe(String ore, int quantity, Item input2, String outputName)
+	{
+		Item outputItem = ForgeRegistries.ITEMS.getValue(new ResourceLocation(CYCLIC_MOD_ID, outputName));
+		if(outputItem == Items.AIR) {
```
thiakil: that's not the default key as far as I can find (unless it has changed recently)

thiakil:
> averaging only 2.2 redstone and 4.4 lapis per ore

Please make it the maximum possible yield or max +1

josephcsible:
> Please make it the maximum possible yield or max +1

Why? As long as it's more expensive than the average yield, it can't be an infinite loop. Also, the costs of the vanilla combiner recipes aren't like that. For example, for vanilla lapis lazuli ore and a fortune 3 pickaxe, the average yield is 13.2, the maximum yield is 32, and the combiner cost is 16.

(I'll fix the two inline comments tonight.)

src/main/java/mekanism/common/integration/MekanismHooks.java
josephcsible: Oops, you're right.

thiakil:
> As long as it's more expensive than the average yield
needs to be what machine processing of the ore outputs

josephcsible:
To my knowledge, no machines in either Mekanism or Cyclic let you do better processing of those redstone and lapis ores.

PR #5203, opened 2018-11-23
Clamp power to max values rather than letting it wrap

josephcsible:
One example of what this fixes is the Creative Energy Cube showing "-1 RF"
and confusing a lot of other mods.

### DIFF
--- a/src/main/java/mekanism/common/integration/forgeenergy/ForgeEnergyItemWrapper.java
+++ b/src/main/java/mekanism/common/integration/forgeenergy/ForgeEnergyItemWrapper.java
@@ -58,16 +58,25 @@ public class ForgeEnergyItemWrapper extends ItemCapability implements IEnergySto
 		return 0;
 	}
 
+	private static int clampToInt(double d) {
+		// Not using Math.min because it makes you cast things wrong
+		if(d < Integer.MAX_VALUE) {
+			return (int)d;
+		} else {
+			return Integer.MAX_VALUE;
+		}
+	}
+
 	@Override
 	public int getEnergyStored() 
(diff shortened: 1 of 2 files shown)


src/main/java/mekanism/common/integration/forgeenergy/ForgeEnergyItemWrapper.java
```diff
@@ -58,16 +58,25 @@ public int extractEnergy(int maxExtract, boolean simulate)
 		return 0;
 	}
 
+	private static int clampToInt(double d) {
+		// Not using Math.min because it makes you cast things wrong
```
thiakil: please explain

josephcsible:
Mixing integral and floating-point types with Math.min can produce results that are slightly less or greater than what you'd consider the "right" answer to be. Example:
```java
public class MinCastTest {
  public static void main(String[] args) {
    double d = 1e300; // way bigger than longs, so the long should always be what's returned by Math.min
    System.out.println((long)Math.min(123456781234567812L, d)); // result is 123456781234567808 - 4 less than what you'd expect
    System.out.println((long)Math.min(123456789012345678L, d)); // result is 123456789012345680 - 2 more than what you'd expect
  }
}
```

thiakil:
right, of course (I don't even know why Mek uses doubles in the first place).

Not really sure why these need to be two methods though, or just a temp variable and a ternary

josephcsible:
I'd need to repeat the temporary variable and the ternary in both methods in each file then, and this seems cleaner than that.

thiakil:
Merged manually

PR #5204, opened 2018-11-25
move log on cardboard box blacklist outside loop

temp1011:
Prevents the log from being spammed with messages about this.


thiakil:
How strange, nice catch.

PR #5214, opened 2018-12-06
[1.7.10] Fix dupe bug with chemical washer

clienthax:
https://www.youtube.com/watch?v=i86989eh4Kc&feature=youtu.be


thiakil:
1.7 is dead

clienthax:
So you are using your personal opinion to refuse a bug fix for a older edition?

Seems non logical.

DarkGuardsman:
@thiakil Might be best to just pull it in then, if its dead then what is the harm?

PR #5222, opened 2018-12-18
cleanup a bit

temp1011:
This is just automated code inspection stuff for now while I try and understand the structure of the mod a bit more. Hopefully nothing too controversial.


temp1011:
I said automated more to point out that this is stuff that just is true by the java spec but fair enough, I'll try and clean up if I see other stuff I can sort out.

PR #5227, opened 2018-12-25
Fix dupe bug with factories

kmecpp:
This is a simple fix for a dupe bug which involves sorting items which contain inventories where the inventories would copy to the other items when sorting is being done. This fixes that by simply disabling sorting for items which have NBT data.

I only tested this on 1.7.10 but it may be an issue on other versions as well


thiakil:
a) thats not how that should be fixed
b) we don't update 1.7

PR #5282, opened 2019-02-24
Update to Forge 14.23.5.2808

dizzyd:
Updating to a more recent forge. There were a large number of broken functions where the mapping names had changed. 

In addition, there was a fix (e9a0e660077cc8f374aabfca029f30a1582495e) which broke all custom item rendering.


thiakil:
There is no actual need to update mappings versions. Nor does anything actually change in the final build with updating forge.

Also you're not quite right on the item rendering issue. That's not what broke it; the whole BakedCustomItemModel thing needs to go

PR #5335, opened 2019-04-03
Fix TransporterStack home locations & DM inventory interaction

aidancbrady:
Fixed Logistical Transporters not sending stuff back to home locations. It looks like this was a bug introduced from a PR in the main branch from a while ago.
Also, fixed Digital Miner's charge slot inventory functionality by trashing ISidedInventory stuff in IAdvancedBoundingBlock

### DIFF
--- a/src/main/java/mekanism/common/base/IAdvancedBoundingBlock.java
+++ b/src/main/java/mekanism/common/base/IAdvancedBoundingBlock.java
@@ -11,7 +11,6 @@ import mekanism.common.integration.MekanismHooks;
 import mekanism.common.integration.computer.IComputerIntegration;
 import mekanism.common.security.ISecurityTile;
 import net.minecraft.inventory.ISidedInventory;
-import net.minecraft.item.ItemStack;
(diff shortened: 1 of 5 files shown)


pupnewfster:
The one thing I am a bit unsure about is what is the point behind the changes to ChargeUtils. Other than that the changes look good.

aidancbrady:
canBeCharged() and canBeDischarged() are now a little more strict which fixes some transporter functionality. Basically, canBeCharged() will only return true if energy can actually be put inside an item (it's not full), and canBeDischarged() only returns true if energy can be pulled out (it's not empty).

pupnewfster:
> canBeCharged() and canBeDischarged() are now a little more strict which fixes some transporter functionality. Basically, canBeCharged() will only return true if energy can actually be put inside an item (it's not full), and canBeDischarged() only returns true if energy can be pulled out (it's not empty).

Sounds good you may want to modify the JavaDocs slightly to make that fact a bit more clear.

PR #5343, opened 2019-04-04
Introduce bounds to search for sparkles!

dizzyd:
In c368692a3, the recursive search was changed to iterative. The big problem with both approaches, however is that as more and more positions are scanned, it takes longer to actually determine if the position was scanned. This PR introduces a radius which is used to get a fixed set of block positions to scan and as each position is scanned, particles are generated. This reduces memory footprint and avoids a O(n^2) search.

Not quite working yet; pending me not being on a plane. :)

### DIFF
--- a/src/main/java/mekanism/client/SparkleAnimation.java
+++ b/src/main/java/mekanism/client/SparkleAnimation.java
@@ -1,15 +1,13 @@
 package mekanism.client;
 
-import java.util.HashSet;
-import java.util.Random;
-import java.util.Set;
-import mekanism.api.Coord4D;
+import java.util.concurrent.ThreadLocalRandom;
 import mekanism.common.MekanismSounds;
 import mekanism.common.config.MekanismConfig.general;
 import net.minecraft.tileentity.TileEntity;
-import net.minecraft.util.EnumFacing;
 import net.minecraft.util.EnumParticleTypes;
 import net.minecraft.util.SoundCategory;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.BlockPos.MutableBlockPos;
 import net.minecraft.world.World;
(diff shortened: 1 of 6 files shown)


thiakil:
Have you measured performance from checking the already iterated set before adding to the queue? (would help reduce at least some of the memory impact in this case)

src/main/java/mekanism/client/SparkleAnimation.java
```diff
@@ -16,59 +18,69 @@
...
+                      pos.getY() + random.nextDouble(), pos.getZ() + 1.01, 0, 0, 0);
+                world.spawnParticle(EnumParticleTypes.REDSTONE, pos.getX() + -.01,
+                      pos.getY() + random.nextDouble(), pos.getZ() + random.nextDouble(), 0, 0, 0);
+                world.spawnParticle(EnumParticleTypes.REDSTONE, pos.getX() + 1.01,
+                      pos.getY() + random.nextDouble(), pos.getZ() + random.nextDouble(), 0, 0, 0);
+            }
+        }
+
+        long elapsed = System.nanoTime() - start;
+        Mekanism.logger.info("Generated {} sparkles in {} ms", sparkles, elapsed / 1000000d);
```
thommy101: Is this still some debug info?

dizzyd: Yes - I left that in there while the review is ongoing. Will remove before final commit.

dizzyd:
In reference to your comment @thiakil - it was already checking if a coordinate was in the set, but it was having to check n^2 times if it was in the set, due to the traversal. The lookup wasn't _that_ expensive, but that coupled with memory allocation and the creation of a thread could push total generation time into 10s of milliseconds at times. With this setup, it's reliably 1ms or less and we avoid memory allocation and thread creation.

PR #5347, opened 2019-04-05
Port some mod compat/hook stuff

pupnewfster:
Rewrote the metallurgy ore doubling compat to use an automated system rather than having specific integration. Note: This does not work until Metallurgy releases a new build including this commit: https://github.com/Davoleo/Metallurgy-4-Reforged/commit/ebfc835a288754f3d53c9a4c6f5d346a4d0aba82 but I tested the dev build and it works properly with the version of Metallurgy that registers things at the correct times.

The rewrite of the metallurgy ore doubling removes the need for these two commits:
```
d2f0c24e5
847e78004
```

The new method also adds compat with some misc ores that previously did not have any for doubling (such as some of the ones from tech reborn).

Also ported the cyclic integration from #5201 made up of these commits:
```
3456396d7
fd638902a
26abf5086
```

Rewrote parts of the hook system so that things happen at the proper time, so that recipes added by hooks can be removed by CraftTweaker instead of adding them a lot later than the recipe registry revent.

### DIFF
--- a/etc/core/mcmod.info
+++ b/etc/core/mcmod.info
@@ -21,7 +21,6 @@
       "ic2",
       "computercraft",
       "galaticraft api",
-      "metallurgy",
       "crafttweaker",
       "cofhcore",
       "cyclicmagic"
--- a/etc/core/mcmod.info
+++ b/etc/core/mcmod.info
@@ -23,7 +23,8 @@
       "galaticraft api",
       "metallurgy",
       "crafttweaker",
-      "cofhcore"
+      "cofhcore",
+      "cyclicmagic"
     ],
     "requiredMods": [
       "forge@[14.23.2.2639,)"
--- a/src/main/java/mekanism/common/Mekanism.java
+++ b/src/main/java/mekanism/common/Mekanism.java
@@ -327,6 +327,7 @@ public class Mekanism {
         event.getRegistry().register(new BinRecipe());
         addRecipes();
         OreDictManager.init();
+        hooks.hookRecipes();
     }
 
     /**
@@ -952,6 +953,8 @@ public class Mekanism {
                     .setTranslationKey("obsidian"));
 
         Capabilities.registerCapabilities();
+
+        hooks.hookPreInit();
     }
 
     @EventHandler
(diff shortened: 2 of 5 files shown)


thiakil:
Ok, so things are this way for a couple of reasons
 - so we *can* catch some mods who don't register things immediately
 - because integrations with other mods are not supposed to happen until PostInit - *that's the entire purpose of that phase*
 - Crafttweaker things aren't acted upon *until the end of our post init* - so talking about moving things so CT can process it should't even be a thing, unless of course you broke that

The recipe event is needed for crafting table recipes only. Nothing else *has* to be there.

pupnewfster:
I will respond to this better later/when I have more time. But I didn't break the CT integration, it was more just poor wording on my part where I meant more in general if API things try to remove recipes. The CraftTweaker integration still properly handles all the extra processing and matching when it is ran rather than immediately. I probably will actually end up moving at least *some* of the recipe hook stuff back to post init as I did not realize that the point of that stage is for things that rely on other mods. Though I likely will end up keeping the cyclic integration earlier as that does not rely on communicating with cylcic at all and just has to do with if the mod is loaded; and personally I believe where possible (even if it doesn't *have* to be there) it is cleaner to register our custom recipes as well during the recipe registry event.

thiakil:
I don't even know if anyone uses the api removal anyway, its from a time before we had CT integration (and if someone wants to use it, their testing should reveal to them at what point it works or not)

I'm a bit sceptical about using `ore*`, it looks like you may be creating combiner recipes using cobble for nether based ores too?

> and personally I believe where possible (even if it doesn't have to be there) it is cleaner to register our custom recipes as well during the recipe registry event.

Sure, *our* items are fine there, but other mods won't be reliably, since MC mods are held together with duct tape sometimes :P

pupnewfster:
> I'm a bit sceptical about using `ore*`, it looks like you may be creating combiner recipes using cobble for nether based ores too?

Ya I actually have changed how I am doing it quite a bit in my local branch as I found some other bugs with the system I was using. So doing a bunch of stuff to try and make things a lot more dynamic in general and have good calculated defaults. It probably will take a couple days for me to finish up though as there are a lot of possibilities and math I need to double check, and then clean up the code for.

pupnewfster:
Closing this for now, I will do a more direct port of those commits, and aim for v10 at trying to make things a lot more dynamic.

PR #5353, opened 2019-04-06
When creating fluids from gases set their alpha value to no transparency

pupnewfster:
Tints for gases are int he format `0xRRGGBB` whereas for fluids the format is `0xAARRGGBB`. This means that mods like Blood Magic, Tinker's Construct, Ceramics, etc that obey they alpha value of fluids for rendering would render our liquids as invisible.

At some point it may make sense for us to also obey the alpha value for when we render fluid's but I am currently unaware of any fluids that actually make use of this feature.

### DIFF
--- a/src/main/java/mekanism/api/gas/Gas.java
+++ b/src/main/java/mekanism/api/gas/Gas.java
@@ -271,8 +271,8 @@ public class Gas {
                 int tint = getTint();
                 //Fluids use ARGB so make sure that we are not using a fully transparent tint.
                 // This fixes issues with some mods rendering our fluids as invisible
-                if ((tint >> 24 & 0xFF) == 0) {
-                    tint = 0xFF << 24 | tint;
+                if ((tint & 0xFF000000) == 0) {
+                    tint = 0xFF000000 | tint;
                 }
                 fluid = new Fluid(name, getIcon(), getIcon(), tint);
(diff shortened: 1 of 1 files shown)


src/main/java/mekanism/api/gas/Gas.java
```diff
@@ -268,7 +268,13 @@ public Fluid getFluid() {
     public Gas registerFluid(String name) {
         if (fluid == null) {
             if (FluidRegistry.getFluid(name) == null) {
-                fluid = new Fluid(name, getIcon(), getIcon(), getTint());
+                int tint = getTint();
+                //Fluids use ARGB so make sure that we are not using a fully transparent tint.
+                // This fixes issues with some mods rendering our fluids as invisible
+                if ((tint >> 24 & 0xFF) == 0) {
+                    tint = 0xFF << 24 | tint;
```
thiakil: Why the bitshifting? just use `0xFF000000` for OR & AND?

pupnewfster: Will change it tomorrow, I mainly copied the bitshifting for checking alpha from how blood magic was calculating the alpha and my mind was blanking for combining values.

dizzyd:
Pending the change to the bit shifting, LGTM.

PR #5360, opened 2019-04-08
Make many more Mekanism fluid renderers Stack aware for NBT purposes

pupnewfster:
Port of `72251446f`


thiakil:
image

Need to double check the gas display here, this was reported to me

thiakil:
A good way to test this BTW is to use Thermal Expansion's Alchemical Imbuer. It uses a single potion fluid with NBT

pupnewfster:
The osmium color looks fine for guis now and it certainly is easier to see.

PR #5364, opened 2019-04-08
Disable some capabilities in missed spots

pupnewfster:

- Don't expose the item handler capability in some missed spots where there never are any slots for it.
- Changes Oredictionificator to be able to accept/extract from any side except the front, instead of connect to on any side but only insert on the left and extract on the right.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -124,7 +124,9 @@ dependencies {
     }
     compileOnly "dan200.computercraft:ComputerCraft:${computercraft_version}:api"
     compileOnly "mcjty.theoneprobe:TheOneProbe-1.12:${top_version}:api"
-    compileOnly "appeng:appliedenergistics2:${ae2_version}:api"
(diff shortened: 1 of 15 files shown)


thiakil:
Looks good on the surface (minor niggle that it could have maybe been a rebase instead?)

Have you tested with null side requests for itemhandlers? Easiest is historically The One Probe's inventory mode

pupnewfster:
> Looks good on the surface (minor niggle that it could have maybe been a rebase instead?)

I realized that I did it off a previous branch by accident when I was making the PR

> Have you tested with null side requests for itemhandlers? Easiest is historically The One Probe's inventory mode

I forgot about null sided itemhandler requests, however I just tested with TOP's inventory mode and it seems to be able to show it correctly so I guess it is working properly. (I should probably double check it at some point but for now I think it is fine.)

thiakil:
Yep, if it shows the complete contents it should be good as far as I recall (we wrote the special read-only NullItemHandler for that specifically)

PR #5366, opened 2019-04-09
Put recipe values as an immutable list

pupnewfster:

- Port of `5d205e725`


thiakil:
Context: not *necessary*, but makes it a touch harder for non-Mekanism code to screw with it

pupnewfster:
The reason I hadn't did that initially was it isn't *necessary* but on second thought I realized it doesn't make sense for someone to try to add a recipe type anyways to us instead of handling it themselves.

PR #5372, opened 2019-04-09
Add some null safety for EnergyWrapperAcceptor(s).

pupnewfster:
- Port of `588fc13ce`
- Change order of attempted energy types (in EnergyAcceptorWrapper) to:
  - Ours
  - Forge
  - RF
  - Tesla
  - IC2


thiakil:
Side note: we can probably make more improvements with the cap utils. `hasCapability` is not really necessary if we're just going to use `getCapability`, so either doing a combined test for null & assignment or using some J8 magic with lambdas

PR #5374, opened 2019-04-09
Readd voice server

pupnewfster:

- Add back the voice server (context can be found here: https://github.com/mekanica/mekanica/pull/137 ) disabled by default
- Cleaned the voice server up so that all the variables in it are not public
- Voice connections now keep track of the uuid of a player rather than the username. (Shouldn't matter given they have to reconnect when changing usernames but it is better to be consistent)

### DIFF
--- a/src/main/java/mekanism/client/ClientConnectionHandler.java
+++ b/src/main/java/mekanism/client/ClientConnectionHandler.java
@@ -21,8 +21,7 @@ public class ClientConnectionHandler {
                     MekanismClient.voiceClient = new VoiceClient("127.0.0.1");
                     // Will probably not work when multiple integrateds are running on one computer
                 } catch (Throwable e) {
-                    Mekanism.logger.error("Unable to establish VoiceClient on local connection.");
-                    e.printStackTrace();
+                    Mekanism.logger.error("Unable to establish VoiceClient on local connection.", e);
                 }
             } else {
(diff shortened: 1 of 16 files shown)


src/main/java/mekanism/client/ClientConnectionHandler.java
```diff
@@ -0,0 +1,39 @@
...
+    public void onConnection(ClientConnectedToServerEvent event) {
+        if (general.voiceServerEnabled) {
+            if (event.isLocal()) {
+                // If the client is connecting to its own corresponding integrated server.
+                try {
+                    MekanismClient.voiceClient = new VoiceClient("127.0.0.1");
+                    // Will probably not work when multiple integrateds are running on one computer
+                } catch (Throwable e) {
+                    Mekanism.logger.error("Unable to establish VoiceClient on local connection.");
+                    e.printStackTrace();
```
thommy101: You can combine these, log the throeable with the error.

pupnewfster: Thanks seems I missed a couple spots that it could be collapsed in.

thiakil:
Looks fine at a surface level, I'll do a more thorough check later

pupnewfster:
Confirmed with Aidan and the reason the config option is different is so that it defaults to disabled in existing installs.

PR #5376, opened 2019-04-10
Port a couple missed commits

pupnewfster:
- Port `d303295` Move all version number references to gradle based
- Port `ebb0c46` don't set main multiblock position to air if its not loaded or not the right tile. Log if its not null but not what we expect


pupnewfster:
I think it is relatively clear that `999.999.999` is not an actual version number and is getting replaced, though maybe if/when we add a version checker back (using forge's API) we may decide to add the comment for extra clarity.

PR #5378, opened 2019-04-10
Port config rewrite

pupnewfster:
- Port a large number of commits and their changes to the config system.

I still need to test it a bunch and make sure everything works properly, but it compiles and correctly can generate the config from scratch.


src/main/java/mekanism/common/config/ClientConfig.java
```diff
@@ -0,0 +1,54 @@
...
+          "Christmas/New Years greetings in chat.");
+
+    public final DoubleOption baseSoundVolume = new DoubleOption(this, "client", "SoundVolume", 1D,
+          "Adjust Mekanism sounds' base volume. < 1 is softer, higher is louder.");
+
+    public final BooleanOption machineEffects = new BooleanOption(this, "client", "MachineEffects", true,
+          "Show particles when machines active.");
+
+    public final BooleanOption enableAmbientLighting = new BooleanOption(this, "client", "EnableAmbientLighting", true,
+          "Should active machines produce block light. Causes chunk redraws!");
```
thiakil: Is this comment still accurate after the lighting changes you guys did?

src/main/java/mekanism/common/config/options/EnumOption.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public void load(Configuration config) {
+        final Property prop = config.get(this.category, this.key, this.defaultValue.name(), this.comment);
+
+        prop.setRequiresMcRestart(this.requiresGameRestart);
+        prop.setRequiresWorldRestart(this.requiresWorldRestart);
+
+        this.value = Enum.valueOf(this.enumClass, prop.getString());
+
+        //TODO: Special loading for enums, such as to default to RF and also allow aliases
```
thiakil: imo the aliases aren't necessary, only the enum names are ever saved so idk how someone would have the other names in the config. Default changing is just changing the value in the constructor call

pupnewfster: It isn't quite that simple because how we changed it to try and make it so you have to fully specify "joules". Also the aliases is in terms of the temp ones more.

src/main/java/mekanism/common/content/entangloporter/InventoryFrequency.java
```diff
@@ -21,8 +21,8 @@
 
     public static final String ENTANGLOPORTER = "Entangloporter";
 
-    public static final int FLUID_TANK_SIZE = Tier.FluidTankTier.ULTIMATE.output;
-    public static final int GAS_TANK_SIZE = Tier.GasTankTier.ULTIMATE.output;
+    public static final int FLUID_TANK_SIZE = Tier.FluidTankTier.ULTIMATE.getOutput();
```
thiakil: Looks like these constants should be removed and their usages replaced by the config read

src/main/java/mekanism/common/config/ClientConfig.java
pupnewfster: You are right it doesn't still cause chunk redraws, I will fix it in the next commit I make.

src/main/java/mekanism/common/config/options/EnumOption.java
pupnewfster: Though I am fine with just changing the default value for new configs if that is preferable.

thiakil: > It isn't quite that simple because how we changed it to try and make it so you have to fully specify "joules".

Could you elaborate on this?

thiakil:
On the whole, looks fine to me at the moment

src/main/java/mekanism/common/config/ClientConfig.java
thiakil: hmm just looked into it, `mekanism.common.block.BlockMachine#getLightValue` does still change block light, which causes redraws - its just less often

src/main/java/mekanism/common/config/options/EnumOption.java
pupnewfster: I think I will just change the default, except what I had meant is we had changed it so that "joules" has to be specified in the config otherwise it defaults to displaying "RF" as that is the number people are more familiar with what it means. Though I think it is fine until we switch internally to using FE that it shows Joules for the old installs.

thiakil: Yeah since it's display, probably best to not change existing installs.

As for changing to FE internally, that's probably something we need to discuss since some parts might need the Joules based storage

src/main/java/mekanism/common/config/ClientConfig.java
pupnewfster: It is a lot less often, for Mekanica I didn't have the comment describing that it causes chunk redraws. As with the rate it does it at, it shouldn't really be something for people to worry about.

PR #5379, opened 2019-04-10
improve bio generator rendering slightly

temp1011:
improve rendering of bio generator model when fuel level is low

The GUI output also flickers annoyingly but I have yet to work out how to solve that (seems it may be that the while packet of energy generated is output at once)


thiakil:
Code looks fine, do you have a simple test setup for this?

temp1011:
I just put a small amount of bio fuel (1 or 2) in a generator and looked at it until it ran out.

PR #5380, opened 2019-04-11
port tesr pull

temp1011:
Everything works correctly now (thanks @pupnewfster for fixing the client sync).
Still need to decide about the CTM stuff. It will be possible in our code, but a bit more complex.


src/main/java/mekanism/common/tile/prefab/TileEntityEffectsBlock.java
```diff
@@ -32,7 +32,7 @@
     private long lastActive = -1;
 
     // Number of ticks that the block can be inactive before it's considered not recently active
-    private final int RECENT_THRESHOLD = 100;
+    private final int RECENT_THRESHOLD;
```
pupnewfster: The changes to this file are not needed

PR #5327, opened 2019-03-27
implement #4688 and implement #5236

ToMe25:
Disable Free Runners while Sneaking and add Sawing Factories.

PR #5383, opened 2019-04-11
Port ToMe25's PR for sawing factories

pupnewfster:
- Port of #5327 I figured it would be quicker to manually port the PR rather than explain the slight differences done to the upgrading system and other parts of mekanism.

@ToMe25 if you could double check this branch works as intended that would be great. I inlined a helper method of yours and also added the missing blockstate files as the blocks and item variants were untextured.

I still need to test this more myself, and I am not quite sure about how clear it will be to people that if they want to multiprocess using the sawing factories the secondary outputs of all items need to be the same or one stack won't be processed. Though at least on the surface everything seems to be working properly and looks good.


pupnewfster:
Everything seems to work properly to me, so I believe this can be merged once a second set of eyes reviews it.

PR #5384, opened 2019-04-11
Port #4952 by @danielzgtg Item drop compatibility

pupnewfster:
## Changes proposed in this pull request since #4952:
- Use parent super classes to reduce duplicate code being added
- Fix `BlockBounding` not being removed when state forcefully changed.


pupnewfster:
Haven't done all that much testing on this yet, but I went through and created a couple intermediary classes so that it the dif is easier to read and there is less duplicate code/overrides happening.

I think for the most part it is fine now, though I am not to sure if the implementation for block bounding is proper.

pupnewfster:
Did some cleanup and I think BlockBounding does things properly now. It also now properly invalidates/breaks the "multiblock" if something changes the state of one of the blocks, such as with the `/setblock` command.

pupnewfster:
> Code looks good (though as mentioned on discord we can probably get rid of `BlockContainer` from the hierarchy), no bugs found in testing

Going to leave it be for now as changing the overall hierarchy that much has a larger chance of breaking something, and it will be easier to track down if it caused issues if it is in a separate release.

PR #5386, opened 2019-04-12
Avoid scheduling a sound to be played if it's fully muffled.

dizzyd:
Per the commentary on https://minecraft.curseforge.com/projects/mekanism?comment=230 - avoid scheduling any sounds for a machine that is fully muffled.

### DIFF
--- a/src/main/java/mekanism/common/tile/prefab/TileEntityEffectsBlock.java
+++ b/src/main/java/mekanism/common/tile/prefab/TileEntityEffectsBlock.java
@@ -86,11 +86,10 @@ public abstract class TileEntityEffectsBlock extends TileEntityElectricBlock imp
 
             // If this machine isn't fully muffled and we don't seem to be playing a sound for it, go ahead and
             // play it
-            if (!isFullyMuffled() && (activeSound == null || !Minecraft.getMinecraft().getSoundHandler().isSoundPlaying(activeSound))) {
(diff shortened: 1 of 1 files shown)


src/main/java/mekanism/common/tile/prefab/TileEntityEffectsBlock.java
```diff
@@ -82,10 +84,17 @@ private void updateSound() {
...
 
-            if (activeSound == null || !Minecraft.getMinecraft().getSoundHandler().isSoundPlaying(activeSound)) {
+            // If this machine isn't fully muffled and we don't seem to be playing a sound for it, go ahead and
+            // play it
+            if (!isFullyMuffled() && (activeSound == null || !Minecraft.getMinecraft().getSoundHandler().isSoundPlaying(activeSound))) {
                 activeSound = SoundHandler.startTileSound(soundEvent.getSoundName(), getInitialVolume(), getPos());
-                playSoundCooldown = 20;
             }
+
+            Mekanism.logger.info("Fully muffled: {}", isFullyMuffled());
```
thiakil: lingering debug?

pupnewfster: I think it was for purposes of when we tested, I just removed it though and the changes seem to work fine.

PR #5387, opened 2019-04-12
Pipe none connection

thiakil:

- Port of Thommy101's NONE connection between pipes changes
- Couple of Array index checks, borne out of not having the sawmill changes on the branch but nbt with it (can happen in cases of corrupted nbt too)

## Outstanding issues

- [x] Visual connection update / auto-NONE on incompatible networks
- [x] Test more network recalculations for voiding
- [x] Bump api version
- [x] Disconnected networks (from incompatible networks) do not like reconnecting even once the contents area drained.

### DIFF
--- a/src/main/java/mekanism/api/transmitters/IGridTransmitter.java
+++ b/src/main/java/mekanism/api/transmitters/IGridTransmitter.java
@@ -64,4 +64,6 @@ public interface IGridTransmitter<A, N extends DynamicNetwork<A, N>> extends ITr
     void updateShare();
 
     Object getBuffer();
+
+    boolean isCompatibleWith(IGridTransmitter<A,N> other);
 }
--- a/src/main/java/mekanism/api/transmitters/IGridTransmitter.java
+++ b/src/main/java/mekanism/api/transmitters/IGridTransmitter.java
@@ -5,7 +5,7 @@ import mekanism.api.Coord4D;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.world.World;
 
-public interface IGridTransmitter<ACCEPTOR, NETWORK extends DynamicNetwork<ACCEPTOR, NETWORK>> extends ITransmitter {
+public interface IGridTransmitter<ACCEPTOR, NETWORK extends DynamicNetwork<ACCEPTOR, NETWORK, BUFFER>, BUFFER> extends ITransmitter {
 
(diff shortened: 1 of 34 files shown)


pupnewfster:
This still actually has the issue where fluid/gas pipes can connect if their contents are not the same as if you place a pipe between two different networks the checks say that it has null contents so can connect to both networks, and then the network gets polluted again.

pupnewfster:
Also another bug (I think more having to do with the not having sawmill changes in this branch), when I landed with invalid factories (the sawmill ones) in my inventory on a chargepad my game crashed. Probably because they still have the NBT that makes it think they can be charged.

thiakil:
Progress, but some orphans still seem to merge incorrectly.

Breaking the highlighted pipe will cause the one above it to turn green

src/main/java/mekanism/api/transmitters/IGridTransmitter.java
```diff
@@ -64,4 +64,6 @@
     void updateShare();
 
     Object getBuffer();
+
+    boolean isCompatibleWith(IGridTransmitter<A,N> other);
```
pupnewfster: We should probably bump the api version in https://github.com/mekanism-mod/Mekanism/blob/1.12/src/main/java/mekanism/api/transmitters/package-info.java to at least `9.1` or something given this is a change in API.

pupnewfster: We should probably bump the api version in https://github.com/mekanism-mod/Mekanism/blob/1.12/src/main/java/mekanism/api/transmitters/package-info.java to at least 9.1 or something given this is a change in API.

thiakil: agreed

thiakil: I can add a default to the interface too

pupnewfster: Sounds good, not aware of anyone else using this part of our API but no reason to break things on people when we don't have to.

pupnewfster:
I believe the network voiding part is now fixed, though we still need to bump the API version and update the visual connections

thiakil:
Seems to work ok now, I made it so you can change the mode for unconnected pipes too

Issue #4698, opened 2017-09-18
Mekanism not using domain in localization files.


halvors:
#### Issue description:
Mekanism not using domain in localization files. This causes translations to clash with other mods, as well as other mods to clash with mekanism.

One example is "fluid.tritium" which is translated to "Tritium" by other mods, and to "Liquid Tritium" by mekanism. The result is that the translation that is loaded first, wins.

thiakil:
Additional conflict key `gui.delete` (see linked issue)

thommy101:
Started a branch for this issue, trying to make all localisation key Mekanism specific.

thiakil:
from #5111 

`gui.delete`
- customnpcs `Are you sure you want to delete this?`
- mekanism `Delete`

`gui.volume`
in other language there are different words
- extraambiance `Volume` <- sound
- mekanism `Volume` <- cubage

`gui.active`
- ftblib `Active`
- mekanism `Water-Cooled`

pupnewfster:
All lang strings properly include mekanism (or the submodules name, such as mekanismgenerators), in their entries as of the 1.15 versions.

PR #5389, opened 2019-04-13
Keybindings optimisations

thiakil:

- Shifts to a 'builder' style method of supplying keybindings, removing that array length issue
- Tidying up the handler method a little; using proper text components rather than a string concatenation (not too important here since its client side already)
- Keybindings use lang key properly, prefixed with `mekanism.`
- Changed the translation keys used in the key handler to use `String.format` placeholders so languages can better decide where the subject goes
- Adds coremod for keybinding remap
- Uses a closure in `build.gradle` to use the same manifest params block for the all in one & core jar
- `ItemConfigurator.ConfiguratorMode` changed to use `TransmissionType` instead of string, with ability to output a TextComponent

Side note: I'm not even sure the voice key needs to be marked as repeatable, since its not handled here


src/main/java/mekanism/client/MekanismKeyHandler.java
```diff
@@ -43,20 +47,23 @@
...
     public static final String keybindCategory = Mekanism.MOD_NAME;
-    public static KeyBinding modeSwitchKey = new KeyBinding("Mekanism " + LangUtils.localize("key.mode"),
-          Keyboard.KEY_M, keybindCategory);
-    public static KeyBinding armorModeSwitchKey = new KeyBinding("Mekanism " + LangUtils.localize("key.armorMode"),
-          Keyboard.KEY_G, keybindCategory);
-    public static KeyBinding voiceKey = new KeyBinding("Mekanism " + LangUtils.localize("key.voice"), Keyboard.KEY_U,
-          keybindCategory);
-    public static KeyBinding freeRunnerModeSwitchKey = new KeyBinding("Mekanism " + LangUtils.localize("key.feetMode"),
-          Keyboard.KEY_H, keybindCategory);
+    public static KeyBinding modeSwitchKey = new KeyBinding("key.mode", Keyboard.KEY_M, keybindCategory);
```
pupnewfster: might want to do something like "key.mekanism.mode" just because I feel like "key.mode" may not be an uncommon name for a keybinding. Given I am pretty sure it saves this stuff in the options.txt file for keybindings.

thiakil: yeah we definitely need to do that as part of solving #4698
Will check what impact that has on options.txt as part of this

pupnewfster: I think at the very least changing the keybind will not save previously saved keybinding settings. If I remember correctly our current way is not ideal anyways as it has a space in the path.

thiakil: Yeah, I'll write a datafixer for it

src/main/java/mekanism/coremod/KeybindingMigrationHelper.java
```diff
@@ -0,0 +1,65 @@
...
+        public MethodVisitor visitMethod(int access, String name, String desc, String signature, String[] exceptions) {
+            MethodVisitor visitor = super.visitMethod(access, name, desc, signature, exceptions);
+            if (name.equals(datafixMethodName) && desc.equals(DATAFIX_SIG)){
+                LOGGER.info("Patching GameSettings.datafix");
+                return new GeneratorAdapter(Opcodes.ASM5, visitor, access, name, desc) {
+                    @Override
+                    public void visitCode() {
+                        super.visitCode();
+                        loadArg(0);
+                        invokeStatic(Type.getObjectType("mekanism/common/fixers/KeybindingFixer"), new Method("runFix", "(Lnet/minecraft/nbt/NBTTagCompound;)V"));
```
pupnewfster: What is the point of doing this as invokeStatic instead of just calling the method directly? Does it have to do with the coremod not "actually" knowing anything about Mekanism's code?

thiakil: what do you mean 'calling the method directly'?

pupnewfster: `KeybindingFixer.runFix()`

pupnewfster:
Going ahead and merging this as upon looking at it a bit closer, I believe at the point we have the fixer get called saving the file could potentially even cause corruption/data desync. Given it can still read things properly and has them in memory properly, them not being stored to file until controls are viewed is fine.

Issue #5394, opened 2019-04-13
Server tick loop TileEntityFactory


bigenergy:
*Please use the search functionality before reporting an issue. Also take a look at the closed issues!*

#### Issue description:
Server crash

#### Steps to reproduce:
unknown

#### Version (make sure you are on the latest version before reporting):
**Forge:** 2814
**Mekanism:** 9.7.0.369

#### If a (crash)log is relevant for this issue, link it here: (_It's almost always relevant_)

https://pastebin.com/T0VbGuC6

pupnewfster:
Duplicate/same crash/cause as #5392

Issue #5392, opened 2019-04-13
Server crash Ticking GridNode


bigenergy:
#### Issue description:
Server crash on start

#### Steps to reproduce:
1. Start server
2. Get crash

#### Version (make sure you are on the latest version before reporting):
**Forge:** 2814
**Mekanism:** 9.7.0.369

#### If a (crash)log is relevant for this issue, link it here: (_It's almost always relevant_)

https://pastebin.com/7xNSfC1c

pupnewfster:
It seems I forgot the fact when I was reusing code about it being possible to attempt to insert things to a machine that do not have a recipe. I will take a look into this later today, as well as, improve the performance a bit more of autosort.

thiakil:
released on https://minecraft.curseforge.com/projects/mekanism/files/2699534

PR #5399, opened 2019-04-14
Factory recipe check

pupnewfster:
- Improve the recipe check for if an item can go in an input slot based on the current output to use the cached recipes. This updates the cached recipe if it seems to be out of date with the information we know.
- Also fixes the null pointer in #5392 and #5394 caused by attempting to get the recipe output of a null recipe. (For when items attempt to be inserted if they don't have a recipe that they match)
- Include some partially implemented code for if we ever have PRC factories to make it easier to have them support this recipe check.


thiakil:
Are factories updating setting the cached recipe on the call to `canInsertItem`?

pupnewfster:
> Are factories updating setting the cached recipe on the call to `canInsertItem`?

Good catch when I was cleaning it up slightly I changed the canInsertItem back to using the method forgetting that it shouldn't be updating the cached recipe.

Issue #5397, opened 2019-04-14
[Suggestion] CraftTweaker getGas utility method


justinrusso:
**Describe the feature you'd like**
Essentially, I'd love for https://github.com/mekanism-mod/Mekanism/blob/1.12/src/main/java/mekanism/common/integration/crafttweaker/gas/GasBracketHandler.java#L33 this method to be exposed to crafttweaker. This will allow you to get a Gas more dynamically in scripts.

thiakil:
Released on https://minecraft.curseforge.com/projects/mekanism/files/2699534

PR #5400, opened 2019-04-14
Implement #5397 add a way to get CrT gas from string

pupnewfster:
- Exposes `getGas` via the call `mods.mekanism.mekaTypeUtils.getGas(string);`
This change is in its own class so that eventually if/when we add more types (such as making infuse type a proper type instead of a string), then we can add ways to get them in this class.


thiakil:
I'd call it `mods.mekanism.GasRegistry` in CT if it were me

Issue #4676, opened 2017-09-05
Mekanism cable networks are breaking when leaving an area and re-entering it.


RileyGuy:
# Note from dev: Please do _not_ add a 'me too' to this issue. 
If you wish to subscribe for updates, use the button in the sidebar. It is a known issue that we have not figured out. Of course it's going to affect you.

------------------------------------------------------------------------

#### Issue description:
Mekanism cable networks cease to exist when leaving an area and coming back, chunk-loaded or not.
Breaking one of the cables doesn't work as it doesn't transmit power to the adjacent cables. I have to break all of them to have a complete network again.

#### Steps to reproduce:
1. Place mekanism cables to form a cable network.
2. Leave the area they are in (at least 32 chunks away).
3. Coming back, the cables will have ceased working and the network will no longer be existent. You will have to replace all cables for it to function again.

**Forge:** 2460
**Mekanism:** 9.4.1
**Other relevant version:** 

The crashlog here is ***NOT*** relevant in any way, I'm just posting it so you can see the modlist (the mods haven't changed but they have been updated recently to their latest versions with no impact on this bug whatsoever).
https://gist.github.com/Xerasin/3a52eae8f6d920eb97510f08eeb0bb05

thommy101:
Could you please provide a log file when this is happening? It is certainly relevant as that does provide versions you're using.. Giving an old one and saying, I only updated everything after it isn't so helpful.

I'm almost certain that this issue has been fixed in the latest release.

If this does happen, could you please provide some information on the transmitter networks of the misbehaving transmitters? You can get that with the network probe.

RileyGuy:
Ah. You see, I don't actually have a log file as I don't own the server, I'll contact the owner about it. However I did scan with the network probe and when the cables are functioning normally, it reports the correct amount of energy receivers and givers. When they break though, the network probe reports that no network is formed whatsoever. I also think that this is only happening when they are connected to entangloporters, because I have cables that are receiving power through different means which are fine when the others break.

RileyGuy:
Oh by the way, just as an update this seems to happen with item pipes as well, which leads me to believe there could be more pipes that have this issue.

thommy101:
All pipes use the same logic ;) It's very related to a long lasting bug, which we tought was fixed in one of the latest releases (certainly in 1.12), apparently it's not fixed completely...

Do you have a usecase that is always reproducable? That makes debugging a lot easier.

RileyGuy:
[...]

PR #5407, opened 2019-04-15
Various fixes to data not loading properly

pupnewfster:
1. Fix capacity of fluid tanks sometimes being incorrect when loading an empty fluid tank.
2. Make transmitter networks update and correct and broken connection data on load. (Probably fixes any broken networks/breaking networks seen in #4676 )
3. Properly mark the transmitters as dirty if the redstone mode changed during refresh:

Before point 3 for example with the setup seen below, if you turned on redstone sensitivity for the middle pipe, flicked the lever, and then teleported a thousand blocks away. Teleporting back and disabling the redstone sensitivity (while the lever is still flicked) would get it into the broken state seen below. This also could happen where if you had a redstone signal going to the middle pipe (from a non broken state) and toggled the middle pipe to being redstone sensitive it would visually disconnect it from the other pipes except water would continue to flow through the setup.

### DIFF
--- a/src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
+++ b/src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
@@ -412,7 +412,7 @@ public abstract class TileEntitySidedPipe extends TileEntity implements ITileNet
             }
             //If the redstone mode changed properly update the connection to other transmitters/networks
             if (previouslyPowered != redstonePowered) {
-                markDirtyTransmitters();
+                notifyTileChange();
             }
 
             redstoneSet = true;
@@ -490,7 +490,7 @@ public abstract class TileEntitySidedPipe extends TileEntity implements ITileNet
         onWorldJoin();
         if (getPossibleTransmitterConnections() != currentTransmitterConnections) {
             //Mark the transmitters as invalidated if they do not match what we have stored/calculated
-            markDirtyTransmitters();
+            refreshConnections();
         }
         super.onLoad();
     }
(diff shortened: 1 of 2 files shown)


src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
```diff
@@ -479,6 +488,10 @@ public void onAdded() {
     @Override
     public void onLoad() {
         onWorldJoin();
+        if (getPossibleTransmitterConnections() != currentTransmitterConnections) {
+            //Mark the transmitters as invalidated if they do not match what we have stored/calculated
+            markDirtyTransmitters();
```
thiakil: I'm not sure `markDirtyTransmitters` is the right method to call here, as it mostly notifies neighbours of a change. Perhaps `refreshConnections()` would be more appropriate?

Seems otherwise it's relying on an adjacent transmitter to call the updateNeighbour path

src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
```diff
@@ -401,13 +403,20 @@ protected void onRefresh() {
...
+        if (handlesRedstone()) {
+            boolean previouslyPowered = redstonePowered;
+            if (redstoneReactive) {
+                redstonePowered = MekanismUtils.isGettingPowered(getWorld(), new Coord4D(getPos(), getWorld()));
+            } else {
+                redstonePowered = false;
+            }
+            //If the redstone mode changed properly update the connection to other transmitters/networks
+            if (previouslyPowered != redstonePowered) {
+                markDirtyTransmitters();
```
thiakil: Probably better to directly call `notifyNeighbours`

src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
pupnewfster: Your right `refreshConnections()` does seem to behave a lot better here.

src/main/java/mekanism/common/tile/transmitter/TileEntitySidedPipe.java
pupnewfster: Now calls `notifyTileChange()` which seems to be a helper method to keep it cleaner when notifying neighbors.

PR #5410, opened 2019-04-16
improve texture of tank break particle

temp1011:
Makes break particles on fluid tank slightly darker so they don't get confused with missing texture, hopefully an improvement.


pupnewfster:
I think this looks better when it is using `#glass` instead of `#frame`

pupnewfster:
Contained as part of #5411

PR #5411, opened 2019-04-16
Particle fixes

pupnewfster:
- Make transmitters return an actual particle instead of the default WHITE
- Make fluid tank particle/break particle be the glass wall
- Have hit particle for fluid tank and transmitters use the correct tier instead of being basic.
  - Running and Falling particle still displays as the basic tier


pupnewfster:
The falling and running particles for transmitters are now showing the basic tier particles instead of the white particles. I am not going to struggle with getting the particles to show the proper tiers in those specific cases, as the issue should go away once we stop handling tier data once we get rid of NBT. The hit particle for fluid tanks and transmitters are now properly showing the correct particles and tiers.

PR #5416, opened 2019-04-19
use getters & setters for InfuseStorage

### LINKED ISSUE #5408: Machines again stuck at 0 while dumping
Forge: 2816
Mekanism-1.12.2-9.7.1.371
MekanismGenerators-1.12.2-9.7.1.371

Have a Metallurgic Infuser with Obsidian that I can't empty completely. It stays at zero. I had this issue before but it seems to be back. Have to break it and put it back seem to fix it.

Edit: Says 0 then but still can't put stuff in.

Edit: I'm dumb. Works after breaking the machine

thiakil:

Makes InfuseStorage members private and tries to handle an empty state from either field. Should fix #5408 / ensure it never happens


pupnewfster:
The infuser *seems* to work properly, however the infusing factory's progress bars are screwed up and it is working without any infusion type in it.

I have a basic factory with 8 speed upgrades and one piece of coal and it seems to be able to process infinite iron.

pupnewfster:
I believe it all works properly now, except the infuse column in the metallurgic infuser GUI seems darker than I remembered. I don't think that was caused by this PR, but we may want to change it as it is somewhat hard to see if there is CARBON in it or not.

PR #5418, opened 2019-04-19
Add support for configurable machines storage.

### LINKED ISSUE #5367: Config Option For Energy Storage In A Machine
A way to change the amount of energy a machine can hold in a config option (maybe similar to the usage config).  This would be useful since if you increase the usage of the machines the factories are then unable to process all the items without energy upgrades due them not having enough energy storage

Zajozor:
fixes #5367 

I added a configuration category called storage, which is loaded both on client and server.
Created default values for the storage, for most of the machines, that were in the switch in `getUsage()` function.
Notably added an option for `TELEPORTER` and `CHARGEPAD`, which had hard-wired values in `getUsage()`. Using the previous method of determining storage as default.

As suggested by @thiakil , the code does:  `baseEnergy = Math.max(usage, storage);` 
This is not perfect though, if you put some speed upgrades into the machine, you can still get it into a state when it requires more energy, than can hold, or when the energy use increases due to some settings (eg. using silk touch in digital miner).

I see no straightforward way how to fix these machine issues (which can anyway happen only with very unusual config values). If you would like it to work somehow, feel free to suggest me something.
I did test the changes and they seem to work as expected, other than what was mentioned.

### DIFF
--- a/src/main/java/mekanism/common/block/states/BlockStateMachine.java
+++ b/src/main/java/mekanism/common/block/states/BlockStateMachine.java
@@ -195,7 +195,6 @@ public class BlockStateMachine extends ExtendedBlockState {
         public int meta;
         public String blockName;
         public int guiId;
-        public double baseEnergy;
         public Class<? extends TileEntity> tileEntityClass;
         public boolean isElectric;
         public boolean hasModel;
@@ -253,11 +252,6 @@ public class BlockStateMachine extends ExtendedBlockState {
             return null;
         }
 
-        public static void updateAllUsages() {
-            for (MachineType type : values()) {
-                type.updateUsage();
-            }
-        }
 
         public static MachineType get(ItemStack stack) {
             return get(Block.getBlockFromItem(stack.getItem()), stack.getItemDamage());
@@ -287,31 +281,28 @@ public class BlockStateMachine extends ExtendedBlockState {
             }
         }
 
-        /**
-         * Used for getting the base energy storage.
-         */
         public double getUsage() {
             switch (this) {
                 case ENRICHMENT_CHAMBER:
-                    return MekanismConfig.current().usage.enrichmentChamberUsage.val();
+                    return MekanismConfig.current().usage.enrichmentChamber.val();
                 case OSMIUM_COMPRESSOR:
-                    return MekanismConfig.current().usage.osmiumCompressorUsage.val();
+                    return MekanismConfig.current().usage.osmiumCompressor.val();
                 case COMBINER:
-                    return MekanismConfig.current().usage.combinerUsage.val();
+                    return MekanismConfig.current().usage.combiner.val();
                 case CRUSHER:
-                    return MekanismConfig.current().usage.crusherUsage.val();
+                    return MekanismConfig.current().usage.crusher.val();
                 case DIGITAL_MINER:
-                    return MekanismConfig.current().usage.digitalMinerUsage.val();
+                    return MekanismConfig.current().usage.digitalMiner.val();
                 case METALLURGIC_INFUSER:
-                    return MekanismConfig.current().usage.metallurgicInfuserUsage.val();
+                    return MekanismConfig.current().usage.metallurgicInfuser.val();
                 case PURIFICATION_CHAMBER:
-                    return MekanismConfig.current().usage.purificationChamberUsage.val();
+                    return MekanismConfig.current().usage.purificationChamber.val();
                 case ENERGIZED_SMELTER:
-                    return MekanismConfig.current().usage.energizedSmelterUsage.val();
+                    return MekanismConfig.current().usage.energizedSmelter.val();
                 case TELEPORTER:
                     return 12500;
                 case ELECTRIC_PUMP:
-                    return MekanismConfig.current().usage.electricPumpUsage.val();
+                    return MekanismConfig.current().usage.electricPump.val();
                 case PERSONAL_CHEST:
                     return 30;
                 case CHARGEPAD:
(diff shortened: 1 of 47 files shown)


src/main/java/mekanism/common/config/StorageConfig.java
```diff
@@ -0,0 +1,82 @@
+package mekanism.common.config;
+
+import mekanism.common.config.options.DoubleOption;
+
+public class StorageConfig extends BaseConfig {
+
+    public final DoubleOption enrichmentChamberStorage = new DoubleOption(this, "storage",
```
thiakil: I don't think these need `Storage` in the field name (as they'll always be accessed via `usage.X`, probably don't need to be in the config key name either, but that's less of an issue as it does help context there

src/main/java/mekanism/common/block/states/BlockStateMachine.java
```diff
@@ -367,7 +367,88 @@ public double getUsage() {
...
+                case FLUIDIC_PLENISHER:
+                    storage = MekanismConfig.current().storage.fluidicPlenisherStorage.val();
+                    break;
+                case LASER:
+                    storage = MekanismConfig.current().storage.laserStorage.val();
+                    break;
+                case FORMULAIC_ASSEMBLICATOR:
+                    storage = MekanismConfig.current().storage.formulaicAssemblicatorStorage.val();
+                    break;
+                default:
```
thiakil: how does this go with factories?

src/main/java/mekanism/common/config/StorageConfig.java
Zajozor: should I also remove the Usage part in here?

```
public class UsageConfig extends BaseConfig {

    public final DoubleOption enrichmentChamberUsage = new DoubleOption(this, "usage", "EnrichmentChamberUsage", 50D,
          "Energy per operation tick (Joules).");
```

src/main/java/mekanism/common/block/states/BlockStateMachine.java
Zajozor: Not well, my bad.

Working on a fix now. The factory calculations as well as some of the GUI display elements have some hardwired constants.

src/main/java/mekanism/common/config/StorageConfig.java
thiakil: yeah go for it

Zajozor:
Hey, sorry for the huge changes, the renaming got a little wild, but most files are changed really just cosmetically.
Also sorry for the force-push history, I had a little trouble getting `ItemBlockMachine.java` right, as it has  CRLF endings, which were auto deleted, leading to a heavy diff.
The branch should be rebased above 1.12 now though.

Changes:
- add the storage config with default values
- remove `Usage` and `Storage` from config field names, and renamed their respective usages
- previously, as the storage was calculated as `400*usage`, on every config sync, `BlockStateMachine.MachineType.updateAllUsages();` was called, to update `MachineType.something.baseEnergy` values. When the storage is in the config though, this is unnecessary, because the machines can get their storage directly from there. So I removed the `baseEnergy` field, added a `getStorage()` one (simillar to the existing `getUsage()`), and made the code use that, where applicable.
- Machines extending from `TileEntityMachine` and similar were accessing the `MekanismConfig` directly, and using the old `baseEnergy` field to get their usage/storage. Now, they get them through `MachineType.something.getUsage/Storage()` (hence a lot of diff). This allows to use `max(usage,storage)==actualStorage` (on a direct config access we could not alter it).
- I did allow myself to rename some method arguments from `maxEnergy` to `baseMaxEnergy`, as that is what they are really used for, and it also conforms with the existing `baseEnergyUsage` argument naming. If this is a problem, I shall revert it.

- Regarding factories, they should now work as intended -> previously `200*processes*baseUsage`, now `0.5*processes*baseStorage`, which gives same defaults. Will follow up with some screenshots, demonstrating this.
- Also added storage config option for the chargePad and Teleporter (which do not have a usage config option)

Zajozor:
Example screenshots:

When we set enriching factory to have only 10kJ = 2kRF storage, the basic factory will have 1.5x times that much.
image
And the elite one will have 3.5x that much:
image

When you set some storage to too low (eg. EnergizedSmelterUsage = 0 here), then storage gets capped to usage (20):
image
And the basic factory will have 1.5x times (30 storage), meaning it can only power one operation at a time:
image
And an advanced factory with an energy upgrade will have 50+16.67 capacity, being able to smelt 4 items at a time 15rf/t each:
image

Custom chargepad storage:
image
Custom teleporter storage:
image

As mentioned above, sometimes I get an issue, where the client ignores the custom config value, then it looks like this -> The teleporter is set to have 690kRF as maximumStorage (and won't charge any further, the server limits it correctly), but the client still displays the maximum as 2MRF. I am unable to determine when this happens, if you have any suggestions I would be glad. (Maybe just a random workspace issue)
image

thiakil:
> previously `200*processes*baseUsage`, now `0.5*processes*baseStorage`

I think the cap here should allow all processes to be used in a factory at a minimum

Your explanations are fantastic by the way, I appreciate the effort you've put into this :)

Zajozor:
Sure, will do soon. (sorry easter got on me in the last couple of days :) )
Other than that, is there anything else i should tweak or some special cases that I should test?

Zajozor:
Now the storage should behave as you suggested. So now:

When you set some storage to too low (eg. EnergizedSmelterUsage = 0 here), then storage gets capped to usage (20*tier.process), here is a basic factory:
image

An advanced factory with a energy upgrade:
image

It is still possible though, to get into a situation, when the factory cannot do all of the processes, when you apply speed upgrades and the storage is too low. Here is an elite factory with two speed upgrades:
image
It correctly consumes 
`126.5 = 2 processes * 20 baseUsage * (10 maxUpgradeMultiplier ^ (2 * 2/8 upgrades))`.

Other functionality remains unaffected.

thiakil:
Sounds good (it's always been possible to get into speed/energy upgrade mismatch problems)

I'll have a proper code review this weekend

Zajozor:
Indeed, this fixed all desync problems I can observe, thank you :)

PR #5422, opened 2019-04-21
Emitting logic overhaul

pupnewfster:
Basic explanation of how the emitting code works now. There are some small performance improvements in the code to reduce the number of calculations needed that are not described below. (This is to keep the description as clear as possible.)

There are now implementations of a Target class that keep track of the Handler's being interacted with and which side they are from. This is important because the previous implementation loop over all the sides that are connected to a specific Coord4D and try to insert into it using the first Wrapper/Handler we found for that Coord4D. We also do that for that Coord4D however many times we have a connection to it. (And energy would keep trying to force send energy up to 100 times on top of that if we had any left over.)

Next we now loop over all the targets once, (and each of their handlers), simulating how much  the wrapper can take (out of the maximum amount we are emitting). This way we can know the max amount they are willing to accept.

We then calculate an even distribution to all our handlers (more inputs target means it gets more total as it can receive it from more than one side). If one target needs less power than we are able to give it, we adjust the amount that we are willing to send to our targets so that they can get as much as possible while still fairly given all the other targets their fair share.

Finally we then go over the targets and their wrappers sending them their calculated share.

Implementation Note: When a target can be fully fulfilled by the split, we send it to it immediately, and adjust the split accordingly. This allows us to use up all the power and not get as stuck if a multiblock returns same power needed on multiple connections and we satisfy it from a single one of them.

## Other minor changes included in this PR:
- Fixed a bunch of places not clamping/rounding upwards when converting from doubles to an int for power. (This caused issues such as taking more power than we had and causing networks to go negative)
- Halved the number of loops the Heat Network did by combining the internals of the two identical for loops
- Deprecated and removed our implementations of `DynamicNetwork#getAcceptors` as we no longer use that method
- Added support for simulation to the `IC2Acceptor` `EnergyAccetporWrapper`
- Fixed simulating for the `TeslaAcceptor`


src/main/java/mekanism/common/base/EnergyAcceptorWrapper.java
```diff
@@ -140,6 +140,10 @@ public IC2Acceptor(IEnergySink ic2Acceptor) {
         @Override
         public double acceptEnergy(EnumFacing side, double amount, boolean simulate) {
             double toTransfer = Math.min(Math.min(acceptor.getDemandedEnergy(), toEU(amount)), Integer.MAX_VALUE);
+            if (simulate) {
+                //IC2 has no built in way to simulate, so we have to calculate it ourselves
+                return toTransfer;
```
thiakil: should probably be `toTransfer - acceptor.getDemandedEnergy()`

src/main/java/mekanism/common/base/EnergyAcceptorTarget.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+import net.minecraft.util.EnumFacing;
+
+public class EnergyAcceptorTarget {
+
+    private Map<EnumFacing, EnergyAcceptorWrapper> wrappers = new HashMap<>();
+    private Map<EnumFacing, Double> needed = new HashMap<>();
```
thiakil: I'd use `EnumMap` for `wrappers` and `it.unimi.dsi.fastutil.objects.Object2DoubleArrayMap#Object2DoubleArrayMap(6)` for needed/given

thiakil: Tho the double might not work after generification

src/main/java/mekanism/common/base/EnergyAcceptorWrapper.java
```diff
@@ -140,6 +140,10 @@ public IC2Acceptor(IEnergySink ic2Acceptor) {
         @Override
         public double acceptEnergy(EnumFacing side, double amount, boolean simulate) {
             double toTransfer = Math.min(Math.min(acceptor.getDemandedEnergy(), toEU(amount)), Integer.MAX_VALUE);
```
thiakil: not your code, but should we remove the MAX_VALUE cap there? since EU appears to be a `double` now

src/main/java/mekanism/common/base/target/Target.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    public Map<EnumFacing, HANDLER> getHandlers() {
+        return handlers;
+    }
+
+    /**
+     * Sends the remaining amount to each handler we still have not settled on an amount for. We increment the amount
+     * sent in splitInfo as well as adjust the split as needed if one ends up accepting less than it originally wanted.
+     * (The most likely case this would
```
thiakil: i think you missed something here

src/main/java/mekanism/common/base/target/Target.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    protected abstract TYPE simulate(HANDLER handler, EnumFacing side, EXTRA extra);
+
+    public void simulate(EXTRA toSend, SplitInfo<TYPE> splitInfo) {
+        for (Entry<EnumFacing, HANDLER> entry : handlers.entrySet()) {
+            TYPE amountNeeded = simulate(entry.getValue(), entry.getKey(), toSend);
+            if (amountNeeded.compareTo(splitInfo.getAmountPer()) <= 0) {
+                //Add the amount, in case something changed from simulation only mark actual sent amount
+                // in split info
+                acceptAmount(entry.getKey(), splitInfo, amountNeeded);
```
thiakil: inside a `simulate` method, but the `acceptAmount` implementations are for actually sending?

src/main/java/mekanism/common/util/EmitUtils.java
```diff
@@ -0,0 +1,73 @@
...
+     * @param <EXTRA> Any extra information we may need.
+     * @param <TARGET> The emitter target.
+     * @param availableTargets The targets to distribute toSend fairly among.
+     * @param totalTargets The total number of targets. Note: this number is bigger than availableTargets.size if any
+     * targets have more than one acceptor.
+     * @param splitInfo Information containing the split.
+     * @param toSend Any extra information such as gas stack or fluid stack.
+     * @return The amount that actually got sent.
+     */
+    public static <HANDLER, TYPE extends Number & Comparable<TYPE>, EXTRA, TARGET extends Target<HANDLER, TYPE, EXTRA>>
```
thiakil: This method could do with a brief explanation of its logic in the javadoc, since it uses variables that are changed outside this class

pupnewfster: Made it private for now as it is more of a helper generic method that doesn't need to be exposed anyways. And improved the comments inside it for describing how it works.

src/main/java/mekanism/common/base/target/Target.java
pupnewfster: Renamed the method and added JavaDocs of how it works.

src/main/java/mekanism/common/base/EnergyAcceptorWrapper.java
pupnewfster: Done

PR #5424, opened 2019-04-21
Fix a few missing fluid color renders for various guis

pupnewfster:
- Fixes a few spots where the color was not drawn for fluidstacks.
- Deduplicated the gauge code for those spots
- Fixed the Industrial Turbine GUI not getting the proper fluid scale.

### DIFF
--- a/src/main/java/mekanism/client/gui/GuiEmbeddedGaugeTile.java
+++ b/src/main/java/mekanism/client/gui/GuiEmbeddedGaugeTile.java
@@ -1,0 +1,57 @@
+package mekanism.client.gui;
+
+import mekanism.client.render.MekanismRenderer;
+import mekanism.client.render.MekanismRenderer.FluidType;
+import mekanism.common.tile.prefab.TileEntityContainerBlock;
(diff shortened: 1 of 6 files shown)


src/main/java/mekanism/client/gui/GuiEmbeddedGaugeTile.java
```diff
@@ -0,0 +1,57 @@
...
+    protected void displayGauge(int xPos, int yPos, int scale, FluidStack fluid, int side /*0-left, 1-right*/) {
+        if (fluid == null) {
+            return;
+        }
+        int guiWidth = (width - xSize) / 2;
+        int guiHeight = (height - ySize) / 2;
+        int start = 0;
+        int color = fluid.getFluid().getColor(fluid);
+        if (color != -1) {
+            MekanismRenderer.color(color);
```
thiakil: better to avoid `MekanismRenderer.color/resetColor` and use `GlStateManager` instead. (we need to cleanup the direct GL calls, but it wasn't a simple find/replace when I last tried)

PR #5427, opened 2019-04-22
Improve Compat with TOP

pupnewfster:
- Fix capabilities sometimes being disabled even for null side (TOP)
- Made the null handlers for mechanical pipes and pressurized tubes show the network's data so that TOP can show the proper amount of gas/liquid in the network.
- Fix dynamic valve item handler always being disabled
- Display gas color for the TOP gas tank view. (Note: Disabled for hydrogen as it renders weirdly)

### DIFF
--- a/src/main/java/mekanism/common/base/IFluidHandlerWrapper.java
+++ b/src/main/java/mekanism/common/base/IFluidHandlerWrapper.java
@@ -20,12 +20,4 @@ public interface IFluidHandlerWrapper {
     FluidTankInfo[] getTankInfo(EnumFacing from);
 
     FluidTankInfo[] getAllTanks();
-
-    /**
-     * Called when accessing the handler from a null side. Overriding this allows replacing the local tanks with the
-     * networks tanks. Used by TileEntityMechanicalPipe to report the network's contents to TOP
-     */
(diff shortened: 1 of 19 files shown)


src/main/java/mekanism/common/base/IFluidHandlerWrapper.java
```diff
@@ -20,4 +20,12 @@
     FluidTankInfo[] getTankInfo(EnumFacing from);
 
     FluidTankInfo[] getAllTanks();
+
+    /**
+     * Called when accessing the handler from a null side. Overriding this allows replacing the local tanks with the
+     * networks tanks. Used by TileEntityMechanicalPipe to report the network's contents to TOP
+     */
+    default FluidTankInfo[] getAllNetworkTanks() {
```
thiakil: I'm not sure this needs to be a thing, shouldn't the Mechanical pipes be returning the network's buffer when connected anyway?

pupnewfster: Fixed them not exposing the full network's buffer. Originally thought this was intentional is why I did it differently. (Though I am keeping the read only null side handler I added to the gas network)

PR #5429, opened 2019-04-23
Fix Crash when adding a dot or a E...

ToMe25:
...to the Laser Amplifier Delay.


ToMe25:
lol. why does it say there are so many contributions? The most of them should not listed there because they also happened here.

thiakil:
you merged the wrong 1.12 branch, looks like what you merged was what is now 1.12_old

And what is said crash & how does this fix it?

ToMe25:
If you add a dot or a E to the Laser Amplifier Delay field and press Enter it Crashes, and this Fixes it by not calling the Delay Field if the Player presses dot or E.

ToMe25:
so i guess after this i will need to create a new Branch to stop that contributions occurring, right?

thiakil:
basically yeah, then cherry pick your relevant commit. Remember that you can force push it to the same name and it will update the pr.

Is it a new fork or an old fork?

ToMe25:
this is the fork from my first pr.

ToMe25:
1. if i try to make a new pull request with the same name using the Web Interface does that work to update the PR?
2. the original name or the edited?

ToMe25:
so it doesn't works i guess.

PR #5430, opened 2019-04-23
Fix Crash when adding a dot or a E...

ToMe25:
...to the Laser Amplifier Delay.

### DIFF
--- a/src/main/java/mekanism/client/gui/GuiLaserAmplifier.java
+++ b/src/main/java/mekanism/client/gui/GuiLaserAmplifier.java
@@ -134,10 +134,14 @@ public class GuiLaserAmplifier extends GuiMekanismTile<TileEntityLaserAmplifier>
             }
         }
 
-        if (Character.isDigit(c) || c == '.' || c == 'E' || isTextboxKey(c, i)) {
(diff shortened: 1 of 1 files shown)


src/main/java/mekanism/client/gui/GuiLaserAmplifier.java
```diff
@@ -134,11 +134,15 @@ public void keyTyped(char c, int i) throws IOException {
             }
         }
 
-        if (Character.isDigit(c) || c == '.' || c == 'E' || isTextboxKey(c, i)) {
+        if (Character.isDigit(c) || isTextboxKey(c, i)) {
             minField.textboxKeyTyped(c, i);
             maxField.textboxKeyTyped(c, i);
             timerField.textboxKeyTyped(c, i);
         }
+        else if(c == '.' || c == 'E') {
```
thiakil: why not just wrap timerField in an `if (c!='.' &&  c !='E')`?

ToMe25: because i think it looks better like this but i can change that :)

ToMe25: ok i changed that.

PR #5431, opened 2019-04-24
Use security utils for checking access to teleporter and quantum entangloporter

pupnewfster:
- Fixes being unable to let other people view the teleporter/quantum entangloporter (including ops with the bypass config option set)
- Also add the security tab to the corresponding gui's


pupnewfster:
> Looks & tests ok to me (though I haven't checked changing another player's Entangoloporter to a private frequency)
> 
> I am concerned those with existing private frequency settings will suddenly have public gui access?

I don't think there is a great way of dealing with this (especially if someone tries to set it to a private frequency for someone else's block). I think the best thing to do for now is to just put a specific note in the change log saying the accessibility of the teleporter and entangloporter changed and to make sure the security settings match what is intended. (Or to use the security override in the security desk)

Issue #5083, opened 2018-06-13
[Suggestion]: Support for "Mystical Agriculture" / "Mystical Agradditions" ores


PearlCanopus:
I suggest to implement a compatibility for "Mystical Agriculture" / "Mystical Agradditions" ores.
I mine Prosperity- and Inferium-Ore with silk touch. Right now I cannot process these ores with Pulverizer or Enrichment Chamber or Crusher.

I'd like to see that in a future release, as far as these ores are registered well in OreDict.

thiakil:
In the meantime it can be added by crafttweaker

ToMe25:
I am Starting to implement this and a little bit more compat now.

PR #5445, opened 2019-04-29
Add some more Compatibility

ToMe25:
This adds compatibility for Mystical Agriculture(implements #5083),
compatibility for Mods adding their own versions of Vanilla Ores
and Compatibility for some Standard Gems(Ruby, Sapphire, Peridot and Topaz).

### DIFF
--- a/src/main/java/mekanism/common/integration/MekanismHooks.java
+++ b/src/main/java/mekanism/common/integration/MekanismHooks.java
@@ -388,7 +388,6 @@ public final class MekanismHooks {
         OreDictManager.addStandardOredictMetal("Manganese");
         OreDictManager.addStandardOredictMetal("Meutoite");
         OreDictManager.addStandardOredictMetal("Midasium");
-        OreDictManager.addStandardOredictMetal("Mithril");
         OreDictManager.addStandardOredictMetal("Orichalcum");
         OreDictManager.addStandardOredictMetal("Oureclase");
         OreDictManager.addStandardOredictMetal("Prometheum");
--- a/src/main/java/mekanism/common/integration/MekanismHooks.java
+++ b/src/main/java/mekanism/common/integration/MekanismHooks.java
@@ -26,6 +26,7 @@ import mekanism.common.integration.wrenches.Wrenches;
 import mekanism.common.recipe.RecipeHandler;
 import mekanism.common.recipe.inputs.MachineInput;
 import mekanism.common.util.StackUtils;
+import net.minecraft.block.Block;
 import net.minecraft.creativetab.CreativeTabs;
 import net.minecraft.init.Blocks;
 import net.minecraft.init.Items;
@@ -60,6 +61,7 @@ public final class MekanismHooks {
     public static final String TOP_MOD_ID = "theoneprobe";
     public static final String BUILDCRAFT_MOD_ID = "BuildCraft";
     public static final String CYCLIC_MOD_ID = "cyclicmagic";
+    public static final String MYSTICALAGRICULTURE_MOD_ID = "mysticalagriculture";
     public static final String CRAFTTWEAKER_MOD_ID = "crafttweaker";
 
     public boolean IC2Loaded = false;
@@ -70,8 +72,9 @@ public final class MekanismHooks {
     public boolean RFLoaded = false;
     public boolean MetallurgyLoaded = false;
     public boolean CyclicLoaded = false;
-    public boolean CraftTweakerLoaded = false;
     public boolean OCLoaded = false;
+    public boolean MALoaded = false;
+    public boolean CraftTweakerLoaded = false;
 
     public void hookPreInit() {
         if (Loader.isModLoaded(IC2_MOD_ID)) {
@@ -100,6 +103,9 @@ public final class MekanismHooks {
         }
         if (Loader.isModLoaded(METALLURGY_MOD_ID)) {
             MetallurgyLoaded = true;
+        }
+        if (Loader.isModLoaded(MYSTICALAGRICULTURE_MOD_ID)) {
+            MALoaded = true;
         }
         if (Loader.isModLoaded(CRAFTTWEAKER_MOD_ID)) {
             CraftTweakerLoaded = true;
@@ -142,6 +148,9 @@ public final class MekanismHooks {
         if (MetallurgyLoaded) {
             addMetallurgy();
             Mekanism.logger.info("Hooked into Metallurgy successfully.");
+        }
+        if (MALoaded) {
+            registerMysticalAgricultureRecipes();
         }
 
         if (CraftTweakerLoaded) {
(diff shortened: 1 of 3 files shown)


src/main/java/mekanism/common/integration/OreDictManager.java
```diff
@@ -348,4 +393,35 @@ public static void addStandardOredictMetal(String suffix) {
...
+    public static void addStandardOredictGem(String suffix) {
+        NonNullList<ItemStack> gems = OreDictionary.getOres("gem" + suffix);
+        NonNullList<ItemStack> dusts = OreDictionary.getOres("dust" + suffix);
+        NonNullList<ItemStack> ores = OreDictionary.getOres("ore" + suffix);
+        for (ItemStack ore : ores) {
+            RecipeHandler.addEnrichmentChamberRecipe(StackUtils.size(ore, 1), StackUtils.size(gems.get(0), 2));
+        }
+
+        if (dusts.size() > 0) {
+            for (ItemStack ore : gems) {
```
thiakil: please fix your `for` loop variables

src/main/java/mekanism/common/integration/OreDictManager.java
```diff
@@ -148,6 +151,8 @@ public static void init() {
                       StackUtils.size(OreDictionary.getOres("dustNetherQuartz").get(0), 1));
             } catch (Exception ignored) {
             }
+            RecipeHandler.addCombinerRecipe(StackUtils.size(ore, 8), new ItemStack(Blocks.NETHERRACK),
+                  StackUtils.size(OreDictionary.getOres("oreQuartz").get(0), 1));
```
thiakil: `OreDictionary.getOres("oreQuartz").get(0)` should be pulled out of the loop

src/main/java/mekanism/common/integration/OreDictManager.java
```diff
@@ -159,7 +164,46 @@ public static void init() {
         }
 
         for (ItemStack ore : OreDictionary.getOres("oreQuartz")) {
-            RecipeHandler.addEnrichmentChamberRecipe(StackUtils.size(ore, 1), new ItemStack(Items.QUARTZ, 6));
+            RecipeHandler.addEnrichmentChamberRecipe(StackUtils.size(ore, 1),
+                  StackUtils.size(OreDictionary.getOres("gemQuartz").get(0), 6));
```
thiakil: please pull all `OreDictionary.getOres` calls out of the loops in this block

thiakil: also please remove all the try/catch blocks where the catch is empty, they're bad legacy code

src/main/java/mekanism/common/integration/MekanismHooks.java
```diff
@@ -393,4 +402,38 @@ private void addMetallurgy() {
         OreDictManager.addStandardOredictMetal("Vyroxeres");
         OreDictManager.addStandardOredictMetal("Zinc");
     }
+
+    private void registerMAEnrichmentRecipe(int itemMeta, int oreType) {
```
thiakil: Not a fan of using magic numbers like this, and the double ternary is quite bad.

Is there absolutely no way to get this info from an api or anything? If not a couple of enums would do nicely that have prefixes as a member field & the input/output meta on another enum's fields.

Same goes for the Combiner ones - methods look like they could be combined anyway?

src/main/java/mekanism/common/integration/OreDictManager.java
```diff
@@ -348,4 +393,35 @@ public static void addStandardOredictMetal(String suffix) {
...
+    
+    public static void addStandardOredictGem(String suffix) {
+        NonNullList<ItemStack> gems = OreDictionary.getOres("gem" + suffix);
+        NonNullList<ItemStack> dusts = OreDictionary.getOres("dust" + suffix);
+        NonNullList<ItemStack> ores = OreDictionary.getOres("ore" + suffix);
+        for (ItemStack ore : ores) {
+            RecipeHandler.addEnrichmentChamberRecipe(StackUtils.size(ore, 1), StackUtils.size(gems.get(0), 2));
+        }
+
+        if (dusts.size() > 0) {
```
thiakil: I think this block should be inside the check for `ores.size() > 0`?

ToMe25: That's right this was just a problem caused by the copy pasting :)

src/main/java/mekanism/common/integration/OreDictManager.java
ToMe25: @thiakil what is wrong with this for loop?
do you mean it shouldn't be name ore? but the entire class does it like that.

src/main/java/mekanism/common/integration/OreDictManager.java
ToMe25: to your first comment: yes. i just did that bad.
to your second one: ok i will remove them, but i didn't made them all.

src/main/java/mekanism/common/integration/OreDictManager.java
thiakil: Indeed, something looping on `gems` should call it's variable `gem`

I know there's lots of legacy cruft, but new code should do better - fixing the old ones is optional

ToMe25: ok i will change that after i finished some testing with andother mod.

src/main/java/mekanism/common/integration/OreDictManager.java
ToMe25: it got it's own if now.

ToMe25:
@thiakil i think i implemented it all.

src/main/java/mekanism/common/integration/OreDictManager.java
```diff
@@ -336,15 +402,48 @@ public static void addStandardOredictMetal(String suffix) {
...
+                RecipeHandler.addEnrichmentChamberRecipe(StackUtils.size(dusts.get(0), 1), StackUtils.size(gem, 1));
+                RecipeHandler.addCrusherRecipe(StackUtils.size(gem, 1), StackUtils.size(dusts.get(0), 1));
             }
         }
 
         if (ores.size() > 0) {
-            for (ItemStack ore : dusts) {
-                RecipeHandler.addCombinerRecipe(StackUtils.size(ore, 8), new ItemStack(Blocks.COBBLESTONE),
-                      StackUtils.size(ores.get(0), 1));
+            if (dusts.size() > 0) {
```
pupnewfster: I am not necessarily against this, but I am unsure if it is a good idea to be changing the combiner recipes from using the gem directly to using the dust. For one thing it will break people's CraftTweaker scripts if they are trying to remove the old recipe and now that recipe no longer exists (which would fail gracefully with a warning), except the dust to ore recipe would not get removed and they may not understand why the script no longer works. Not to mention I am unsure how to feel about the recipe changing based on the mods installed, (such as if there is emerald dust or not then that may be a bit annoying for pack makers, however that is less of a concern than breaking existing things).

It depends on what @thiakil thinks, but I think the best solution here may be to leave a TODO saying to switch to dust where possible for gems in Mekanism v10. Given that is a large enough version change that we can make breaking changes.

ToMe25: @pupnewfster "I am unsure if it is a good idea to be changing the combiner recipes from using the gem directly to using the dust" this is only a change for the Materials without a dust added by Mekanism(diamond already uses the dust) and i don't like to use a different version of that Method for Diamond, so that is the detection mainly for handling diamonds correctly.(and i was thinking it's a good idea to use dust for everything if they are allready used for diamonds)

ToMe25: By the Way the Only Recipe that is different than in the last version is that one for Emerald Ore.
All other Recipes that Method adds are new.

pupnewfster: Ah I could have sworn it was 3 diamonds in the combiner to make the ore not three dust. Seems I was wrong then. Looks fine overall then

ToMe25: Should i add the remaining BoP gems to the list of supported gems?

pupnewfster: Might as well

ToMe25: I Implemented this.

thiakil:
Thanks for cleaning up the other loop variables @ToMe25 :)

Issue #4817, opened 2017-11-19
Logistical sorter round robin limit


BaneOfSerenity:
MC: 1.12.2
Forge: 14.23.0.2515
Mek: 9.4.1

I noticed when trying to RR with the LS it tries to RR a stack at a time. I would be nice to be able to set this to RR 1 item at a time in order to correctly fill multiple machines at a time when you only have enough items to partially fill multiple machines.

Zajozor:
I could look into this, if it is still requested. I looked through the code and it seems like something I should be able to grasp. Should the issue be reopened/shall I create a new one, or just submit a PR?

thiakil:
Just a PR is fine :)

Issue #4813, opened 2017-11-15
[Suggestion] [1.12.2] Add Muffling Upgrade to Logistical Sorter


Skyler-Altol:
The title basically says it all. Would it be possible to add upgrade functionality to the logistical sorter, so that we could use muffling upgrades in it? The consistent clicking while its working may be nice for some, but it would be nice to be able to have a choice in the matter :)

Cheers.

GGG-KILLER:
If possible could the logistical sorter also be made to accept speed upgrades?

My mob grinder is being slowed down by the logistical sorter not taking items out of the chest fast enough and the items that aren't picked up are sometimes being deleted by the server because of the delay.

Issue #5435, opened 2019-04-26
Duplication of items using a logical sorter


bigenergy:
#### Issue description:
Duplication of items using a logical sorter and cells AE2

#### Steps to reproduce:
1. Take any two EQUAL cells of the same size (this is important)
2. Put in them the items you want to duplicate
3. Put two chests and between them a logical conveyor
4. Configure the cell filter
5. Put two cells in the chest

#### Version (make sure you are on the latest version before reporting):
**Forge:** 2814
**Mekanism:** 9.7.2.373

#### If a (crash)log is relevant for this issue, link it here: (_It's almost always relevant_)

Video (russian) https://www.youtube.com/watch?v=mEowd8Dpru8

thiakil:
Cant reproduce with #5451

Issue #5211, opened 2018-12-02
Logistical Sorter Ordering


AnibalDuarte:
Forge: 14.23.5.2768
Mekanism: 1.12.2-9.4.12.349
Other relevant version: Minecraft 1.12.2

Sometimes I want to create an rule and put it in the begin (or the end) of the list, to override (or not be overrided by) another rules. But in sorters with a lot of filters it's a hard task.

Would be cool if the filter list had a button to "send to top/bottom" of the list.

PR #5446, opened 2019-05-01
Logistical sorter - single item & muffling upgrade

Zajozor:
Fixes https://github.com/mekanism/Mekanism/issues/4817

I added a button for the logistical sorter, that allows you to toggle between sending single items and sending whole stacks(the way it works now).
Other than that, I added some entries to the lang file and updated the tooltips for buttons to say a little more meaningful messages.
The GUI png file was updated, and I removed the xcf file, which I believe was kept in the repository only by mistake.

What I did not test:
- whether the method `setSingleItem` for CC works properly (only wrote the code analogously to other buttons)
- Obscure situations when the LS gets un-chunkloaded, or the target inventories do, etc. but I guess the change should not introduce new bugs.

### Screenshots
Looks like this:
image
Credits for the button icon go to @Sandyna :)

Tested with a setup like this:
image
image

Sending items with single item mode on works pretty well.
lsorter_in_action

The new tooltips:
image
image
image


Zajozor:
Also wanted to ask:

1) The code in `TileEntityLogisticalSorter` seems a little messy due to duplicate code between methods (`getNetworkedData` `getGenericPacket` `getConfigurationData``readFromNBT` etc..). If you want, I can refactor it to what I believe would look and read a little nicer.

2) I was also thinking about looking into #5211 , #4813 (this one annoyed me personally too :D ) , #4966 , maybe #4756 , or any other problems regarding logistical sorter if you know of any. Do you feel like all of these would be cool to implement? Any order of priority, or shall I choose? Or perhaps other plans/problems with the LogSorter I could look into?

pupnewfster:
At least from the pictures it looks good (I will look into the code and testing it a bit later when I have some more time). Technically it is already possible to do this sort of by setting the stack size of each filter to 1, however I do agree that this is a nice QoL thing to have and makes it even more powerful and easier to use.

In terms of your questions, it is up to you if you make the code more readable, at some point we do plan on rewriting the network/sync code more in general so that it isn't all done through one master packet which should make it easier to not accidentally put things in the wrong order causing the data to be corrupted. Though having the code be easier to read and manage is never a bad thing.

In terms of any priority of what order you want to make PRs that is up to you. It would be nice to have #5435 fixed if you are able to reproduce it easily (I tried some and was unable to), but likely that just has some issue to do with NBT/it grouping them into a single request packet.

Also in terms of if you do look at #4756 I would recommend using JEI's GhostSlot stuff instead of the IRecipeTransferHandler (I think those two are different systems). Basically would do it the same way Ender IO has it, where you can drag items from JEI into the GUI for the filter. Though honestly that may be better to hold off on until after the internals of the filter system get cleaned up as there is a lot of duplicate code between Logistical Sorter and the Digital Miner, and the various types of filters/subfilters they have.

src/main/java/mekanism/client/gui/GuiLogisticalSorter.java
```diff
@@ -511,11 +522,15 @@ protected void drawGuiContainerForegroundLayer(int mouseX, int mouseY) {
         }
 
         if (xAxis >= 12 && xAxis <= 26 && yAxis >= 110 && yAxis <= 124) {
-            drawHoveringText(LangUtils.localize("gui.autoEject"), xAxis, yAxis);
+            drawHoveringText(LangUtils.localize("gui.logisticalSorter.autoEject.tooltip"), xAxis, yAxis);
```
pupnewfster: I think the tooltips (this one and the other two added) should wrap the text rather than just straight up be drawn. This is because some of the descriptions are somewhat long so if you do not have Minecraft in full screen it can run off the edge slightly. (Should be manageable automatically similar to how our item tooltips have their description.)

Zajozor: Good point, will look into it 👍

thiakil: Please also ensure any new language keys are prefixed with `mekanism.`

thiakil:
The logistical sorter seems to already have a size based mode, but it's just on the ItemStack Filter 🤔

Zajozor:
The tooltips are wrapped, and I prefixed all logistical sorter entries in the lang files as requested.

Good that you mention that this could be achieved using `ItemStack` filters, I actually broke that functionality by calling `getItemAmountFromInventory` instead of `getStackFromInventory` (which is overriden there). I will go through all the filter options thoroughly again, to make sure I did not forget stuff like this.

I feel like the purpose of this button is two things:
- You want to always take one item, while you may have many filters that you do not want to configure manually.
- You do not want to configure any filters -> want to take out any item, but send it one by one (usually with round robin).

Is that enough to have that functionality added, or you rather do not feel like its worth it?

In the next days -> will fix my changes breaking setting item count with item filters, will try to replicate #5435 , and #4813 seems like something I can take on as well.

thiakil:
> Is that enough to have that functionality added, or you rather do not feel like its worth it?

Yeah it's fine, if this option is enabled it should override the individual filter settings I think. I'd recommend adding `(overrides filters)` to the tooltip

Zajozor:
Sorry for the long time, here are the requested changes (#5435 was fixed in the meantime).

- fixed the code, to work as requested = single item mode overrides ItemStack sizeMode
- did some minor code cleanups
- gui information
When single item mode is turned off, looks like normal:
image

When you turn on single item mode and also have sizeMode turned on:
image
With a tooltip (only when singleItem is turned on):
image

Also, the singleItem button tooltip tells you it overrides sizeMode:
image


- rebased the branch above your 1.12

Will look into muffling upgrade support as next feature.

Zajozor:
Also fixes #4813 

Changes to add muffling upgrade to logistical sorter:
- Add Upgrades tab to the GUI and implement `IUpgradeTile`
- Turn logistical sorter into `TileEntityEffectsBlock`, so that it plays the sound through `SoundHandler` (which respects muffling)
- Add a constructor to `TileEntityEffectsBlock` that allows you to set the `rapidChangeThreshold` (otherwise, the logistical sorter would keep playing the clicking sound for 10 ticks like other machines, but here it results in "more" ticks than we want)
- Add guiID 59, so that when you press the back button from the UpgradesManagement GUI you can get back

Images:
image
image

thiakil:
For future, please try to keep separate features in separate PRs (where possible)

src/main/java/mekanism/client/gui/GuiLogisticalSorter.java
```diff
@@ -377,13 +382,16 @@ protected void drawGuiContainerForegroundLayer(int mouseX, int mouseY) {
         fontRenderer.drawString(LangUtils.localize("gui.filters") + ":", 11, 19, 0x00CD00);
         fontRenderer.drawString("T: " + tileEntity.filters.size(), 11, 28, 0x00CD00);
 
-        fontRenderer.drawString("RR:", 12, 74, 0x00CD00);
+        fontRenderer.drawString(LangUtils.localize("mekanism.gui.logisticalSorter.singleItem") + ":", 12, 48, 0x00CD00);
+        fontRenderer.drawString(LangUtils.localize("gui." + (tileEntity.singleItem ? "on" : "off")), 27, 60, 0x00CD00);
```
thiakil: Would you kindly change these on/off localisations to use `mekanism.common.util.LangUtils#transOnOff`?

Will save some trouble down the track when I fix the lang prefix issues

src/main/java/mekanism/client/gui/filter/GuiTItemStackFilter.java
```diff
@@ -128,7 +128,11 @@ protected void drawGuiContainerForegroundLayer(int mouseX, int mouseY) {
         fontRenderer.drawString(LangUtils.localize("gui.itemFilter.details") + ":", 35, 32, 0x00CD00);
         fontRenderer.drawString(LangUtils.localize("gui.itemFilter.min") + ":", 128, 20, 0x404040);
         fontRenderer.drawString(LangUtils.localize("gui.itemFilter.max") + ":", 128, 32, 0x404040);
-        fontRenderer.drawString(LangUtils.localize("gui." + (filter.sizeMode ? "on" : "off")), 141, 46, 0x404040);
+        String sizeModeString = LangUtils.localize("gui." + (filter.sizeMode ? "on" : "off"));
+        if (tileEntity.singleItem && filter.sizeMode)
```
thiakil: Please always use braces.

If you're using IntelliJ IDEA, this can be imported (save it to the project, no overwrite your defaults): https://github.com/mekanism/Mekanism/blob/1.12/docs/GoogleStyle4Indents.xml

src/main/java/mekanism/client/gui/filter/GuiTItemStackFilter.java
```diff
@@ -151,7 +155,10 @@ protected void drawGuiContainerForegroundLayer(int mouseX, int mouseY) {
         int xAxis = (mouseX - (width - xSize) / 2);
         int yAxis = (mouseY - (height - ySize) / 2);
         if (xAxis >= 128 && xAxis <= 139 && yAxis >= 44 && yAxis <= 55) {
-            drawHoveringText(LangUtils.localize("gui.sizeMode"), xAxis, yAxis);
+            String sizeModeTooltip = LangUtils.localize("gui.sizeMode");
+            if (tileEntity.singleItem && filter.sizeMode)
```
thiakil: Braces

src/main/java/mekanism/common/content/transporter/TItemStackFilter.java
```diff
@@ -20,18 +20,10 @@
...
-            if (max == 0 || itemStack.getCount() < min) {
-                return false;
-            }
-        }
-
-        return (itemType.getHasSubtypes() ? itemType.isItemEqual(itemStack)
-              : itemType.getItem() == itemStack.getItem());
+        return super.canFilter(itemStack, strict) &&
+                !(strict && sizeMode && (max == 0 || itemStack.getCount() < min)) &&
+                (itemType.getHasSubtypes() ?
```
thiakil: This subtype check chain (from `itemType.getHasSubtypes() `) should be replaced with `net.minecraftforge.items.ItemHandlerHelper#canItemStacksStackRelaxed`

src/main/java/mekanism/common/content/transporter/TransporterFilter.java
```diff
@@ -60,12 +60,18 @@ public static TransporterFilter readFromPacket(ByteBuf dataStream) {
...
+    }
 
     public abstract Finder getFinder();
 
     public InvStack getStackFromInventory(StackSearcher searcher) {
-        return searcher.takeTopStack(getFinder());
+        return getItemAmountFromInventory(searcher, 64);
+    }
+
+    public InvStack getItemAmountFromInventory(StackSearcher searcher, int amount) {
```
thiakil: this should have the same method name since one calls the other

src/main/java/mekanism/common/tile/TileEntityLogisticalSorter.java
```diff
@@ -95,22 +100,25 @@ public void onUpdate() {
                 outer:
                 for (TransporterFilter filter : filters) {
                     for (StackSearcher search = new StackSearcher(back, facing.getOpposite()); search.getSlotCount() >= 0; ) {
-                        InvStack invStack = filter.getStackFromInventory(search);
+                        InvStack invStack = singleItem ? filter.getItemAmountFromInventory(search, 1)
```
thiakil: I think this would be clearer to put the ternary as the number param `singleItem ? 1 : 64`

thiakil:
Functionality checks out

src/main/java/mekanism/common/tile/TileEntityLogisticalSorter.java
```diff
@@ -46,32 +46,37 @@
...
     public HashList<TransporterFilter> filters = new HashList<>();
     public RedstoneControl controlType = RedstoneControl.DISABLED;
     public EnumColor color;
     public boolean autoEject;
     public boolean roundRobin;
+    public boolean singleItem;
     public int rrIndex = 0;
     public int delayTicks;
-    public boolean isActive;
     public boolean clientActive;
```
pupnewfster: I haven't tested this stuff yet, but my suspicion is that some stuff between clientActive/isActive is still duplicated now/not needed due to the changing superclass.

src/main/java/mekanism/client/gui/filter/GuiTItemStackFilter.java
Zajozor: I have it imported, yet it showed me nothing wrong with missing braces, could you maybe point me in some direction where i can look to fix this?
image

src/main/java/mekanism/common/tile/TileEntityLogisticalSorter.java
Zajozor: well, this was used because there were two methods - `getStackFromInventory` and getItemAmountFromInventory, only one was overriden in `TItemStackFilter`, so I wanted to call it only in one of the cases.

Will rewrite this to require only one `getStackFromInventory` method and make it cleaner.

src/main/java/mekanism/client/gui/filter/GuiTItemStackFilter.java
pupnewfster: On the left under `Code Style` click on the sub tab for `Java`. It should have things like `Force braces` set to `always` under the `Wrapping and Braces` tab for if statements. My guess is that for some reason the scheme didn't load for the java specific scheme.

Zajozor: strange, I do have it turned on. So probably some cache problem at my side, will try clearing it.
image
thanks anyway

pupnewfster: Actually thinking about it, probably has to do with it not automatically running. (As in if you didn't put the braces out of habit anyways it wouldn't add them until you run the formatter `CTRL+ALT+L` for current file. For now if you want you can just add the braces manually though given after we get the open PRs merged I am going to do a quick pass making sure things are easier to read across the board since we are changing some of the formatting styling in #5463 .

src/main/java/mekanism/common/tile/TileEntityLogisticalSorter.java
Zajozor: Probably right.

Not sure how to test this, as I understand only the surface of this networking stuff, but I removed the `isActive` and `clientActive` logic from `TileEntityLogisticalSorter` and all seems good.

Zajozor:
So I did incorporate all the requested changes:
- Using `Langutils#transOnOff`
- Code style changes (braces, simpler method structure in `getStackFromInventory`
- I removed the `isActive` and `clientActive` logic from `TileEntityLogisticalSorter` which seemed duplicate to me as well, but I am not totally certain about this.

Please let me know if there is anything else I should change/improve.

src/main/java/mekanism/common/tile/TileEntityLogisticalSorter.java
pupnewfster: Added back one small part you removed that made the sorter update so the light on the side would turn on. Given it turns out the `TileEntityEffectsBlock` doesn't call it quite often enough in specific cases at least with the sorter which should have it happen more often.

PR #5451, opened 2019-05-03
Itemstack equality

### LINKED ISSUE #5449: Digital Miner dupe bug
**Issue description:**
You can duplicate items using a Digital Miner

**Steps to reproduce:**
1. Place a Digital Miner in the ground
2. Put items inside a stackable block that holds items in after being broken(Like RFTools Crafter)
3. Put more "RFTools Crafters" in the ground
5. Set a filter in the Digital Miner to mine the Crafters(Fuzzy Mode:ON)
6. Turn on the digital miner
7. All the collected crafters by the miner will be the same as the first (See the video for more info)

**Version (make sure you are on the latest version before reporting):**
**Forge:** 1.12.2 - 14.23.5.2836
**Mekanism:** Mekanism-1.12.2-9.7.3.374

**If a (crash)log is relevant for this issue, link it here:**

https://www.youtube.com/watch?v=IKg_pf44QBo

thiakil:
NB: haven't tested this yet.

A few item stack equality check changes to hopefully fix #5449 & #5435

### DIFF
--- a/src/main/java/mekanism/common/tile/TileEntityFactory.java
+++ b/src/main/java/mekanism/common/tile/TileEntityFactory.java
@@ -67,7 +67,6 @@ import net.minecraft.util.NonNullList;
 import net.minecraft.util.text.translation.I18n;
 import net.minecraftforge.common.capabilities.Capability;
 import net.minecraftforge.fml.common.FMLCommonHandler;
-import net.minecraftforge.items.ItemHandlerHelper;
 
 public class TileEntityFactory extends TileEntityMachine implements IComputerIntegration, ISideConfiguration,
       IGasHandler, ISpecialConfigData, ITierUpgradeable, ISustainedData {
@@ -406,11 +405,8 @@ public class TileEntityFactory extends TileEntityMachine implements IComputerInt
                 for (int j = i + 1; j < inputSlots.length; j++) {
                     int checkSlotID = inputSlots[j];
                     ItemStack checkStack = inventory.get(checkSlotID);
-                    if (Math.abs(count - checkStack.getCount()) < 2) {
-                        continue;
-                    }
-                    if (!stack.isEmpty() && !checkStack.isEmpty() && !ItemHandlerHelper
-                          .canItemStacksStack(stack, checkStack)) {
+                    if (Math.abs(count - checkStack.getCount()) < 2 ||
+                          !InventoryUtils.areItemsStackable(stack, checkStack)) {
                         continue;
                     }
                     //Output/Input will not match
--- a/src/main/java/mekanism/common/util/InventoryUtils.java
+++ b/src/main/java/mekanism/common/util/InventoryUtils.java
@@ -100,7 +100,7 @@ public final class InventoryUtils {
         for (int i = inventory.getSlots() - 1; i >= 0; i--) {
             ItemStack stack = inventory.extractItem(i, max, true);
 
-            if (!stack.isEmpty() && StackUtils.equalsWildcard(stack, type)) {
+            if (!stack.isEmpty() && StackUtils.equalsWildcardWithNBT(stack, type)) {
                 int current = ret.getStack().getCount();
 
                 if (current + stack.getCount() <= max) {
--- a/src/main/java/mekanism/common/util/InventoryUtils.java
+++ b/src/main/java/mekanism/common/util/InventoryUtils.java
@@ -1,6 +1,7 @@
 package mekanism.common.util;
 
 import java.util.Map;
+import net.minecraftforge.items.ItemHandlerHelper;
 import org.apache.commons.lang3.tuple.Pair;
 import mekanism.api.EnumColor;
 import mekanism.common.Mekanism;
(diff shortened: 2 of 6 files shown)


src/main/java/mekanism/common/util/InventoryUtils.java
```diff
@@ -70,12 +71,21 @@ public static TransitResponse putStackInInventory(TileEntity tile, TransitReques
...
 
+    /**
+     * Like {@link ItemHandlerHelper#canItemStacksStack(ItemStack, ItemStack)} but empty stacks mean equal (either param).
+     * Thiakil: not sure why.
+     *
+     * @param toInsert stack a
+     * @param inSlot stack b
+     * @return true if they are compatible
+     */
     public static boolean areItemsStackable(ItemStack toInsert, ItemStack inSlot) {
```
pupnewfster: This can (or more accurately `!areItemsStackable()`) probably now be used for https://github.com/mekanism/Mekanism/blob/1.12/src/main/java/mekanism/common/tile/TileEntityFactory.java#L412

pupnewfster:
I also haven't tested it quite yet, but I cleaned up a few checks in the robit and switched the spot I mentioned in the review over to the method. From doing a search for `isItemEqual` I wonder if we should switch over the Bin to using the stackable check.

src/main/java/mekanism/common/tile/TileEntityFactory.java
```diff
@@ -406,11 +405,8 @@ public void sortInventory() {
                 for (int j = i + 1; j < inputSlots.length; j++) {
                     int checkSlotID = inputSlots[j];
                     ItemStack checkStack = inventory.get(checkSlotID);
-                    if (Math.abs(count - checkStack.getCount()) < 2) {
-                        continue;
-                    }
-                    if (!stack.isEmpty() && !checkStack.isEmpty() && !ItemHandlerHelper
-                          .canItemStacksStack(stack, checkStack)) {
+                    if (Math.abs(count - checkStack.getCount()) < 2 ||
```
thiakil: any idea what the `Math.min` is for here? @pupnewfster

pupnewfster: If I remember correctly it is to make it not run the code if the two slots are only one item apart in amount because it can't make them even as it does not divide evenly.

thiakil:
Factories still good (tested with auto sorting with an rftools crafter filled and a few unfilled)

src/main/java/mekanism/common/util/StackUtils.java
```diff
@@ -11,8 +11,8 @@
 
     //ignores count
     public static boolean equalsWildcard(ItemStack wild, ItemStack check) {
-        if (wild.isEmpty() || check.isEmpty()) {
-            return check == wild;
+        if (wild.isEmpty() && check.isEmpty()) {
```
pupnewfster: I am not quite sure this is a correct change (or at least fully correct), as if wild is empty we want it to return false if check is not empty, however I think the below checks will now happen. I do agree the previous `check == wild` is not the proper way to check to make sure they are both empty, however I think the new way misses a false branch.

src/main/java/mekanism/common/util/InventoryUtils.java
```diff
@@ -70,12 +71,21 @@ public static TransitResponse putStackInInventory(TileEntity tile, TransitReques
         return TransitResponse.EMPTY;
     }
 
+    /**
+     * Like {@link ItemHandlerHelper#canItemStacksStack(ItemStack, ItemStack)} but empty stacks mean equal (either param).
+     * Thiakil: not sure why.
```
pupnewfster: It is checking if two item stacks can stack with each other. From what I can gather it is more of a utils method for purposes of if a stack is empty that means we can put our stack in that slot as well, not just if they have the same item type.

thiakil: imo it shouldnt hit this method if the stack its going into  is empty

src/main/java/mekanism/common/util/StackUtils.java
thiakil: it probably would have always returned false here anyway. If only one is empty the checks below will fail, because its not equal anyway

src/main/java/mekanism/common/util/InventoryUtils.java
pupnewfster: I can see that, though I also can see how this is a nice wrapper so that you don't always have to include extra checks before hand (especially when if caring about them being empty you can directly use the `ItemHandlerHelper#canItemStacksStack` method. Was more just trying to give some clarity on why I think it is probably like this.

src/main/java/mekanism/common/util/StackUtils.java
pupnewfster: I agree that the check was probably returning false here when it shouldn't have been, but doesn't `isEmpty()` also return true if the stack size is 0, so technically the other data could still match if something is giving us a malformed ItemStack?

thiakil:
```
public Item getItem()
    {
        return this.isEmpty || this.delegate == null ? Items.AIR : this.delegate.get();
    }
```
nope

pupnewfster: Forgot to look at things if they return different values if it is empty.

Issue #5453, opened 2019-05-03
Digital Miner Shulker Box Duplication Bug


Focamacho:
**Issue description:**
You can duplicate shulker boxes using a Digital Miner(It duplicates only the Shulker Box, not the items inside)

**Steps to reproduce:**
1. Place a Digital Miner in the ground
2. Place a Shulker Box in the ground
3. Put any item inside the Shulker Box
5. Set a filter in the Digital Miner to mine the Shulker Box(Fuzzy Mode:ON)
6. Turn on the digital miner
7. Digital Miner will collect a Shulker Box, and at the same time, one shulker box is dropped on the ground. (See the video for more info)

**Version (make sure you are on the latest version before reporting):**
**Forge:** 1.12.2 - 14.23.5.2836
**Mekanism:** Mekanism-1.12.2-9.7.3.374

**If a (crash)log is relevant for this issue, link it here:**

https://www.youtube.com/watch?v=ZESNiDCf_XA

pupnewfster:
Started looking into this slightly, and a few different things seem to be going on.

1. The reason there is no dupe of the items inside seems to be because shulker boxes handle that and drop themselves on the break event. (We need to somehow figure this out and have it so the items get stored. This will cause the items inside to also get duped, but that should already be happening and is a bug that it is not.)
2. The reason the box itself is duping is that given it is manually dropping itself when we tell it that it was broken, a second one gets created by setting it to air.

I think what is likely going to need to be done is to have special handling for the shulker box where we duplicate the code so that we get the proper drop NBT, and then we clear the shulker box so that the one in world does not get dropped. The question I will have to look into however, is there some better way to do it so that we have better support if some modded block behaves the same way.

Focamacho:
I tested some other mods:

**Vertical Digger(Actually Additions)**: It is the same as digital miner, takes an empty box, and drops one with items.
**Auto-Breaker(Actually Additions)**: Same thing.
**Block Breaker(Industrial Foregoing)**:  Drop all items on the floor and pick up an empty box (https://imgur.com/a/ikfpqr4)
**Mechanical Miner(Extra Utilities 2)**: Works exactly how it should work, take the box with the items inside and do not drop anything.
**Builder + Quarry Card(RFTools)**: Works the same as mechanical miner, without any problem.

Shadows-of-Fire:
The only way to make things like this respond properly to a block break is to do it like AE does for annihilation planes: break the block then capture the dropped items in the blockspace.

[...]

PR #5456, opened 2019-05-03
Miner cleanup

pupnewfster:
- Cleans up calls to world.playEvent, making it more clear which event is being called
- Made `MinerUtils` easier to read by cleaning up some of the logic so that it has less else branches calling `fallbackGetSilkTouch` this allowed me to also inline `fallbackGetSilkTouch` as it only was being called in one spot.

I still need to look more into #5453 and https://github.com/mekanism/Mekanism/issues/5453#issuecomment-489216976 to see if that is how we should do it or if there is a better way.

### DIFF
--- a/src/main/java/mekanism/common/util/MinerUtils.java
+++ b/src/main/java/mekanism/common/util/MinerUtils.java
@@ -46,29 +46,25 @@ public final class MinerUtils {
 
         if (silk && (block.canSilkHarvest(world, coord.getPos(), state,
               Mekanism.proxy.getDummyPlayer((WorldServer) world).get()) || specialSilkIDs.contains(block))) {
-            List<ItemStack> ret = new ArrayList<>();
+            Object it = null;
             if (getSilkTouchDrop != null) {
                 try {
-                    Object it = getSilkTouchDrop.invoke(block, state);
-                    if (it instanceof ItemStack) {
-                        //Should always be an ItemStack
-                        ItemStack silkDrop = (ItemStack) it;
-                        if (!silkDrop.isEmpty()) {
-                            ret.add(silkDrop);
-                        } else {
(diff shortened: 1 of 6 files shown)


src/main/java/mekanism/common/util/MinerUtils.java
```diff
@@ -28,67 +31,56 @@
...
-                net.minecraftforge.event.ForgeEventFactory
-                      .fireBlockHarvesting(ret, world, obj.getPos(), state, 0, 1.0f, true, null);
+                ForgeEventFactory.fireBlockHarvesting(ret, world, coord.getPos(), state, 0, 1.0F, true, null);
                 return ret;
             }
         } else {
-            @SuppressWarnings("deprecation")
-            List<ItemStack> blockDrops = block.getDrops(world, obj.getPos(), state, 0);
+            NonNullList<ItemStack> blockDrops = NonNullList.create();
+            block.getDrops(blockDrops, world, coord.getPos(), state, 0);
```
thiakil: This needs to use the deprecated one, as if anyone overrides the old method it won't get caught.

Override = new one, Call = old one

thiakil:
Added a specific workaround for shulker boxes (other mods should be told to fix their code if they do the same thing).

Also added a method for getting placement state - only tested with dirt, but should work a little better for any modded blocks who do special state placement stuff there (also helps to remove `getStateFromMeta` calls which I will probably pull into my cleanup branch at a later date)

pupnewfster:
Seems good to me, the one thing I am unsure about is with the `getStateForPlacement` is I think it potentially should be "placing" against the top as birch logs (and probably logs in general) get placed on their side, and it would be more logical to me that they get placed upwards by default.

PR #5460, opened 2019-05-06
Change unused map (values) to use a set

thiakil:
`mekanism.api.transmitters.DynamicNetwork#possibleAcceptors` no longer seems to have its values used, so given it stores TileEntities, best to be rid of it to avoid being the cause of a memory leak.

Technically is a breaking change to the api hovever.

### DIFF
--- a/src/main/java/mekanism/api/MekanismAPI.java
+++ b/src/main/java/mekanism/api/MekanismAPI.java
@@ -18,7 +18,7 @@ public class MekanismAPI {
     /**
      * The version of the api classes - may not always match the mod's version
      */
-    public static final String API_VERSION = "9.7.3";
+    public static final String API_VERSION = "9.7.5";
 
     /**
      * Mekanism debug mode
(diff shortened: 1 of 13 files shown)


thiakil:
> we need to bump the API version

doh! of course

Yeah I was considering making things properly private, but would be even more of a breaking change - though of course that's if anyone actually uses this class directly or in a subclass

pupnewfster:
> Yeah I was considering making things properly private, but would be even more of a breaking change - though of course that's if anyone actually uses this class directly or in a subclass

I think it is probably worth it because if anyone actually does use it the chance is they still implement the getAcceptors that we deprecated in a previous API update, so I think it is better to make the changes in one breaking API change rather than in multiple. Though I do somewhat doubt if anyone uses this part of the API anyways.

thiakil:
Alrighty, I'm good with that :)

It doesn't affect your Gas Conduits at all?

pupnewfster:
> It doesn't affect your Gas Conduits at all?

Those use Enderio's conduit system. For mek all they really rely on is things properly implementing IGasHandler.

thiakil:
Made the API changes and tidied some things up a little

Still need to just boot the game up and make sure nothing weird happened

PR #5462, opened 2019-05-07
Induction matrix optimization

### LINKED ISSUE #3956: Max size Induction Matrix (18x18x18) Causes Server Lag
This isn't the first Mekanism multiblock to cause server lag. We've previously taken down all the salination towers for a similar reason, but the extent that the Induction Matrix causes lag is far, far greater.

`TPS: 14.1
                           Single Entity | Time/Tick |     %
TileEntityInductionPort {2792,59,5431}:0 |     2.036 | 4.927
        TileCustomCable {2793,76,5446}:0 |     1.010 | 2.445
        TileCustomCable {2793,76,5433}:0 |     0.996 | 2.411
TileEntityInductionPort {2808,59,5432}:0 |     0.976 | 2.361
TileEntityInductionPort {2807,74,5448}:0 |     0.975 | 2.359
      Chunk | Time/Tick |      %
0: 174, 339 |     6.168 | 15.034
0: 174, 340 |     5.074 | 12.367
0: 175, 339 |     4.070 |  9.920
0: 175, 340 |     4.044 |  9.856
    0: 0, 9 |     0.805 |  1.961
          All Entities of Type | Time/Tick |      %
       TileEntityInductionPort |    16.361 | 39.878
               TileCustomCable |     2.092 |  5.098
             TileConduitBundle |     1.654 |  4.032
TileMultipartContainer$Ticking |     1.321 |  3.220
           monster.EntityWitch |     1.001 |  2.439
  Average Entity of Type | Time/tick | Calls
 TileEntityInductionPort |     0.824 |  2800
TileEntityWeatherStation |     0.116 |    88
     tile.TileController |     0.088 |   987
         EntityButterfly |     0.070 |   141
     TileExternalStorage |     0.067 |   705`

The difference between having that thing loaded or not is night and day. Since the lag seems to be concentrated at the induction ports, look to the code that adds energy and see what an 18x18x18 matrix filled with Ultimate cells does. I think you'll find a serious inefficiency there.

This is with the latest—as of today—version. The Avant 3 pack got it's 1.16 update JUST to make sure this hadn't been fixed.

pupnewfster:
- Cache the total amount of energy stored. This is a major performance booster for output ports, as `getEnergy()` used to be an expensive call having to check each cell and sum up their total energy. Output ports would call this method twice, once to see how much power there is, and a second time after emitting power to see how much got emitted.
- Added two methods addEnergy/removeEnergy to be able to better optimize the logic for how soon it is able to stop iterating over all the cells. Both of these methods also adjust the cached total amount of energy stored.
- setEnergy now wraps itself around those two methods
- Adding cells increase the cached total based on how much is in the added cell.
- Internalize MatrixData to ensure some data is not lost such as the rate limiting.
- Add getters for various now private data from MatrixData.
- Fixes a bug where saving and closing the world does makes some of the energy gained since last opening the world be lost. (The modified cells never got saved).
- Now that it keeps track of all the variables itself, the MatrixData only exposes queuing energy additional/removal rather than directly adding/removing it. The energy change then gets written to the cells when ticked by the casing that is running the matrix. This allows for the induction matrix to *basically* not have a performance impact when the net change is 0, and to have a much lower impact especially with multiple ports for other cases. This also has the added benefit that things like lag goggles properly blame (and show performance) of the induction matrix rather than offloading it into the things attempting to give power to the matrix.
- Removed old rate limit checks and allow the matrix to simulate how much it believes it has room for. This allows for the code to be a lot cleaner to read
- Various minor cleanup/readability improvements to the multiblock update protocol in general.
- When an induction matrix is broken it now also saves all queued energy to the cells to ensure there is no loss of power.
- Fixes the creative energy cube being capped at Integer.MAX_VALUE for output instead of at the value specified by the config. This bug was introduced by this commit https://github.com/mekanism/Mekanism/commit/1e0d04ab2b28c85d298932ddc650c9a7a11f5de1 and I believe it had to do with the old energy sending methods. I have done some testing and it properly seems to distribute power to various things that only support integers. So I believe this is no longer an issue because of the new emitting code introduced in 9.7.3
- Comparator support for Induction Ports.
- Should fix #3956 

## Some caveats of note (Also affects the current implementation):
- Because of double precision the total amount of energy stored may be incorrect at times/rounding to the wrong double. This is mainly when adding/removing very small amounts of energy when we are at very large amounts stored. Eventually we probably will want to switch to longs for cells/providers and BigInteger for the matrix itself.
- TOP is unable to show past MAX_INT for how much power the induction matrix has. We should either add support for their IBigPower (though eventually I think they plan on not packing that with TOP as it is supposed to just be part of McJtyLib; and that isn't a hard dep on TOP), or the option that might make more sense is to make a PR to TOP that supports `IStrictEnergyStorage` for rendering. The biggest thing with that is then ensuring the conversion ratio is proper.
- From what I can tell every time a block is placed (or even just a neighbor updates...) the entire multiblock gets recalculated. We probably will at some point want to not update from such minor changes, but for now this PR is not addressing that. (But because of this it does not actually break support for cells/providers being added/removed while the multiblock is actively formed.


pupnewfster:
Some small profiling tests for my test world
- Before these changes: https://sparkprofiler.github.io/#0u74ATHkP4
- After these changes: https://sparkprofiler.github.io/#HIurgIKetw
This is for a 16x16x16 induction matrix (cross chunk), with 9 input ports from creative energy cubes, and one output port not connected to anything.

src/main/java/mekanism/common/content/matrix/SynchronizedMatrixData.java
```diff
@@ -1,65 +1,274 @@
...
+    }
+
+    public double getEnergyPostQueue() {
+        //The reason we do remainingOutput - remainingInput when it logically appears that
+        // it should be the other way around, is because in reality our value is
+        // (transferCap - remainingInput) - (transferCap - remainingOutput)
+        // which simplifies to remainingOutput - remainingInput
+        // This is because remainingInput and remainingOutput go down if we have
+        // the corresponding one queued.
+        return cachedTotal - remainingInput + remainingOutput;
```
thiakil: Unless I'm being dense, that comment doesn't match the variables used in the calculation?

thiakil:
Yet to test, code looks fine enough

src/main/java/mekanism/common/content/matrix/SynchronizedMatrixData.java
pupnewfster: The comment is explaining why it is `cachedTotal - remainingInput + remainingOutput` instead of being `cachedTotal + remainingInput - remainingOutput` (what I originally thought it should be).

Given what we want to return is `cachedTotal + queuedReceiving - queuedSending`. I may look into it a bit more before I merge and cleanup the variables *slightly* so that it has a bit better precision by storing the queued receiving/sending in their own variable that is `> 0` and then merge it for the method for the remaining limit.

Issue #4464, opened 2017-06-13
Atomic Disassembler Path Block Tweak


ajakk:
It has been a slight annoyance that the atomic disassembler does not have the ability to create path blocks. It would be very much appreciated to see such a feature added. I notice there are a couple of duplicate tilling settings under the "Normal, slow, etc" settings for the tools, perhaps one of those dupes could be replaced with the aforementioned tweak.

I could attempt to implement such a thing myself, but I'm curious to hear others' opinions on the issue.

SuperFanatic:
Oh, definitely. I support this 100%. In fact, I would rather it just do path blocks *instead* of farmland altogether. Shovels? Those break every two minutes! Hoes? I still have the stone one I made on day 2!

ToMe25:
I would do it like if there is grass it creates grass paths, but if there is dirt it makes farmland instead.

ToMe25:
I will start implementing this now.

pupnewfster:
> I will start implementing this now.

Make sure to do it off of the latest 1.12 branch because otherwise you will probably run into merge conflicts. Given I did a bunch of formatting cleanup yesterday.

PR #5471, opened 2019-05-12
Atomic Disassembler Grass Path Block Tweak

ToMe25:

this PR just implements #4464

### DIFF
--- a/src/main/java/mekanism/common/item/ItemAtomicDisassembler.java
+++ b/src/main/java/mekanism/common/item/ItemAtomicDisassembler.java
@@ -171,7 +171,7 @@ public class ItemAtomicDisassembler extends ItemEnergized {
     public EnumActionResult onItemUse(EntityPlayer player, World world, BlockPos pos, EnumHand hand, EnumFacing side, float hitX, float hitY, float hitZ) {
         ItemStack stack = player.getHeldItem(hand);
         Block block = world.getBlockState(pos).getBlock();
-        if (!player.isSneaking() && (block == Blocks.DIRT || block == Blocks.GRASS || block == Blocks.GRASS_PATH)) {
+        if (!player.isSneaking() && block == Blocks.DIRT) {
             if (useHoe(stack, player, world, pos, side) == EnumActionResult.FAIL) {
                 return EnumActionResult.FAIL;
             }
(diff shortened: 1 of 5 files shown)


src/main/java/mekanism/common/item/ItemAtomicDisassembler.java
```diff
@@ -208,7 +229,7 @@ private EnumActionResult useHoe(ItemStack stack, EntityPlayer playerIn, World wo
         IBlockState iblockstate = worldIn.getBlockState(pos);
         Block block = iblockstate.getBlock();
         if (facing != EnumFacing.DOWN && worldIn.isAirBlock(pos.up())) {
-            if (block == Blocks.GRASS || block == Blocks.GRASS_PATH) {
+            if (block == Blocks.GRASS) {
```
ToMe25: This allows creation of Farmland out of Grass if the Player clicks on dirt.
Should i remove it?

pupnewfster: Haven't looked at things yet but part of me is wondering if overall it should be the reconstructor behaves as it did before when right clicking grass, dirt, or grass paths... And then have it be when you are sneaking that it instead turns grass into grass paths.

Either way I think also adding the ability that right clicking coarse dirt turns it into regular dirt would make sense for this PR. (For reference right clicking coarse dirt with a regular hoe will turn it into regular dirt.)

ToMe25: I think that's a bad idea as if you make path you most of the times want to be fast and you don't if you sneak.

pupnewfster: Actually thinking about it some more maybe a good middle ground would grass -> grass path, and then continue allowing the grass path -> farmland if they right click it again. Because if I wanted to turn grass into farmland I would get rather annoyed having to break it and place it again just to be able to convert it.

thiakil: For now let's go with the sneaking to make paths as pupnewfster said - I would think plenty of people already use the right click on grass to make farmland, just like a hoe does

ToMe25: So i guess that if keeps being there and i add Grass Paths to the farmland if right?

ToMe25: thiakil your comment just rendered after i replied, so i didn't replied to it but i think i will implement pupnewfster's new suggestion.

thiakil: normal right click -> pre-PR functionality; all pre-existing blocks
sneak+right click on grass -> grass path

Is how i understood it

ToMe25: that was the first suggestion but i would realy prefer the second one.

ToMe25: this is how i understood the second suggestion. is it okay like this or do you want me to change it to
"normal right click -> pre-PR functionality; all pre-existing blocks
sneak+right click on grass -> grass path"?

pupnewfster: I think the second suggestion (and how you changed things is fine for now). At some point we may want to give more configuration/upgrades to the atomic disassembler but for now this should be fine.

PR #5475, opened 2019-05-14
Enum Retrieval Index Safety

pupnewfster:
Improve safety of interacting with Enums. This way when there is invalid data we can handle it rather than directly causing a crash.

Note: Not yet tested, but it should work.

## Important things to do
- [ ] Add some warning messages for things that when null are invalid. So that people can know to report it if it somehow happens.
- [ ] Check other spots and see if there are things that *should* crash or at least throw a warning that got more safety than it should have.
- [x] JavaDoc EnumUtils
- [x] Potentially move EnumUtils to the API package for purposes of being able to use it for the Enum types that code is duplicated in that are in the API package.


src/main/java/mekanism/client/ClientProxy.java
```diff
@@ -590,13 +588,14 @@ public void registerBlockRenders() {
...
                 if (type == MachineType.BASIC_FACTORY || type == MachineType.ADVANCED_FACTORY || type == MachineType.ELITE_FACTORY) {
-                    RecipeType recipe = ((ItemBlockMachine) stack.getItem()).getRecipeTypeOrNull(stack);
+                    RecipeType recipe = ((ItemBlockMachine) stack.getItem()).getRecipeType(stack);
                     if (recipe != null) {
                         resource = "mekanism:" + type.getName() + "_" + recipe.getName();
                     }
                 }
                 return machineResources.get(resource);
             }
+            //TODO: This is annotated Nonnull
```
thiakil: Only because of Forge's hamfisted nonnull package annotations - any use of the `ItemMeshDefinition` interface checks for null.

I'd recommend adding an external annotation in IDEA for it

pupnewfster:
Closing as we are not actively working on 1.12 currently and this is done in a much better way already in the 1.15 branch.

PR #5490, opened 2019-05-24
Use Recipe Generics in more places, and some general readability improvements

pupnewfster:
- Use Generics in more places when interacting with recipes
- Removed Recipe oldRecipeName as I was incorrect about that having actually been needed.
- Pass machine type to super constructors for Machine Tile's rather than passing three pieces of the type up each time. Improves readability. Only noteable changes about that are:
  - Factory now sets base energy usage and storage is overwritten at end of constructor so probably doesn't matter
  - Electrolytic separator now sets usage instead of passing zero. This still seems to work properly and doesn't break anything.

### DIFF
--- a/src/main/java/mekanism/common/inventory/container/ContainerAdvancedElectricMachine.java
+++ b/src/main/java/mekanism/common/inventory/container/ContainerAdvancedElectricMachine.java
@@ -3,6 +3,7 @@ package mekanism.common.inventory.container;
 import javax.annotation.Nonnull;
 import mekanism.common.inventory.slot.SlotEnergy.SlotDischarge;
 import mekanism.common.inventory.slot.SlotOutput;
+import mekanism.common.recipe.inputs.AdvancedMachineInput;
 import mekanism.common.recipe.machines.AdvancedMachineRecipe;
 import mekanism.common.tile.prefab.TileEntityAdvancedElectricMachine;
 import mekanism.common.util.ChargeUtils;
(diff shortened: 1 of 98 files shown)


src/main/java/mekanism/common/inventory/container/ContainerAdvancedElectricMachine.java
```diff
@@ -80,12 +78,7 @@ public ItemStack transferStackInSlot(EntityPlayer player, int slotID) {
     }
 
     private boolean isInputItem(ItemStack itemstack) {
-        for (Entry<AdvancedMachineInput, ItemStack> entry : ((Map<AdvancedMachineInput, ItemStack>) tileEntity.getRecipes()).entrySet()) {
-            if (entry.getKey().itemStack.isItemEqual(itemstack)) {
-                return true;
-            }
-        }
-        return false;
+        return tileEntity.getRecipes().keySet().stream().anyMatch(input -> input.itemStack.isItemEqual(itemstack));
```
thiakil: I'm a little concerned at the overhead of streams here, a for loop on the keyset would be a tiny bit faster with less GC

Is `isItemEqual` the right one here too? or is NBT not an issue?

pupnewfster: If I remember correctly the JVM is actually pretty good at handling streams efficiently though I can change it if you want.

And I will look into if NBT is an issue or not (I just copied what we had before), though it is certainly possible for all these compares that it was just missing NBT as well.

thiakil: If you can find one of those comparison articles I'm prepared to eat my hat :D

thiakil: Don't get me wrong though, it's not a huge issue

pupnewfster: Looking at articles it seems that there is a bit of overhead. I think it was when I was first talking with a friend and he was theorizing that streams don't really have much of an overhead. (From what I can tell it seems the main use of them is to simplify parallelism which they have good performance for.) So later I think I will change a couple of these back to just being a simple for loop, and check if you are right about the NBT issue. Given I suspect it probably should check it.

PR #5491, opened 2019-05-25
Check ItemStack NBT for comparison in more places

pupnewfster:
- Checks NBT in more places to see if two itemstacks are stackable/the same
- Also fixes Gas Conversions checking stack size for if two items are the same.

Remaining spots:
- Filters (I believe they are meant to not be "advanced" enough to support NBT though if this is not the case I can modify it to check the NBT
- The miner has a couple spots for the replace stack comparison which I believe is the same as the filters I mentioned above

### DIFF
--- a/src/main/java/mekanism/common/item/ItemAtomicDisassembler.java
+++ b/src/main/java/mekanism/common/item/ItemAtomicDisassembler.java
@@ -41,6 +41,7 @@ import net.minecraft.world.World;
 import net.minecraftforge.common.util.Constants.WorldEvents;
 import net.minecraftforge.fml.relauncher.Side;
 import net.minecraftforge.fml.relauncher.SideOnly;
+import net.minecraftforge.items.ItemHandlerHelper;
 
 public class ItemAtomicDisassembler extends ItemEnergized {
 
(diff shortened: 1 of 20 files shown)


src/main/java/mekanism/common/item/ItemAtomicDisassembler.java
```diff
@@ -365,8 +368,7 @@ public void loop(Coord4D pointer) {
                 Coord4D coord = pointer.offset(side);
                 ItemStack blockStack = coord.getBlock(world).getPickBlock(coord.getBlockState(world), rayTraceResult, world, coord.getPos(), player);
                 if (coord.exists(world) && checkID(coord.getBlock(world)) &&
-                    (stack.isItemEqual(blockStack) || (coord.getBlock(world) == startBlock && OreDictCache.getOreDictName(stack).contains("logWood")
-                                                       && coord.getBlockMeta(world) % 4 == stack.getItemDamage() % 4))) {
+                    (ItemHandlerHelper.canItemStacksStack(stack, blockStack) || (coord.getBlock(world) == startBlock && isWood && coord.getBlockMeta(world) % 4 == stack.getItemDamage() % 4))) {
```
thiakil: that `meta % 4` check should probably get a `TODO` for visiting later - as it would only be relevant to vanilla logs (and I'm not 100% sure on what it does anyway, something to do with the rotations perhaps)

PR #5496, opened 2019-05-31
Use sendToAllTracking for update packets

pupnewfster:
- Switch to using SimpleImpl sendToAllTracking for update packets, rather than calculating a bunch of the stuff ourselves.
- Add helper methods to cleanup readability when it comes to sending update packets

## To be looked at
- [ ] The networks (GasNetwork, EnergyNetwork, etc) still use the old `sendToReceivers` method. Not sure if we should just continue with how this PR is and ignore the Y for them, or if there is some good way to switch it over. (If some way is decided upon that switches it over then we do not need to bump the API version as no changes/additions to Range4D will have been made from an API standpoint.)

### DIFF
--- a/src/main/java/mekanism/common/PacketHandler.java
+++ b/src/main/java/mekanism/common/PacketHandler.java
@@ -7,6 +7,8 @@ import java.util.UUID;
 import javax.annotation.Nonnull;
 import mekanism.api.Coord4D;
 import mekanism.api.Range4D;
+import mekanism.api.TileNetworkList;
+import mekanism.common.base.ITileNetwork;
 import mekanism.common.config.MekanismConfig;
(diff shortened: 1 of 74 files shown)


src/main/java/mekanism/common/PacketHandler.java
```diff
@@ -291,11 +293,35 @@ public void sendToCuboid(IMessage message, AxisAlignedBB cuboid, int dimId) {
         }
     }
 
+    public <TILE extends TileEntity & ITileNetwork> void sendToAllTracking(TILE tile) {
```
thiakil: I'd have called this `sendUpdatePacket` myself

thiakil:
This looks fine to me - the transmitter net stuff is about as good as we'll get; sending to any player whose view distance is within the bounds of the net

PR #5529, opened 2019-06-20
Reworked PipeUtils and GasUtils side access

DarkGuardsman:
Rewrote pipe and gas utilities orginally to fix chunk loading ghosting. In which accessing a block in another chunk would ghost load the chunk. While I was in the methods I rewrote them to be more reusable in terms of code. Replacing existing loops and calls with lambda driven versions. This makes the code easier to read and recycles common implementation. While at the same time not costing anything in terms of performance.

### DIFF
--- a/src/main/java/mekanism/common/util/CapabilityUtils.java
+++ b/src/main/java/mekanism/common/util/CapabilityUtils.java
@@ -21,6 +21,22 @@ public final class CapabilityUtils {
         return provider.getCapability(cap, side);
     }
 
+    /**
+     * Runs an action on a capability if it exists
+     *
+     * @param provider - provider hosting the capability
+     * @param cap      - capability type
+     * @param side     - side to access
+     * @param action   - handler for completing the action on the capability
+     * @param <T>      - type to return
+     */
+    public static <T> void runIfCap(ICapabilityProvider provider, Capability<T> cap, EnumFacing side, Consumer<T> action) {
+        final T capability = getCapability(provider, cap, side);
+        if (capability != null) {
+            action.accept(capability);
+        }
+    }
+
     public static <T> OptionalCapability withCapability(ICapabilityProvider provider, Capability<T> cap, EnumFacing side, Consumer<T> consumer) {
         return new OptionalCapability(provider).orElseWith(cap, side, consumer);
     }
--- a/src/main/java/mekanism/common/util/GasUtils.java
+++ b/src/main/java/mekanism/common/util/GasUtils.java
@@ -27,13 +27,10 @@ import net.minecraft.world.World;
 public final class GasUtils {
 
     public static IGasHandler[] getConnectedAcceptors(BlockPos pos, World world, Set<EnumFacing> sides) {
-        IGasHandler[] acceptors = new IGasHandler[]{null, null, null, null, null, null};
-        for (EnumFacing orientation : sides) {
-            TileEntity acceptor = world.getTileEntity(pos.offset(orientation));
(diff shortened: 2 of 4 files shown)


src/main/java/mekanism/common/util/CapabilityUtils.java
```diff
@@ -21,6 +22,38 @@ public static boolean hasCapability(ICapabilityProvider provider, Capability<?>
...
+     * Checks if the capability exists then retrieves it for use
+     *
+     * @param provider  - provider hosting the capability
+     * @param cap       - capability type
+     * @param side      - side to access
+     * @param <T>       - type to return
+     * @return null if not found/valid or cap
+     */
+    public static <T> T get(ICapabilityProvider provider, Capability<T> cap, EnumFacing side) {
+        if (provider == null || cap == null || !hasCapability(provider, cap, side)) {
```
thiakil: `hasCapability` is unnecessary here, `getCapability` just needs to be null checked instead

DarkGuardsman: The previous code was running 'has' then 'get'... as well it's proper to run 'has' then 'get' for the API usage.

thiakil: Previous code is legacy on the cards for a rejig. Do you call `containsKey` before you get an element from a `Map`?

From the javadoc: `Basically, this method functions analogously to {@link Map#containsKey(Object)}.`

DarkGuardsman: I honestly don't want to argue over it for the 4th or 5th time. We have been over this in DM's on discord before. All I'm doing is replicating existing code, reducing to reusability, and solving the ghosting problem. 

I could have easily just slotted in two ugly block exists calls and been done. However, I decided to spend some extra time by making the code easier to read and more reusable. In which I didn't make assumptions on how you wanted the code. Instead, I replicated what was already in place. Which really is just a 1 line change on your side if you don't like it.

DarkGuardsman: Since I want this PR pulled I made the change rather than deal with fighting you over it.

DarkGuardsman:
Force pushed changes that include format fixes using the provide style sheet in the repo. As well changed has & get cap to just get cap since Thiakil is unhappy about it.

src/main/java/mekanism/common/util/GasUtils.java
```diff
@@ -27,13 +27,10 @@
...
-        IGasHandler[] acceptors = new IGasHandler[]{null, null, null, null, null, null};
-        for (EnumFacing orientation : sides) {
-            TileEntity acceptor = world.getTileEntity(pos.offset(orientation));
-            if (CapabilityUtils.hasCapability(acceptor, Capabilities.GAS_HANDLER_CAPABILITY, orientation.getOpposite())) {
-                acceptors[orientation.ordinal()] = CapabilityUtils.getCapability(acceptor, Capabilities.GAS_HANDLER_CAPABILITY, orientation.getOpposite());
-            }
-        }
+        final IGasHandler[] acceptors = new IGasHandler[]{null, null, null, null, null, null};
+        EmitUtils.forEachSide(world, pos, sides, (tile, side) -> {
+            acceptors[side.ordinal()] = CapabilityUtils.get(tile, Capabilities.GAS_HANDLER_CAPABILITY, side.getOpposite());
```
pupnewfster: Shouldn't this be `getCapability` instead of `get`?

DarkGuardsman: Yes, looks like my local failed to do git.... its been a problem with gitkaken recently.

PR #5533, opened 2019-06-23
Added checks for is block loaded to several methods

DarkGuardsman:
Searched for usage of World#getTileEntity(BlockPos) replacing any instance that looked close to accessing another block with a check for isBlockLoaded.

### DIFF
--- a/src/main/java/mekanism/common/tile/TileEntityBoundingBlock.java
+++ b/src/main/java/mekanism/common/tile/TileEntityBoundingBlock.java
@@ -17,20 +17,30 @@ import net.minecraft.util.EnumFacing;
 import net.minecraft.util.math.BlockPos;
 import net.minecraftforge.common.capabilities.Capability;
 
+/**
+ * Multi-block used by wind turbines, solar panels, and other machines
+ */
 public class TileEntityBoundingBlock extends TileEntity implements ITileNetwork {
 
-    public BlockPos mainPos = BlockPos.ORIGIN;
+    private BlockPos mainPos = BlockPos.ORIGIN;
 
     public boolean receivedCoords;
 
     public int prevPower;
 
     public void setMainLocation(BlockPos pos) {
-        receivedCoords = true;
+        receivedCoords = pos != null;
         if (!world.isRemote) {
             mainPos = pos;
             Mekanism.packetHandler.sendUpdatePacket(this);
         }
+    }
+
+    public BlockPos getMainPos() {
+        if (mainPos == null) {
+            mainPos = BlockPos.ORIGIN;
+        }
+        return mainPos;
     }
 
     @Override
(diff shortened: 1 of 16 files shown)


src/main/java/mekanism/common/tile/TileEntityBoundingBlock.java
```diff
@@ -17,22 +17,32 @@
 import net.minecraft.util.math.BlockPos;
 import net.minecraftforge.common.capabilities.Capability;
 
+/**
+ * Multi-block used by wind turbines, solar panels, and other machines
+ */
 public class TileEntityBoundingBlock extends TileEntity implements ITileNetwork {
 
-    public BlockPos mainPos = BlockPos.ORIGIN;
+    private BlockPos _mainPos = BlockPos.ORIGIN;
```
pupnewfster: I am fine with the change of making this private, though I would rather the name remain `mainPos` rather than becoming `_mainPos`

src/main/java/mekanism/common/tile/TileEntityDigitalMiner.java
```diff
@@ -448,8 +448,12 @@ public TileEntity getPullInv() {
     }
 
     public TileEntity getEjectInv() {
```
pupnewfster: I know it wasn't marked before but we probably should add `@Nullable` to this and the `getEjectTile()`

src/main/java/mekanism/common/tile/TileEntityBoundingBlock.java
DarkGuardsman: its a bit of a habit of mine to place _ in front of internal fields that should never be used outside set methods.

src/main/java/mekanism/common/tile/TileEntityDigitalMiner.java
DarkGuardsman: I think that should be left to another PR or changeset. As there are tons of methods missing those type of annotations.

src/main/java/mekanism/common/tile/TileEntityBoundingBlock.java
```diff
@@ -41,10 +51,18 @@ public void validate() {
         }
     }
 
+    public TileEntity getMainTile() {
+        if (receivedCoords && world.isBlockLoaded(getMainPos())) {
```
thiakil: This doesn't look like it will currently work on the client, as that field is not set there (given the setter for `mainPos` isn't getting used)

DarkGuardsman: I didn't see any issues in testing but its hard to tell with the amount of legacy code left laying around. I'll take a look first chance I get but will be working 10h days this week so might not get to it right away.

DarkGuardsman:
Pushed changes to move '_mainPos' back 'mainPos' and fixed client not having the receivedCoord field. I recommend we sync it as the client will not know if the block pos is valid or not. However, at the same time it can't be null without causing issues. 

@pupnewfster final on local fields will not effect anything but its something I've gotten into the habit of doing to make it clear a field doesn't change. As well in theory it helps the compiler and JIT run a little nicer. Though this is not the reason I do it.

pupnewfster:
> @pupnewfster final on local fields will not effect anything but its something I've gotten into the habit of doing to make it clear a field doesn't change. As well in theory it helps the compiler and JIT run a little nicer. Though this is not the reason I do it.

Sounds good, and I suppose if we ever want to change the logic slightly about how some of the things work we can always remove it anyways.

I will try to get a chance to test the changes sometime this week.

DarkGuardsman:
When you merge this change and the other I can work on a few different PRs. Want to avoid too many PRs right now due to possible merge conflicts. My next few ideas is to do a pass to move NBT save tags to static finals for sanity reasons. As well to fix the solar panel exploits I told you about in DMs.

PR #5537, opened 2019-06-25
Optimized Component Ejector

pupnewfster:
- Fixes Digital miner not combining stacks for ejecting
- Combine TransitRequests for ComponentEjector where possible, and only eject once a second instead of also every time an operation occurs.
- Removed ComponentEjector trackers as from what I can tell they did not actually do anything

Some profiling hints from my test world:
Before: https://sparkprofiler.github.io/#hksdBcW8EO https://gyazo.com/32ac3e0956481b904fa2accc194d1990
After: https://sparkprofiler.github.io/#pUOsQgm7zP https://gyazo.com/497480ae84fdd6f1750139009877c230

### DIFF
--- a/src/main/java/mekanism/common/tile/component/TileComponentEjector.java
+++ b/src/main/java/mekanism/common/tile/component/TileComponentEjector.java
@@ -66,6 +66,7 @@ public class TileComponentEjector implements ITileComponent {
         sideData = ejector.sideData;
     }
 
+    //TODO: Figure out if trackers can just be removed or what the goal of it was and reimplement if needed
     private Set<EnumFacing> getTrackedOutputs(TransmissionType type, int index, Set<EnumFacing> dirs) {
         Set<EnumFacing> sides = EnumSet.noneOf(EnumFacing.class);
         int tracker = trackers.get(type)[index];
@@ -80,14 +81,12 @@ public class TileComponentEjector implements ITileComponent {
 
     @Override
     public void tick() {
-        if (tickDelay == 0) {
-            if (sideData.get(TransmissionType.ITEM) != null) {
-                outputItems();
-            }
-        } else {
-            tickDelay--;
-        }
         if (!tileEntity.getWorld().isRemote) {
+            if (tickDelay == 0) {
+                outputItems();
+            } else {
+                tickDelay--;
+            }
             eject(TransmissionType.GAS);
             eject(TransmissionType.FLUID);
         }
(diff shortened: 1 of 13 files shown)


src/main/java/mekanism/common/tile/component/TileComponentEjector.java
```diff
@@ -139,52 +123,53 @@ private void ejectFluid(Set<EnumFacing> outputSides, FluidTank tank) {
...
+                if (ejectMap.isEmpty()) {
                     break;
                 }
             }
+            BlockPos offsetPos = tileEntity.getPos().offset(side);
+            TileEntity tile = tileEntity.getWorld().isBlockLoaded(offsetPos) ? tileEntity.getWorld().getTileEntity(offsetPos) : null;
+            ILogisticalTransporter capability = CapabilityUtils.getCapability(tile, Capabilities.LOGISTICAL_TRANSPORTER_CAPABILITY, side.getOpposite());
+            TransitResponse response;
+            if (capability == null) {
+                response = InventoryUtils.putStackInInventory(tile, ejectMap, side, false);
```
thiakil: does this need to be moved outside the side loop?

thiakil: (might be TransporterUtils.insert that this applies to)

src/main/java/mekanism/common/tile/component/TileComponentEjector.java
```diff
@@ -139,52 +123,53 @@ private void ejectFluid(Set<EnumFacing> outputSides, FluidTank tank) {
...
+        TransitRequest ejectMap = null;
+        for (EnumFacing side : outputs) {
+            if (ejectMap == null) {
+                ejectMap = getEjectItemMap(data);
+                if (ejectMap.isEmpty()) {
                     break;
                 }
             }
+            BlockPos offsetPos = tileEntity.getPos().offset(side);
+            TileEntity tile = tileEntity.getWorld().isBlockLoaded(offsetPos) ? tileEntity.getWorld().getTileEntity(offsetPos) : null;
```
pupnewfster: If 1.12 is merged back into this, then we can do `tile = MekanismUtils.getTile` to check for the isBlockLoaded in an easier way to read

src/main/java/mekanism/common/tile/component/TileComponentEjector.java
pupnewfster: I believe this is the correct spot as it is trying on specific sides. `InventoryUtils#putStackInInventory` uses tile being inserted to as a param, and `TransporterUtils#insert` uses tile from. The response needs to be in the loop as it the methods require side information.

thiakil:
Merged manually.

I halved the tickDelay for the eject component (half a second from a full second) - ejecting is now a lot slower since it's only timer based. We could consider doing the ejection at the end of a tile's `tick` method instead if there are complaints

Issue #4838, opened 2017-12-01
Disallow snow floating on solar panels


Moobien:
Hey guys just a query which could end up in a suggestion.

I noticed that snow 'settles' or should I say hovers over solar panels. They seem to still work fine. Was wondering if they can be made to not allow snow on them?

PR #5539, opened 2019-06-25
Overwrite getBlockFaceShape for various blocks

pupnewfster:
Fixes how various blocks interact with:
- button placement
- redstone placement
- torch placement
- snow falling (#4838)
- fence/wall connections

There are still some blocks with "complex" models that return SOLID as their BlockFaceShape when potentially it would make more sense if they did not; however these are minor and for the most part any big outliers have been dealt with by this PR.

## To be done before merging
- [x] Make a system for BlockBounding to figure out if the offset block should be solid or not (parts of the digital miner, and also the seismic vibrator)


thiakil:
lgtm

Issue #5413, opened 2019-04-19
Formulaic Assemblicator consume items with transformDamage added by crafttweaker...


Fireztonez:
#### Issue description:
During somes tests, I fond a problem with the Formulaic Assemblicator or possibly 2 different problem;
- First, I have add somes recipe for somes Mystical Agriculture Infusion Crystal (created by ContentTweaker) make with Crafttweaker. When I try the recipe in a regular craftting table, the crystal stay in the grid but get damage... In the Formulaic Assemblicator, the crystal is simply consume when making the first set of Essence... You can see the issue in video: https://streamable.com/x3slq
And heres the crafting recipe:
```
var crystals = [
    <contenttweaker:infusioncrystal_tier0:*>.transformDamage(),
    <contenttweaker:infusioncrystal_tier1:*>.transformDamage(),
    <contenttweaker:infusioncrystal_tier2:*>.transformDamage(),
    <contenttweaker:infusioncrystal_tier3:*>.transformDamage(),
    <contenttweaker:infusioncrystal_tier4:*>.transformDamage()
] as IIngredient[];

recipes.addShaped("mystical_prudentium", <mysticalagriculture:crafting:1>,
    [[null, <mysticalagriculture:crafting>, null],
     [<mysticalagriculture:crafting>, crystals[0]|crystals[1]|crystals[2]|crystals[3]|crystals[4] , <mysticalagriculture:crafting>],
     [null, <mysticalagriculture:crafting>, null]]);
```
- With the basic Infusion Crystal from Mystical Agriculture, the crystal should probably stay in the grid, but this one goes in the output slot after making the first craft, making the automation really annoying. Here a demonstration video:
https://streamable.com/2z056

If you want to test with added recipe, you can simply copy my code and use a sword or thing like that at the place of the crystal and I think it will do the exact same issue.

#### Steps to reproduce:
1. Create a recipe with transformDamage in Crafttweaker
2. Try to craft in the Formulaic Assemblicator
3. Enjoy your item waste!

#### Version (make sure you are on the latest version before reporting):
**Forge:** 14.23.5.2824
**Mekanism:** 9.7.1.371
**Other relevant version:** Crafttweaker: 4.1.17

#### If a (crash)log is relevant for this issue, link it here: 
There have no crash and I no logs for that. But if needed I can send you more informations or scripts from my modpack if it can help...

Thank's for you amazing work, and making Mekanism great again! Your team is awesome! Long life to Mekanism!!
Fireztonez

[...]

PR #5542, opened 2019-06-27
Assemblicator Fixes for Transforming recipes

thiakil:
- Fixes the assemblicator to use `IRecpie.getRemainingItems` instead of only handling containers (see #5413)
- Changes the weird MekanismUtils method to just be used in the assemblicator for repair recipes
- Caches the IRecipe used when not using a formula (yay performance)

## To-Do
- [x] Check formula's NBT sensitivity in the above situation - using the vanilla test script from #5413 it shows the sword as red once it's been damaged
- [x] bug: if you have say a stack of 2 in the recipe grid on loading the world and opening the gui until the gui is refreshed it will show an x and the craft buttons will be grayed out.
- [ ] decide what to do with remaining items - output or input slots? configurable?
- [ ] Test, test, test (on face value the issue is solved)

### DIFF
--- a/src/main/java/mekanism/client/gui/GuiFormulaicAssemblicator.java
+++ b/src/main/java/mekanism/client/gui/GuiFormulaicAssemblicator.java
@@ -32,7 +32,6 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.util.ResourceLocation;
 import net.minecraftforge.fml.relauncher.Side;
 import net.minecraftforge.fml.relauncher.SideOnly;
-import net.minecraftforge.items.ItemHandlerHelper;
 
 @SideOnly(Side.CLIENT)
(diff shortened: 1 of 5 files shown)


pupnewfster:
The assemblicator also has a lot of cases where `inventory.get` is called many many times on the same slot in places we can just assign it to a variable. So we probably also should cleanup some of those uses so it is more clear what it is referencing and so it doesn't need to do the lookup 5 times when it can be done once.

pupnewfster:
Fixed the gui, and also cleaned up some of the inventory.get calls.
- One "bug" I have ran into is that if you have say a stack of 2 in the recipe grid on loading the world and opening the gui until the gui is refreshed it will show an `x` and the craft buttons will be grayed out.
- Another thing that feels slightly less than ideal but probably is fine is that excess outputs do not automatically get put back into the recipe grid if they are still valid. Makes the vanilla example linked above not allow for stack crafting.

Issue #5054, opened 2018-05-18
Suggestion: Coal gasification to produce hydrogen gas and sulfur


WrongWeekend:
It would be useful to add a recipe in the PRC or another machine that allows for combining coal or charcoal with oxygen and water, steam, and/or water vapor to produce hydrogen gas and sulfur. This process is used in the real world and would be an effective way to use excess oxygen and coal to produce hydrogen for energy generation rather than just energy storage in the form of a gas.

ToMe25:
I am implementing this and #5253.

pupnewfster:
Version [9.8.0](https://www.curseforge.com/minecraft/mc-mods/mekanism/files/2746197) was just released and should sync to the twitch client within the next couple hours. It ended up being added as another "mode" rather than a hotkey.

PR #5547, opened 2019-06-29
Implement #5054(Coal Gasification)

ToMe25:
* Added PRC Recipes for Coal Gasification making Hydrogen and Sulfur.(#5054)
* Added a Enrichment Chamber Recipe to convert Sulfur back into Gunpowder.

Using Coal Gasification you can get 18900J out of one coal instead of ~6kJ with a Heat Generator.

### DIFF
--- a/src/main/java/mekanism/common/Mekanism.java
+++ b/src/main/java/mekanism/common/Mekanism.java
@@ -546,8 +546,6 @@ public class Mekanism {
                   new ItemStack(MekanismItems.Substrate, 8), new GasStack(MekanismFluids.Oxygen, 10), 200, 400);
             RecipeHandler.addPRCRecipe(new ItemStack(Items.COAL, 1, OreDictionary.WILDCARD_VALUE), new FluidStack(FluidRegistry.WATER, 100), new GasStack(MekanismFluids.Oxygen, 100),
                   new ItemStack(MekanismItems.OtherDust, 1, 3), new GasStack(MekanismFluids.Hydrogen, 100), 0, 100);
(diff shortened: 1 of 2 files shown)


src/main/java/mekanism/common/Mekanism.java
```diff
@@ -544,6 +544,8 @@ public static void addRecipes() {
                   new GasStack(MekanismFluids.Oxygen, 10), new ItemStack(MekanismItems.Polyethene), new GasStack(MekanismFluids.Oxygen, 5), 1000, 60);
             RecipeHandler.addPRCRecipe(new ItemStack(MekanismItems.Substrate), new FluidStack(FluidRegistry.WATER, 200), new GasStack(MekanismFluids.Ethene, 100),
                   new ItemStack(MekanismItems.Substrate, 8), new GasStack(MekanismFluids.Oxygen, 10), 200, 400);
+            RecipeHandler.addPRCRecipe(new ItemStack(Items.COAL, 1, OreDictionary.WILDCARD_VALUE), new FluidStack(FluidRegistry.WATER, 100), new GasStack(MekanismFluids.Oxygen, 100),
```
thiakil: I don't think this one is necessary, since you're doing it by oredict later?

ToMe25: It is as i only add Coal Block and dust later with OreDict. but i can move/change that to OreDict if you think that would be better.

ToMe25: so just tell me if you think it should be coal OreDict instead of just Coal.

thiakil: Ah, right you are. I got mixed up with dust. Leave it as it is, the recipes will get redone later anyway when we fix it to be more flexible

PR #5551, opened 2019-07-03
IFluidHandlerWrapper implementation cleanup

### LINKED ISSUE #5550: Mechanical pipe not pulling from quantum entangloporter
I have the same issue as a previously closed one, #5100 

mek_entanglo_no_pull

As you can see, I have ult mech pipes pulling from a creative tank going into a quantum entangloporter (all sides set as input), then another quantum entangloporter on the same channel (all sides set as output) where ult mech pipes _should_ be pulling and sending water to a fluid trash can. This is not happening.

Mods:
Mekanism 9.7.8.379
Mekanism: Generators 9.7.8.379
Just Enough Items 4.15.0.268
Extra Utilities 2 1.9.9
Forge 14.23.5.2838

pupnewfster:
- Add some default implementations into IFluidHandlerWrapper so less code is needed when only implementing one of fill and drain support. (Both default to being unsupported).
- Change where null fluid stacks get caught for IFluidHandlerWrapper to ensure the logic is simpler about what happens where. This includes changing some things from `@Nullable` to `@Nonnull`
- Add assumptions that `canDrain` and `canFill` get called before `drain` and `fill` respectively.
- Default drain by fluidstack type to drain by amount as `canDrain` already checked the type is correct
- Cleanup Entangloporter InventoryFrequency to use TileUtils for gas and fluid syncing
- Unify some of the fluid comparison code, checking NBT of fluidstacks in even more places for the fluid now.
- Implement `drain` method for `TileEntityTurbineVent` as previously only `canDrain` was implemented.
- Some ejecting from multiblocks now use the `EmitUtils#forEachSide` to ensure the tile it is trying to eject to is loaded.

This PR fixes #5550 and various other oddities with our fluid handlers by cleaning up the IFluidHandlerWrapper code a good bit making it easier to read/manage.

### DIFF
--- a/src/main/java/mekanism/common/base/IFluidHandlerWrapper.java
+++ b/src/main/java/mekanism/common/base/IFluidHandlerWrapper.java
@@ -1,5 +1,6 @@
 package mekanism.common.base;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import net.minecraft.util.EnumFacing;
 import net.minecraftforge.fluids.FluidStack;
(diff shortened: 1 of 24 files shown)


src/main/java/mekanism/common/base/IFluidHandlerWrapper.java
```diff
@@ -1,21 +1,46 @@
...
+    default FluidStack drain(EnumFacing from, int maxDrain, boolean doDrain) {
+        return null;
+    }
+
+    default boolean canFill(EnumFacing from, @Nonnull FluidStack fluid) {
+        return false;
+    }
+
+    /**
+     * null fluid needs to have amounts end up getting specified for draining rather than using the drain with a null stack
```
thiakil: I don't understand this comment

src/main/java/mekanism/common/tile/TileEntityFluidicPlenisher.java
```diff
@@ -337,44 +336,21 @@ public boolean hasTank(Object... data) {
     }
 
     @Override
-    public FluidStack drain(EnumFacing from, @Nullable FluidStack resource, boolean doDrain) {
-        if (resource != null && fluidTank.getFluid() != null && fluidTank.getFluid().getFluid() == resource.getFluid() && from == EnumFacing.UP) {
-            return drain(from, resource.amount, doDrain);
-        }
-        return null;
+    public int fill(EnumFacing from, @Nonnull FluidStack resource, boolean doFill) {
+        return fluidTank.fill(resource, true);
```
thiakil: Is the `true` there right? (I know its like that before your changes)

pupnewfster: I will look into it tomorrow if you don't end up doing so before then. I didn't even notice that it was specifically set to true, that is probably a good catch.

src/main/java/mekanism/common/base/IFluidHandlerWrapper.java
pupnewfster: Actually it sort of is invalid/doesn't make sense after I changed some nullability declarations for other methods.

thiakil:
Merged manually.

Comment was removed and `doFill` param was passed on

PR #5556, opened 2019-07-06
Implement Wood Gasification

ToMe25:
Implement the Wood counterpart for Coal Gasification
* Implemented Wood Gasification
* Added a Recipe to convert Sawdust to Charcoal
* Added Slab processing to the Sawmill

Using Wood Gasification you can get 18900J out of one log,
9450J out of one plank or sawdust,
4725J out of one slab and
1890J out of one stick.

### DIFF
--- a/src/main/java/mekanism/common/integration/OreDictManager.java
+++ b/src/main/java/mekanism/common/integration/OreDictManager.java
@@ -17,7 +17,6 @@ import mekanism.common.Resource;
 import mekanism.common.config.MekanismConfig;
 import mekanism.common.recipe.RecipeHandler;
 import mekanism.common.recipe.RecipeHandler.Recipe;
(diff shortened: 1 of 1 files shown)


ToMe25:
btw: should this produce Sulfur?
and if yes how much?

thiakil:
Is this based on a real life process of some kind that you can link a reference to?

Sawdust->Coal, shouldn't that be (char)coal _dust_?

ToMe25:
Yes you are right it should be charcoal dust, but as Mekanism has no Charcoal dust its Charcoal.

And yes this kind of is based on a realy life process called "Wood Gasification", but it gives little amounts of charcoal dust and i don't know how to implement that as i can't give the player 0.1 charcoal and i don't want to increase the recipe to use 10 items at once.

PR #5561, opened 2019-07-10
Fix a few issues with sound

pupnewfster:
- Adds some null safety to PlayerSounds and only get the player object when needed
- Fix flame thrower not playing sounds (it still doesn't play a sound when first equipped but unused for other players)
- Cleaned up various logic and method readability in ClientTickHandler and CommonPlayerTickHandler
- Keep track of which sounds we have for a player by UUID instead of EntityPlayer reference. Also use a WeakReference in PlayerSound instead of directly saving the EntityPlayer object (gracefully falls back to last known x, y, z position if the reference gets cleaned up while a reference to PlayerSound is still stored). This should allow the GC to do its job a lot better.
- Updated PacketFlamethrowerData to be like the Jetpack and GasMask ones in terms of MODE_CHANGE, UPDATE, and FULL options

I did do some testing, though it would probably not hurt to double check some of the logic/test more thoroughly on multiplayer.

One other thing to note I believe that at least one case where the player looked up is null may have to do with the fact I believe it syncs data for clients in all dimensions not just the one that is the same as our player. Not sure if/where we should catch that if we decide to but for the most part I believe these changes improve the handling of the various player sounds.

### DIFF
--- a/src/main/java/mekanism/client/ClientPlayerTracker.java
+++ b/src/main/java/mekanism/client/ClientPlayerTracker.java
@@ -11,7 +11,7 @@ public class ClientPlayerTracker {
 
     @SubscribeEvent
     public void onPlayerChangedDimension(PlayerChangedDimensionEvent event) {
-        Mekanism.playerState.clearPlayer(event.player);
+        Mekanism.playerState.clearPlayer(event.player.getUniqueID());
         Mekanism.freeRunnerOn.remove(event.player.getUniqueID());
     }
 }
--- a/src/main/java/mekanism/client/ClientTickHandler.java
+++ b/src/main/java/mekanism/client/ClientTickHandler.java
@@ -20,7 +20,6 @@ import mekanism.common.item.ItemConfigurator;
 import mekanism.common.item.ItemConfigurator.ConfiguratorMode;
 import mekanism.common.item.ItemFlamethrower;
 import mekanism.common.item.ItemFreeRunners;
-import mekanism.common.item.ItemGasMask;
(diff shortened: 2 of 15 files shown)


thiakil:
The null issue I think is to do with it not checking if it actually got a player reference when it receives the packet with the uuid, it just adds it to the map regardless. If we put a `Objects.requireNonNull` in there I bet that's where it'd crash

pupnewfster:
> The null issue I think is to do with it not checking if it actually got a player reference when it receives the packet with the uuid, it just adds it to the map regardless. If we put a `Objects.requireNonNull` in there I bet that's where it'd crash

I will look at it again tomorrow, for now I am just having it not do anything if the player is null, but I may actually be able to catch it earlier to save a bit more on memory.

pupnewfster:
Actually upon further investigation, I think this needs a bigger overhaul as our SoundHandler (and PlayerSound) are keeping track of a reference to EntityPlayer which from my understanding is not a good thing to do for GC interacting with things. The only reason it even works currently I believe is given it keeps track of it in an `IdentityHashMap` then after dimension changes the object is different.

pupnewfster:
Went ahead and rebased this onto the latest 1.12, because even if it didn't have any merge conflicts with it, for testing it still is probably better to have it against the latest version. (Especially given how many changed happened due to the merging of the Rendering PR)

pupnewfster:
> Looks good code wise, but I can't seem to get the active sound to work in my dev environment. Does it work for you?

Ya tested again and it still works fine so going to go ahead and merge this.

Issue #5557, opened 2019-07-08
Personal Chest Dupe Bug


Focamacho:
#### Issue description:
You can duplicate items using two personal chests.

#### Steps to reproduce:
_1._ Put 2 personal chests in the hotbar.
_2._ Open the first, put some items inside, and close.
_3._ Now u need to be fast(or have some lag). Open & drop the personal chest, and instantly change your hand to the other personal chest in the hotbar.
_4._ It's all. See the video for more info.
Video: https://www.youtube.com/watch?v=qlaD_9TIa08

#### Version (make sure you are on the latest version before reporting):
**Forge:** 14.23.5.2836
**Mekanism:** 9.7.9.380

pupnewfster:
Out of curiosity is this on a server or in single player? (Mainly asking so that I can try to reproduce it in the same environment type when it potentially may not happen in the other one)

Focamacho:
Server, because it's easier to do with some lag. You can reproduce it in singleplayer too, but it's a lot harder.

pupnewfster:
Haven't tried to replicate this yet or make a fix, but I believe what will fix it is instead of us passing the hand that is opening the chest between the server and client, if we also pass the slot ID that would fix it being able to have any doubt about what chest is really being opened. (More just putting this here as a reminder for when I get a chance to look at it more).

Also adding this to the 9.8.0 milestone as even if I don't end up replicating it (or even trying to replicate it), I probably will make the change I mentioned above which *should* fix it.

pupnewfster:
Version [9.8.0](https://www.curseforge.com/minecraft/mc-mods/mekanism/files/2746197) was just released and should sync to the twitch client within the next couple hours. I rewrote how the GUI gets opened for the Personal chest so this *should* be fixed as of this release, but if it isn't feel free to reopen/comment here and I can look into it more.

Focamacho:
Hi, maybe I'm a little late but I decided to test the bug again today, and it keeps happening.
I could replicate the bug in singleplayer using this little macro:
![image](https://user-images.githubusercontent.com/48870573/64488928-01101e80-d224-11e9-9d59-d0e4df757247.png)

PvPsFinest:
What did you use to test the macro?

thiakil:
@Focamacho Can you please confirm which version of Mekanism you tested that with?

Focamacho:
Latest version of Mekanism (1.12.2-9.8.3.390)

PR #5566, opened 2019-07-16
Improve implementation of opening various Guis

pupnewfster:
- This should help mitigate any chance of duplication such as seen in #5557.
- Changed the way opening a Personal Chest (block variant) works. It now uses the openGui method.
- Created a wrapper for openGui for opening Gui's for Items
- Created a wrapper for openGui for opening Gui's for Entities (Robit)


thiakil:
> and in fact the client should never be telling the server what it opened 

Agreed

thiakil:
This needs a bit more work, the whole way of opening the GUI is wrong :|

PR #5568, opened 2019-07-16
[WIP] BuildCraft Mj support

pupnewfster:
Partially works, though not in a great state, not sure when I will bother finishing this. Currently it can properly receive power from buildcraft, but has some issues for specific connections where it does not want to send power to buildcraft.

# Known issues
- Quarry stops before finishing forming (when powered by energy cube or energy cube to buildcraft cables. When powered via universal cables it just works very very fast)
- Energy sources cannot output directly to golden kinesis pipes. (This is due to lack of a finished impl on BuildCraft's part: https://github.com/BuildCraft/BuildCraft/blob/8.0.x/common/buildcraft/transport/pipe/flow/PipeFlowPower.java#L173)


pupnewfster:
Forced pushed so that this is up to date with mekanism 9.8.0, but it is unlikely this will be merged any time soon as BuildCraft is planning on entirely rewriting their Mj API; and there are currently implementation bugs in BuildCraft of their API that causes some things to work in ways users would not expect.

pupnewfster:
Closing this as we are not actively working that much on 1.12, and buildcraft in 1.15 is targeted for fabric not forge, so I doubt that their 1.12 version will ever fix the bugs in their API even if we do decide to work on 1.12 ourselves again.

Issue #5571, opened 2019-07-18
No config for stats of atomic destabilize


LordApplesause:
Hello!

I would like a more detailed config option of the atomic destabilizer.

Ex:
- Damage Per Hit
- Enable/Disable modes
- Edit mode stats 

and possibly more that im not thinking of. Thank you!

pupnewfster:
Version [9.8.0](https://www.curseforge.com/minecraft/mc-mods/mekanism/files/2746197) was just released and should sync to the twitch client within the next couple hours.

Issue #5481, opened 2019-05-17
Config options for Atomic Disassembler


DeliciousTea:
The atomic disassembler does not appear to be coded like a regular tool or weapon and has no config options beyond energy use. At the very least, it'd be nice to see config options for damage, as having a multi tool that doesn't break, has high energy efficiency, and is one of the best weapons in the game all at once, is difficult to balance in most mod packs. I've tried adjusting its damage manually with mods like Material Changer and that doesn't even work due to the way it seems to be coded.

TLDR: Please add config options for the atomic disassembler's damage.

Thanks!

pupnewfster:
Version [9.8.0](https://www.curseforge.com/minecraft/mc-mods/mekanism/files/2746197) was just released and should sync to the twitch client within the next couple hours.

PR #5573, opened 2019-07-18
Atomic Disassembler config

pupnewfster:
- Cleaned up the code for `ItemAtomicDisassembler`
- Added config options (#5571, #5481) for min, max damage and energy cost for attacking and using the Disassembler as a hoe (or shovel for grass to grass path).
  - Previously it seemed that there was a value stored for how much power using the disassembler as a hoe should cost, but it did not seem to be used anywhere, so now there is, and it stops early in AOE hoeing if it stops having enough power to continue.
  - Changed how damage works at power levels < 2000 for the atomic disassembler. Previously if the disassembler had 1000 energy remaining it would do full damage (10 hearts) and then drain the disassembler. Now it will do 6 hearts before draining the disassembler.
- Extracted information pertaining to the `Mode` of the disassembler to an enum so that it is easier to add new values/change them. I currently do not have any config options setup to change the efficiencies/range of various modes but that is potentially something we can do in the future.

### DIFF
--- a/src/main/java/mekanism/common/item/ItemAtomicDisassembler.java
+++ b/src/main/java/mekanism/common/item/ItemAtomicDisassembler.java
@@ -88,11 +88,7 @@ public class ItemAtomicDisassembler extends ItemEnergized {
 
     @Override
     public boolean onBlockDestroyed(ItemStack itemstack, World world, IBlockState state, BlockPos pos, EntityLivingBase entityliving) {
-        if (state.getBlockHardness(world, pos) == 0.0D) {
-            setEnergy(itemstack, getEnergy(itemstack) - getDestroyEnergy(itemstack) / 2F);
-        } else {
-            setEnergy(itemstack, getEnergy(itemstack) - getDestroyEnergy(itemstack));
-        }
+        setEnergy(itemstack, getEnergy(itemstack) - getDestroyEnergy(itemstack, state.getBlockHardness(world, pos)));
         return true;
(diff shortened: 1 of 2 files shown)


src/main/java/mekanism/common/item/ItemAtomicDisassembler.java
```diff
@@ -60,41 +60,39 @@ public boolean canHarvestBlock(@Nonnull IBlockState state, ItemStack stack) {
...
-            } else {
-                target.attackEntityFrom(DamageSource.causeMobDamage(attacker), 20);
-            }
-            setEnergy(itemstack, getEnergy(itemstack) - 2000);
+        double energy = getEnergy(itemstack);
+        int energyCost = MekanismConfig.current().general.DISASSEMBLER_WEAPON_USAGE.val();
+        int minDamage = MekanismConfig.current().general.DISASSEMBLER_DAMAGE_MIN.val();
+        int damageDifference = MekanismConfig.current().general.DISASSEMBLER_DAMAGE_MAX.val() - minDamage;
+        //If we don't have enough power use it at a reduced power level
+        double percent = energyCost == 0 ? 1 : energy / energyCost;
```
thiakil: `energy / energyCost` will be > 1 when it has plenty of charge, no?

pupnewfster: Good catch

PR #5577, opened 2019-07-23
a small deduplication of the OreDictManager

ToMe25:
* Fix Sulfur -> Gunpowder Recipe
* Deduplicate the OreDictManager class

### DIFF
--- a/src/main/java/mekanism/common/integration/OreDictManager.java
+++ b/src/main/java/mekanism/common/integration/OreDictManager.java
@@ -399,13 +399,11 @@ public final class OreDictManager {
         }
 
         for (ItemStack dust : OreDictionary.getOres("dustDiamond", false)) {
-            dust = StackUtils.size(dust, 1);
-            InfuseRegistry.registerInfuseObject(dust, new InfuseObject(diamondInfuseType, 10));
+            InfuseRegistry.registerInfuseObject(StackUtils.size(dust, 1), new InfuseObject(diamondInfuseType, 10));
         }
 
         for (ItemStack dust : OreDictionary.getOres("dustTin", false)) {
-            dust = StackUtils.size(dust, 1);
-            InfuseRegistry.registerInfuseObject(dust, new InfuseObject(tinInfuseType, 10));
+            InfuseRegistry.registerInfuseObject(StackUtils.size(dust, 1), new InfuseObject(tinInfuseType, 10));
         }
 
         for (ItemStack sapling : OreDictionary.getOres("treeSapling", false)) {
@@ -415,22 +413,19 @@ public final class OreDictManager {
         }
 
         for (ItemStack coal : OreDictionary.getOres("blockCoal", false)) {
-            coal = StackUtils.size(coal, 1);
-            RecipeHandler.addPRCRecipe(coal, new FluidStack(FluidRegistry.WATER, 1000), new GasStack(MekanismFluids.Oxygen, 1000),
+            RecipeHandler.addPRCRecipe(StackUtils.size(coal, 1), new FluidStack(FluidRegistry.WATER, 1000), new GasStack(MekanismFluids.Oxygen, 1000),
                   new ItemStack(MekanismItems.OtherDust, 9, 3), new GasStack(MekanismFluids.Hydrogen, 1000), 0, 900);
         }
 
         for (ItemStack coal : OreDictionary.getOres("blockCharcoal", false)) {
-            coal = StackUtils.size(coal, 1);
-            RecipeHandler.addPRCRecipe(coal, new FluidStack(FluidRegistry.WATER, 1000), new GasStack(MekanismFluids.Oxygen, 1000),
+            RecipeHandler.addPRCRecipe(StackUtils.size(coal, 1), new FluidStack(FluidRegistry.WATER, 1000), new GasStack(MekanismFluids.Oxygen, 1000),
                   new ItemStack(MekanismItems.OtherDust, 9, 3), new GasStack(MekanismFluids.Hydrogen, 1000), 0, 900);
         }
 
         for (ItemStack sawdust : OreDictionary.getOres("dustWood", false)) {
-            sawdust = StackUtils.size(sawdust, 8);
             //TODO: 1.14 evaluate adding a charcoal dust item to Mekanism, and if so use that instead of charcoal here
-            RecipeHandler.addEnrichmentChamberRecipe(sawdust, new ItemStack(Items.COAL, 1, 1));
-            RecipeHandler.addPRCRecipe(sawdust, new FluidStack(FluidRegistry.WATER, 20), new GasStack(MekanismFluids.Oxygen, 20),
+            RecipeHandler.addEnrichmentChamberRecipe(StackUtils.size(sawdust, 8), new ItemStack(Items.COAL, 1, 1));
+            RecipeHandler.addPRCRecipe(StackUtils.size(sawdust, 1), new FluidStack(FluidRegistry.WATER, 20), new GasStack(MekanismFluids.Oxygen, 20),
                   ItemStack.EMPTY, new GasStack(MekanismFluids.Hydrogen, 20), 0, 30);
         }
     }
--- a/src/main/java/mekanism/common/integration/OreDictManager.java
+++ b/src/main/java/mekanism/common/integration/OreDictManager.java
@@ -50,8 +50,8 @@ public final class OreDictManager {
 
         List<ItemStack> oreDict;
 
-        for (ItemStack plankWood : OreDictionary.getOres("plankWood", false)) {
-            ItemStack plank = StackUtils.size(plankWood, 1);
+        for (ItemStack plank : OreDictionary.getOres("plankWood", false)) {
+            plank = StackUtils.size(plank, 1);
             if (!Recipe.PRECISION_SAWMILL.containsRecipe(plank)) {
                 RecipeHandler.addPrecisionSawmillRecipe(plank, new ItemStack(Items.STICK, 6),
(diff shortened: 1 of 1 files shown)


src/main/java/mekanism/common/integration/OreDictManager.java
```diff
@@ -352,18 +345,31 @@ public static void init(RegistryEvent.Register<IRecipe> event) {
...
+            RecipeHandler.addPRCRecipe(dust, new FluidStack(FluidRegistry.WATER, 100), new GasStack(MekanismFluids.Oxygen, 100),
+                    new ItemStack(MekanismItems.OtherDust, 1, 3), new GasStack(MekanismFluids.Hydrogen, 100), 0, 100);
+        }
 
         oreDict = OreDictionary.getOres("dustSaltpeter", false);
         if (oreDict.size() > 0) {
             ItemStack dustSaltpeter = StackUtils.size(oreDict.get(0), 1);
             RecipeHandler.addCrusherRecipe(new ItemStack(Items.GUNPOWDER), dustSaltpeter);
         }
+        for (ItemStack dust : oreDict) {
```
pupnewfster: This can be moved into the if statement above. I know it *doesn't* really matter when the list is empty because it will just not loop, but if we already "know" there are no entries, there is no point in having it check again.

src/main/java/mekanism/common/integration/OreDictManager.java
```diff
@@ -352,18 +345,31 @@ public static void init(RegistryEvent.Register<IRecipe> event) {
...
+            RecipeHandler.addPRCRecipe(dust, new FluidStack(FluidRegistry.WATER, 100), new GasStack(MekanismFluids.Oxygen, 100),
+                    new ItemStack(MekanismItems.OtherDust, 1, 3), new GasStack(MekanismFluids.Hydrogen, 100), 0, 100);
+        }
 
         oreDict = OreDictionary.getOres("dustCharcoal", false);
         if (oreDict.size() > 0) {
             ItemStack dustCharcoal = StackUtils.size(oreDict.get(0), 1);
             RecipeHandler.addCrusherRecipe(new ItemStack(Items.COAL, 1, 1), dustCharcoal);
         }
+        for (ItemStack dust : oreDict) {
```
pupnewfster: Move inside if statement, see one of my other comments.

src/main/java/mekanism/common/integration/OreDictManager.java
```diff
@@ -352,18 +345,31 @@ public static void init(RegistryEvent.Register<IRecipe> event) {
             ItemStack dustCoal = StackUtils.size(oreDict.get(0), 1);
             RecipeHandler.addCrusherRecipe(new ItemStack(Items.COAL), dustCoal);
         }
+        for (ItemStack dust : oreDict) {
```
pupnewfster: Move inside if statement, see one of my other comments.

src/main/java/mekanism/common/integration/OreDictManager.java
ToMe25: ok ill move that and push it after i got told whether it should use `StackUtils.size(entry, 1)`.

src/main/java/mekanism/common/integration/OreDictManager.java
```diff
@@ -416,20 +418,6 @@ public static void init(RegistryEvent.Register<IRecipe> event) {
...
-                  new ItemStack(MekanismItems.OtherDust, 1, 3), new GasStack(MekanismFluids.Hydrogen, 100), 0, 100);
-        }
-
-        for (ItemStack sulfur : OreDictionary.getOres("sulphur", false)) {
-            RecipeHandler.addEnrichmentChamberRecipe(sulfur, new ItemStack(Items.GUNPOWDER));
-        }
-
         for (ItemStack sawdust : OreDictionary.getOres("dustWood", false)) {
             ItemStack dust = StackUtils.size(sawdust, 8);
             //TODO: 1.14 evaluate adding a charcoal dust item to Mekanism, and if so use that instead of charcoal here
```
ToMe25: Should i change that Enrichment Chamber Recipe to use dustCharcoal instead of Charcoal if available?

ToMe25: that one in the line under this you can't see here because it is unchanged and can't be used for a comment.

ToMe25: and if yes should it be like
````java
    if(oreDict.size() > 0) {
        RecipeHandler.addEnrichmentChamberRecipe(dust, oreDict.get(0));
    } else {
        RecipeHandler.addEnrichmentChamberRecipe(dust, new ItemStack(Items.COAL, 1, 1));
    }
````
or like
````java
    ItemStack charcoal = new ItemStack(Items.COAL, 1, 1);
    if(oreDict.size() > 0) {
        charcoal = oreDict.get(0);
    }
    RecipeHandler.addEnrichmentChamberRecipe(dust, charcoal);
````
?

Comment (thiakil) [reply src/main/java/mekanism/common/integration/OreDictManager.java]: No, leave it as coal for now. Recipes shouldn't really change just because another mod adds an item

**Comment by thiakil** [comment]
> Looks good at a glance, One thing I am unsure about is if we should be having more StackUtils.size(entry, 1) type things for the recipes rather than directly using the stack from the OreDict.

Anything using a stack from the OreDict should be checking for the wildcard meta, probably something this whole class does wrong.

_But_, this will be a non-issue soon enough (hopefully) when I get done on the recipe system rewrite

File: src/main/java/mekanism/common/integration/OreDictManager.java (hunk 4 above)
Comment (ToMe25) [reply src/main/java/mekanism/common/integration/OreDictManager.java]: ok ill let it keep using charcoal.

**Comment by ToMe25** [comment]
So does that mean i leave it like it is now, or should i remove the StackUtils.size(entry, 1), or add them to the remaining positions?

**Comment by thiakil** [comment]
Add it; make sure anything from the OreDictionary gets run through `StackUtils.size`

**Comment by ToMe25** [comment]
ok ill add that and push it.

File: src/main/java/mekanism/common/integration/OreDictManager.java (hunk 1 above)
Comment (ToMe25) [reply src/main/java/mekanism/common/integration/OreDictManager.java]: ok did that.

File: src/main/java/mekanism/common/integration/OreDictManager.java (hunk 2 above)
Comment (ToMe25) [reply src/main/java/mekanism/common/integration/OreDictManager.java]: ok did that.

File: src/main/java/mekanism/common/integration/OreDictManager.java (hunk 3 above)
Comment (ToMe25) [reply src/main/java/mekanism/common/integration/OreDictManager.java]: ok did that.

File: src/main/java/mekanism/common/integration/OreDictManager.java [hunk 5]
```diff
@@ -93,17 +93,20 @@ public static void init(RegistryEvent.Register<IRecipe> event) {
...
 
         for (ItemStack salt : OreDictionary.getOres("dustSalt", false)) {
             RecipeHandler.addChemicalOxidizerRecipe(StackUtils.size(salt, 1), new GasStack(MekanismFluids.Brine, 15));
         }
 
         for (ItemStack dust : OreDictionary.getOres("dustRefinedObsidian", false)) {
-            RecipeHandler.addOsmiumCompressorRecipe(StackUtils.size(dust, 1), new ItemStack(MekanismItems.Ingot, 1, 0));
-            RecipeHandler.addEnrichmentChamberRecipe(StackUtils.size(dust, 1), new ItemStack(MekanismItems.CompressedObsidian));
-            InfuseRegistry.registerInfuseObject(StackUtils.size(dust, 1), new InfuseObject(InfuseRegistry.get("OBSIDIAN"), 10));
+            dust = StackUtils.size(dust, 1);
```
ToMe25: Is it ok like this or should i add a new ItemStack var for the ItemStack StackUtils.size returns?

thiakil: Its fine like this I think, don't stress too much on the minor details, just what you think is "best"

ToMe25: ok than i finished this i guess

src/main/java/mekanism/common/integration/OreDictManager.java
```diff
@@ -407,33 +415,21 @@ public static void init(RegistryEvent.Register<IRecipe> event) {
...
-            RecipeHandler.addEnrichmentChamberRecipe(sulfur, new ItemStack(Items.GUNPOWDER));
-        }
-
         for (ItemStack sawdust : OreDictionary.getOres("dustWood", false)) {
-            ItemStack dust = StackUtils.size(sawdust, 8);
+            sawdust = StackUtils.size(sawdust, 8);
             //TODO: 1.14 evaluate adding a charcoal dust item to Mekanism, and if so use that instead of charcoal here
-            RecipeHandler.addEnrichmentChamberRecipe(dust, new ItemStack(Items.COAL, 1, 1));
+            RecipeHandler.addEnrichmentChamberRecipe(sawdust, new ItemStack(Items.COAL, 1, 1));
             RecipeHandler.addPRCRecipe(sawdust, new FluidStack(FluidRegistry.WATER, 20), new GasStack(MekanismFluids.Oxygen, 20),
```
pupnewfster: This is changing this line from using 1 sawdust to using 8 right?

ToMe25: no. you are right i wasn't thinking about the bottom line :)

ToMe25: ill fix that rn

ToMe25: ok fixed that.

PR #5578, opened 2019-07-23
Cleanup Tools Module code

pupnewfster:
- Made tools extend the various types of tools they are so as to reduce chance of having code deviate across MC versions, as well as, have any mods doing instanceof checks be able to see them.
- Set the repair itemstack for the different ToolMaterial and ArmorMaterial rather than overriding the methods in the various tool classes
- Replace `ToolsItems` class with `ToolsItem` enum that is based on how I am experimenting with doing items in a "flattened" version of Mekanism
- Moved the various ToolMaterial and ArmorMaterial stuff into a Material enum
- Tool smelting to nugget recipes now act the same as vanilla and work on any damage value of the tool/armor, rather than just on the full durability variation.
- Repair recipes for tools and armor now show in JEI in the anvil category


src/main/java/mekanism/tools/common/MekanismTools.java
```diff
@@ -111,172 +77,64 @@ public void init(FMLInitializationEvent event) {
...
-        if (random.nextInt(100) < 50) {
-            setStackIfEmpty(entity, EntityEquipmentSlot.CHEST, new ItemStack(chestplate));
-        }
-        if (random.nextInt(100) < 50) {
-            setStackIfEmpty(entity, EntityEquipmentSlot.LEGS, new ItemStack(leggings));
-        }
-        if (random.nextInt(100) < 50) {
-            setStackIfEmpty(entity, EntityEquipmentSlot.FEET, new ItemStack(boots));
-        }
+    private void setEntityArmorWithChance(Random random, EntityLivingBase entity, ToolsItem sword, ToolsItem helmet, ToolsItem chestplate, ToolsItem leggings, ToolsItem boots) {
```
thiakil: Personally not a fan of this; the separate `if` statements make it clearer that it's a conditional thing, and without, the itemstack gets created whether it's used or not

src/main/java/mekanism/tools/common/ToolsItems.java
```diff
@@ -1,295 +0,0 @@
...
-import mekanism.tools.item.ItemMekanismShovel;
-import mekanism.tools.item.ItemMekanismSword;
-import net.minecraft.inventory.EntityEquipmentSlot;
-import net.minecraft.item.Item;
-import net.minecraft.item.Item.ToolMaterial;
-import net.minecraft.util.ResourceLocation;
-import net.minecraftforge.fml.common.registry.GameRegistry.ObjectHolder;
-import net.minecraftforge.registries.IForgeRegistry;
-
-@ObjectHolder(MekanismTools.MODID)
```
thiakil: As much as I hate `ObjectHolder`, the armour is probably the one case it _should_ be used, on the off chance that someone does a registry replacement on us.

src/main/java/mekanism/tools/item/ItemMekanismPaxel.java
```diff
@@ -1,25 +1,98 @@
...
 
-public class ItemMekanismPaxel extends ItemMekanismTool {
+public class ItemMekanismPaxel extends ItemTool implements IHasRepairType {
 
-    public ItemMekanismPaxel(ToolMaterial toolMaterial) {
-        super(4, -2.4F, toolMaterial, new HashSet<>());
+    private static final Set<Block> EFFECTIVE_ON = new HashSet<>();
+
+    static {
+        EFFECTIVE_ON.addAll(ItemPickaxe.EFFECTIVE_ON);
```
thiakil: Is this this (or the AT entries) necessary? - the only method the field is used in has an override in this class...

pupnewfster: After looking into things a bit more this and the unregistered items are not actually necessary because of forge hooks, and I can remove them if you think I should remove it I can. They just ensure it keeps the same behavior as vanilla if someone directly calls any of the methods.

src/main/java/mekanism/tools/common/ToolsItems.java
pupnewfster: If someone really wants to do a registry replacement on us, I believe they still can with reflection and it is a lot cleaner to just be able to loop over everything. (If you want I could add a method that makes it so they don't *have* to use reflection to change the enum's `Item` reference.

thiakil: The fact is with object holders they dont have to do _anything_ special

src/main/java/mekanism/tools/item/ItemMekanismPaxel.java
thiakil: yeah should be enough to rely on forge hooks

pupnewfster: Left in the one AT addition that is needed so that we don't have to store the value ourselves. (Changing the ItemSword material to protected from private)

PR #5581, opened 2019-07-24
Update Gradle to 4.10.3

pupnewfster:
- Updated gradle from 2.14 to 4.10.3
- Updated various dependencies
- Moved gradle plugins to use the `plugins` block (forge gradle still does not as I could not find the proper maven path for 2.3 of forge gradle. This is a blocker for if we wanted to update to gradle 5.x)

Opening this as a PR mainly because I don't want to accidentally break our build system, but IntelliJ as of 2019.2 no longer supports gradle versions under 3.0.


thiakil:
seems fine, we'll find out next release if it breaks any plugins

PR #5582, opened 2019-07-25
Code cleanup

pupnewfster:
- Cleans up a bit of code after running some of the new IntelliJ 2019.2 inspections.
- Cleanup mismatched lang key entries exposed by minecraft dev plugin

### DIFF
--- a/src/main/java/mekanism/client/gui/robit/GuiRobitRepair.java
+++ b/src/main/java/mekanism/client/gui/robit/GuiRobitRepair.java
@@ -17,7 +17,6 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.network.PacketBuffer;
 import net.minecraft.network.play.client.CPacketCustomPayload;
 import net.minecraft.util.NonNullList;
-import net.minecraft.util.text.translation.I18n;
 import net.minecraftforge.fml.relauncher.Side;
 import net.minecraftforge.fml.relauncher.SideOnly;
 import org.lwjgl.input.Keyboard;
@@ -64,7 +63,7 @@ public class GuiRobitRepair extends GuiRobit implements IContainerListener {
         if (repairContainer.maximumCost > 0) {
             int k = 8453920;
             boolean flag = true;
-            String s = I18n.translateToLocalFormatted("container.repair.cost", repairContainer.maximumCost);
+            String s = LangUtils.localizeWithFormat("container.repair.cost", repairContainer.maximumCost);
 
             if (repairContainer.maximumCost >= 40 && !mc.player.capabilities.isCreativeMode) {
                 s = LangUtils.localize("container.repair.expensive");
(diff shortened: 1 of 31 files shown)


src/main/java/mekanism/client/gui/robit/GuiRobitRepair.java
```diff
@@ -64,7 +63,7 @@ protected void drawGuiContainerForegroundLayer(int mouseX, int mouseY) {
         if (repairContainer.maximumCost > 0) {
             int k = 8453920;
             boolean flag = true;
-            String s = I18n.translateToLocalFormatted("container.repair.cost", repairContainer.maximumCost);
+            String s = LangUtils.localizeWithFormat("container.repair.cost", repairContainer.maximumCost);
```
thiakil: this should use the `I18n` from the client package (there's a bunch of still I still need to do for the lang stuff)

src/main/java/mekanism/common/world/DummyWorld.java
```diff
@@ -1,6 +1,7 @@
 package mekanism.common.world;
 
 import java.io.File;
+import javax.annotation.Nonnull;
```
thiakil: Could probably just use `ParametersAreNonnullByDefault` here as that's what thrown on the package by Forge

src/main/java/mekanism/generators/common/block/BlockGenerator.java
```diff
@@ -243,8 +243,8 @@ public void randomDisplayTick(IBlockState state, World world, BlockPos pos, Rand
             } else if (tileEntity.facing == EnumFacing.EAST) {
                 switch (type) {
                     case HEAT_GENERATOR:
-                        world.spawnParticle(EnumParticleTypes.SMOKE_NORMAL, (double) (xRandom + iRandom), (double) yRandom + 0.5F, (double) (zRandom - jRandom), 0.0D, 0.0D, 0.0D);
-                        world.spawnParticle(EnumParticleTypes.FLAME, (double) (xRandom + iRandom), (double) yRandom + 0.5F, (double) (zRandom - jRandom), 0.0D, 0.0D, 0.0D);
+                        world.spawnParticle(EnumParticleTypes.SMOKE_NORMAL, xRandom + iRandom, (double) yRandom + 0.5F, zRandom - jRandom, 0.0D, 0.0D, 0.0D);
```
thiakil: why do some maintain the cast here?

pupnewfster: Unsure I just let IntelliJ remove the ones it said was unneeded, I can look to see if it works properly still without them when I am next editing this.

src/main/java/mekanism/client/gui/robit/GuiRobitRepair.java
pupnewfster: Is that the same one I removed, or is it something else, as if I remember it had a deprecation warning (though it might have just been before I started looking at I18n methods to replace them). Also I was debating switching to using TextComponentTranslation for the translation stuff to let the client know it should translate it.

thiakil: `net.minecraft.client.resources.I18n` vs `net.minecraft.util.text.translation.I18n`

`util.text` one doesn't exist 1.13+

src/main/java/mekanism/generators/common/block/BlockGenerator.java
pupnewfster: I was able to remove the other ones with it continuing to work fine.

PR #5630, opened 2019-08-29
Added custom GUI's for infusing factories

wezu123:
Dump feature from standard infuser is not present in factory versions, so I made GUI's to them for future use.


thiakil:
This had been added, but in a slightly different way

PR #5654, opened 2019-09-27
update ja_jp.lang

karakufire:
### line 336: unify translation into BuildCraft
* refer to https://github.com/BuildCraft/BuildCraft-Localization/blob/f4f48d826092f57c0b3e5c1186c9f015a12df461/assets/buildcraft/lang/ja_JP.lang#L77

### line 999: make more like to original pronounce
* "Hohlraum" is a borrowed term from German.


thiakil:
Thanks!

PR #5674, opened 2019-11-07
When you walk on ObsidiantTNT, The block become an entity ready to explode

Wvisee:
I added the concept when you walk on an obsidiantTNT, the block launch ready to explode.


thiakil:
Thanks, but no thanks

thiakil:
Thanks, but no thanks

PR #5677, opened 2019-11-12
Flamethrower can make ObsidianTNT explode.

Wvisee:
## Flamethrower can make ObsidianTNT explode:
I've seen that flamethrower couldn't burn ObsidianTNT blocks. So I made the ObsidianTNT able to burn. When the ObsidianTNT burn, like the TNT, it will explode.
I did this feature because it seems right to make new items work together.


pupnewfster:
Closing this as we are not actively developing 1.12 and I implemented this in a "better" way in the 1.15 branch a while back.

Issue #5688, opened 2019-12-03
ThreadMinerSearch does not use a ChunkCache to scan the world, resulting in off thread world access / CME's


clienthax:
*Please use the search functionality before reporting an issue. Also take a look at the closed issues!*

#### Issue description:
ThreadMinerSearch does not use a ChunkCache to scan the world, resulting in off thread world access / CME's
This can cause clients around the miner to disconnect too occasionally.

#### Version (make sure you are on the latest version before reporting):
**Forge:** forge-1.12.2-14.23.5.2847-universal.jar
**Mekanism:** Mekanism-1.12.2-9.8.2.389.jar
**Other relevant version:** 

#### If a (crash)log is relevant for this issue, link it here: (_It's almost always relevant_)

```
Exception in thread "Thread-88" java.lang.IllegalStateException: CauseStackManager called from off main thread (current='ThreadMinerSearch{class=class mekanism.common.content.miner.ThreadMinerSearch, name=Thread-88, priority=5, group=net.minecraftforge.fml.common.thread.SidedThreadGroup[name=SERVER,maxpri=10]}', expected='Thread{class=class java.lang.Thread, name=Server thread, priority=5, group=net.minecraftforge.fml.common.thread.SidedThreadGroup[name=SERVER,maxpri=10]}')!
at org.spongepowered.common.event.SpongeCauseStackManager.enforceMainThread(SpongeCauseStackManager.java:121)
at org.spongepowered.common.event.SpongeCauseStackManager.getCurrentCause(SpongeCauseStackManager.java:136)
at net.minecraft.world.chunk.Chunk.handler$impl$UpdateNeighborsOnLoad$zma000(Chunk.java:3043)
at net.minecraft.world.chunk.Chunk.func_76631_c(Chunk.java:863)
at net.minecraftforge.common.chunkio.ChunkIOProvider.syncCallback(ChunkIOProvider.java:109)
at net.minecraftforge.common.chunkio.ChunkIOExecutor.syncChunkLoad(ChunkIOExecutor.java:94)
at net.minecraft.world.gen.ChunkProviderServer.loadChunk(ChunkProviderServer.java:118)
at net.minecraft.world.gen.ChunkProviderServer.func_186028_c(ChunkProviderServer.java:89)
at net.minecraft.world.gen.ChunkProviderServer.redirect$impl$ProvideChunkForced$zmd000(ChunkProviderServer.java:1158)
[...]

PR #5689, opened 2019-12-03
[1.14.X] Make ThreadMinerSearch use a chunk cache.

clienthax:
Make ThreadMinerSearch use a chunkcache to prevent possible CME's / other issues with non safe world access.

https://github.com/mekanism/Mekanism/issues/5688


pupnewfster:
Code wise it seems mostly fine, I will test it later, but should there be a check at the start of `run` to make sure that the `chunkCache` is not null. (Also don't worry about the check having failed I think that might have to do with the main branch being 1.12, and so it can't evaluate the 1.14 branch or something)

clienthax:
If the chunkcache isnt created it will never run in the first place @pupnewfster as it would crap out before the .start is called

PR #5692, opened 2019-12-07
[1.12.2] Make ThreadMinerSearch use a chunk cache.

clienthax:
Make ThreadMinerSearch use a chunkcache to prevent possible CME's / other issues with non safe world access.

I have had to move the breakable check onto the main thread in canBreak, as 1.12.2 doesn't have the functionality to check this from the cache.


thiakil:
Thanks, sorry it took so long to get to it xD

PR #5737, opened 2020-01-09
merge with fr_fr_v2

Redstylt:


src/tools/resources/assets/mekanismtools/lang/fr_fr.json
```diff
@@ -64,4 +64,5 @@
   "item.mekanismtools.steel_sword": "Épée en acier",
   "item.mekanismtools.stone_paxel": "Piopèche en pierre",
   "item.mekanismtools.wood_paxel": "Piopèche en bois"
```
pupnewfster: You need a comma at the end of this line so that the next line is valid. Only the final entry in a json file doesn't (and shouldn't) need a comma.

PR #5741, opened 2020-01-09
crash fix

ArnoMyOnlinePresence:
Hey guys

there was a missed null check that seems to crash the entire pack when disconnecting from a server and trying to reconnect immediately again without restarting

### DIFF
--- a/src/main/java/mekanism/tools/common/ToolsItem.java
+++ b/src/main/java/mekanism/tools/common/ToolsItem.java
@@ -160,8 +160,12 @@ public enum ToolsItem {
 
     public static void remapItems() {
         for (ToolsItem toolsItem : values()) {
-            ResourceLocation registryName = toolsItem.getItem().getRegistryName();
-            toolsItem.updateItem(ForgeRegistries.ITEMS.getValue(registryName));
+            var toolItem = toolsItem.getItem();
+            
+            if (toolItem != null) {
+                ResourceLocation registryName = .getRegistryName();
+                toolsItem.updateItem(ForgeRegistries.ITEMS.getValue(registryName));
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


thiakil:
It should never be null, you either found a bug in forge or have a bad mod in the pack...

Also your proposed code isn't valid Java

PurpleIsEverything:
I'll make a proper PR for this tonight if it's not already fixed by then.

This is very much a thing and it can be null, it's an edge case that can be caused by a number of things including Join Game transition packets, login delays, server lag.

This crash can be caused without any other mods if the conditions are right just it's exacerbated by the inclusion of additional mods, lag and delays.

PurpleIsEverything:
The technically explanation of what's going on here is a bit more complicated.

It comes down to rapid GameData state changes in the registry causing frozen and unfrozen states to finish with incomplete data and then calling a remap event. This is a side effect of all the registry changes in 1.12 and have since been changed in 1.13+

Basically the reason Mekanism is crashing due to this is because it's not surviving long enough for a proper registry sync to complete and proper and complete registry to be rebuilt. The fix is super simple just checking if getItemStack().isEmpty() will do the trick, subsequent remap events with complete registry data will allow things to function as normal.

pupnewfster:
Not looking too much into either of these PRs personally right now due to us also having to figure out some stuff about our build system/how we are going to handle old versions. But for now I am closing this PR in favor of #5797, as based on your explanation that does seem like a more proper fix (and at the very least is valid java code)

PurpleIsEverything:
Appreciated, It's a frustrating issue we've been dealing with repeatedly we had to create a small patch with these changes to help those experiencing it.

We hope to see that PR merged soon and appreciate your consideration.

Issue #5767, opened 2020-01-12
Atomic Disassembler extended vein mining


Discomanco:
#### Issue description:
When mining something that there's a lot of, like stone, in Extended Vein Mining, it will create odd shapes, instead of something more.. predictable. It has a tendency of preferring to mine north, rather than around the player, or the players facing direction. It mines just below 2 stacks before stopping.
A bit kin to the problem there was the Fluidic Plenisher (#4490) a long time ago

Examples of mining: https://gyazo.com/7af0e27f936a5406c9c29ed279ef34dd
![2020-01-12_23 26 35](https://user-images.githubusercontent.com/4707702/72226682-1e6b4580-3594-11ea-876a-6a097be2d42b.png)

#### Steps to reproduce:
1. Be in Survival
2. Put Atomic Disassembler into Extended Vein Mining
3. Start mining stone deep under ground (or somewhere else with a lot of stone)

#### Version (make sure you are on the latest version before reporting):
**Forge:** 30.0.36
**Mekanism:** 9.9.1
**Other relevant version:**

halvors:
Also, seems like normal vein miner is not working anymore...

pupnewfster:
That is an entirely different issue which you can open an issue up for if you want. But I don't believe I ever got around to porting the atomic disassembler to use tags instead of attempting to use the ore dict (which no longer exists)

halvors:
@pupnewfster Made this instead :) #5852

pupnewfster:
Thank you for this PR, however I am going to close it as I have been working on this some already today and the proposed solution in the linked PR is not the correct way to handle tags, as it is more of a patch than an actual fix. It should be checking block tags (and in fact the whole pick block thing) no longer is needed, as can be seen here: https://github.com/mekanism/Mekanism/commit/8887e89f3f566a046bf3e2b6d48f50a19f873497

I still have more things I plan on working on for the atomic disassembler before releasing 9.9.9, such as fixing this issue for the extended vein mining shape. I also want to fix and improve upon the various right click interactions to use the proper lookup maps and also support log stripping.

pupnewfster:
Closing this as it is fixed in 9.9.9, which is now released and on curseforge.

halvors:
np :-)

PrzemoVi:
What about 1.12.2?
I`m experiencing same problems with 9.8.3.390.

pupnewfster:
1.12.2 is no longer receiving active development.

PR #5852, opened 2020-02-05
Changed ItemAtomicDisassembler to use tags instead of OreDictionary.

halvors:
Implements tags for the Atomic Disassembler.  This fixes vein miner, so that it works on all ores and logs.

Since Ore Dictionary string matching is no longer an option i thought it made sense to have this as a private function to the ItemAtomicDisassembler class.


pupnewfster:
See https://github.com/mekanism/Mekanism/issues/5767#issuecomment-582621154 for details to why I am closing this.

PR #5991, opened 2020-03-26
Changed to use unsigned long

Chiefwright:
Class now uses an unsigned long for value portion, updated the logic for overflows to reflect this.


edkek:
I think `testBasicDivision` should be a property based test. Testing the following property should be good

```java
FloatLong a = ....
FloatLong b = ....
FloatLong c = a.divide(b);

FloatLong d = c.multiply(b);

Assertions.assertEquals(c, d);
```

And then have `QuickTheroy` generate values for `a` and `b`

PR #6235, opened 2020-06-23
Fix players with GravitationalModulation MekaSuit falling in spectator mode.

ToMe25:
 * Fix players with a MekaSuit Bodyarmor with a GravitationalModulationUnit enabled falling in spectator mode.

### DIFF
--- a/src/main/java/mekanism/client/ClientTickHandler.java
+++ b/src/main/java/mekanism/client/ClientTickHandler.java
@@ -80,7 +80,7 @@ public class ClientTickHandler {
         if (player != minecraft.player) {
             return Mekanism.playerState.isJetpackOn(player);
         }
-        if (!player.isCreative() && !player.isSpectator()) {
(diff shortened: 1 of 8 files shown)


pupnewfster:
I think it is getting to the point it may almost make more sense to add a method to `MekanismUtils` that checks creative or spectator status. (Or not both of them, not sure which way it would make more sense for the method to be declared). Given doing a quick search for `.isCreative()` shows a couple other locations that could probably also use check against the player being in spectator. (Mainly pertaining to the fact that spectators probably should not be affected by radiation).

ToMe25:
yes, that sounds like a good idea, but how should that be named?
isGod or canFly are the most reasonable name i could come up with.

pupnewfster:
Not sure to be honest, could also have the default be inverted and have it be something like `isPlayingGamemode`

ToMe25:
ok will do that.

PR #6241, opened 2020-06-26
Update Wood Gasification to atleast for logs and dust return coal dust

ToMe25:
 - Update the log and dust wood gasification recipes to use the equivalent of two logs, and return one charcoal dust.
 - Add stair wood gasification recipes
 - rebalance plank, slab and stick wood gasification recipes, to not result in more then the wood they are made of
 - move coal gasification recipes to their own directory

## Questions:
 - Should the dust recipe require more sawdust? you can get up to 37.75 sawdust from one log, while currently one sawdust is worth 1/8 of a log.
 - should more recipes be increased in size to the equivalent of 2 logs and result in a charcoal dust?
 - should the plank, slab and stick recipes even be rebalanced, or were they fine before because you need energy to convert the wood anyways?


pupnewfster:
I think it probably does make sense to balance them closer to the actual values, if a log is worth roughly 37.75 sawdust it probably makes sense to have it use either four or five (I will leave this up to you as you seem to know move about how gasification works in real life).

ToMe25:
I don't know that much, but some time back, shortly before i implemented it originally i read alot about it, because it interested me at the time.
Also i wouldn't make it 36 sawdust per log value, because that is the absolute max you can get and requires a few sawmills.
I would say 32 sawdust seems fair, also that would allow the input for alteast 2 recipes.
Also i would use 4 logs then, that should make it easier to handle, but this would mean i can't make the dust recipe return any charcoal dust, if each log is worth 32 sawdust.
So should it be 
a) 4 logs per charcoal and nothing else returns any charcoal dust,
b) 2 logs per charcoal and the dust recipe returns a charcoal dust,
c) 4 logs per charcoal and the plank recipe returns charcoal instead of the dust one

ToMe25:
This implements c.
Also i missed something, so the PR isn't finished :)

ToMe25:
This PR should now be finished.

PR #6248, opened 2020-06-27
Fix armor items that shouldn't be enchantable being it anyways

ToMe25:
 - Fix some armor items being enchantable that shouldn't be.


pupnewfster:
I believe all of these could be: `return material.getEnchantability() > 0 && super.isBookEnchantable(stack, book);`

ToMe25:
i noticed just after publishing it could also just be `return material.getEnchantability() > 0;`, i will update that later.

ToMe25:
done.

ToMe25:
btw i suggest adding a Thorns Unit as a replacement for the Thorns enchantment on the MekaSuit.

PR #6249, opened 2020-06-27
Make armored jetpack enchantable

ToMe25:
 * Make the Armored Jetpack enchantable.
This only changes anything after #6248 and is meant as a suggestion, to make the armored jetpack an exception being enchantable.


pupnewfster:
I think at least for now it makes more sense to keep it unenchantable, and is something we can potentially re-evaluate further down the line.

PR #6250, opened 2020-06-28
Make Flamethrower flames be always faster then the player causing them

ToMe25:
 - Add player motion to the flame motion.

This PR currently fails to make the flames faster then then player, if the player is very fast, this is caused by how the vanilla method for shooting projectiles works.


ToMe25:
This is the issue i was talking about:

ToMe25:
This visual bug should be fixed now, but now if the player is very fast, the flames are slower then the player.

ToMe25:
This is ready now, the only problem is, [this](https://github.com/mekanism/Mekanism/pull/6250/files#diff-0b7ead56e190a02a9946246bf8b39714R78) line contains three floats,
the second of them(its 0.5 atm) needs to be ~0.5 for the flames to keep the speed they have without this PR, but needs to be >1 to guarantee the flames are always faster then the player.
Because after a day of testing i don't even know how that could be fixed in theory, and its still way better then before, i see this as finished now.

PR #6257, opened 2020-06-30
Add Transparent Plastic Blocks, Stairs and Slabs

ToMe25:
 * Add Transparent Plastic Blocks, Stairs and Slabs.
 * Add Glow Plastic Stairs and Slabs.
 * Change the Plastic Block recipe to use 4 plastic Sheets instead of 8, while still resulting in 4 blocks, because making it result in 2 blocks would make creating Plastic Blocks require more dye then recoloring them.
 * Move Plastic Item Models and Blockstates to datagen.


ToMe25:
This should work in 1.16 now.

ToMe25:
This new error isn't even in a file i changed at all, so i am pretty sure its not my fault.
Also i tested this, and it seemed to work.

ToMe25:
btw why are there two changed files less since i rebased this to 1.16.1?

pupnewfster:
You could try rerunning the data generator, it is possible that either something didn't rebase correctly, or it was just a line ending change that isn't a change anymore now that I have started running the data generator again in 1.16 itself.

ToMe25:
i did rerun the data generator before the last commit, and that didn't cause any file changes at all.

PR #6260, opened 2020-07-02
Fix Electrolytic Breathing Unit working without water

ToMe25:
 * Fix Electrolytic Breathing Unit working without water.
This currently only requires your feet to be in water tho.


aidancbrady:
Thanks!

Issue #6264, opened 2020-07-04
Owner in the machine tooltip is 'null'


Fukutu:
#### Issue description:

**This doesn't affect security.**
Reopening the world, the problem is solved.

![image](https://user-images.githubusercontent.com/60327358/86502785-6a49bb00-bd7d-11ea-9b00-a067bc4c1c8b.png)

#### Steps to reproduce:
1. You **MUST** create a new world
2. Place any machine
3. Dismantle the machine
4. Take a look at the machine tooltip 

#### Version:
**Forge: 31.2.30** 
**Mekanism: 9.10.8**

aidancbrady:
Now fixed in v10 with 456f752.

PR #6267, opened 2020-07-05
Fix desync between client and server in MekanismClient

Fukutu:
See #6264

Currently, sync  is being done only at login

https://github.com/mekanism/Mekanism/blob/e2fbde5c89f8b3bdcf9e406932436711974d3652/src/main/java/mekanism/common/CommonPlayerTracker.java#L42-L50

https://github.com/mekanism/Mekanism/blob/e2fbde5c89f8b3bdcf9e406932436711974d3652/src/main/java/mekanism/common/network/PacketSecurityUpdate.java#L59-L80

**I'm not sure if this is the *'best'* way to do this**


aidancbrady:
456f752 is a cleaner fix. I'll back port to 1.15.x but I'm not sure we'll be pushing another 1.15 release.

PR #6296, opened 2020-07-11
Refreshed ore textures

Laz-The-Artist:
Hello, i have updated the ore blocks to use the post-1.14 stone texture, to fit into the world, and into vanilla much better.
Please use them, it can only make this mod better, and more fitting into the new version.
Im an artist, and i mostly work with vanilla-style (aka jappa-style) textures.

Updated ore block textures:
-copper ore
-lead ore
-osmium ore
-tin ore
-uranium ore
-fluorite ore


King-of-Creepers:
Soon, i plan to make the ingots as well. in the last update (V10) CyanideX said he wants the mod to get closer to vanilla, moving from x32 to x16 was a good step, but since textures changed in 1.14, i think the ores and ingots are necessary to keep the mod consistent with vanilla

CyanideX:
I think the ore changes look great!

At the moment, I don't think the ingots are a necessary change. When I said that I wanted to move the mod closer to vanilla, I was speaking strictly about texture resolution. I do not plan to have all of the resources match vanilla texture style (although I'm not opposed to minor changes to bring them closer) because the block textures are already quite different from vanilla blocks.

If you're up for the challenge, you could always make an attempt to recreate all the Mekanism assets in vanilla style! 😛 I would be interested in seeing how that would look.

CyanideX:
On a side note, you should look into a program like PNGGauntlet to index your pngs. This saves a ton of space by reducing the file sizes. Currently, some of your PR textures are 1000% of the original size. 😉

Laz-The-Artist:
> On a side note, you should look into a program like PNGGauntlet to index your pngs. This saves a ton of space by reducing the file sizes. Currently, some of your PR textures are 1000% of the original size. 😉

i am using Adobe Photoshop for creating and editing textures, and it seems i forgot to export them correctly, i might have forgot to apply the right settings

Laz-The-Artist:
> I think the ore changes look great!
> 
> At the moment, I don't think the ingots are a necessary change. When I said that I wanted to move the mod closer to vanilla, I was speaking strictly about texture resolution. I do not plan to have all of the resources match vanilla texture style (although I'm not opposed to minor changes to bring them closer) because the block textures are already quite different from vanilla blocks.
> 
> If you're up for the challenge, you could always make an attempt to recreate all the Mekanism assets in vanilla style! 😛 I would be interested in seeing how that would look.

Nice to hear that! i havent planned on making every asset, just the ingots and nuggets, because these were the most significant changes, that in my opinion every mod maker should follow. (specially the machines would look interesting, because there are no such complex thing in the vanilla :P ) but i dont say i wont make every assets ever, at some point sure i will :D
the blocks shouldnt change, they are unique as how they look currently.

CyanideX:
I use Photoshop as well but I find it easier to just export the PNG and drag them all into PNGGauntlet after the fact. It's far too much effort to compress the PNG in the application. 😅 

Regarding the asset changes you suggested, I already have plans to update the nuggets to match our new uranium and copper nugget style which is close enough to vanilla. The ingots won't be adopting the vanilla texture style but I would be willing to meet halfway with something like this:

image

Laz-The-Artist:
> I use Photoshop as well but I find it easier to just export the PNG and drag them all into PNGGauntlet after the fact. It's far too much effort to compress the PNG in the application. 😅
> 
> Regarding the asset changes you suggested, I already have plans to update the nuggets to match our new uranium and copper nugget style which is close enough to vanilla. The ingots won't be adopting the vanilla texture style but I would be willing to meet halfway with something like this:
> 
> image

a team of my friend (which i am in, ofc) are making the new assets for tinkers construct 3, and we had to create a ton of materials which most of them are ingots/metals, so i have experience in this kind of things, and this halfway you mention here is basically what i would done, and similar things has been done there as well. perfect work :D
Keepeing in mind that these ingots have some sort of noise going on, it was necessary to keep it for staying original, so cutting the color number in half is the way, because as you presented it is indeed possible. there are some minor changes that would be good to make, like where outline meets the horizontal shine on the bottom part and right part, and making the outline a tiny bit darker and adopting the second shade for the outline from the vanilla ingot.

nice to see it all coming together, i wish the next minor update will contain your new ingot design and my ores as well :3

PR #6319, opened 2020-07-13
Implement Teleporter portal coloring

ToMe25:
Not yet working with portable teleporters
Also the colored square isn't auto updating yet.

 - Make it possible to color teleporters frequencies

### DIFF
--- a/src/main/java/mekanism/client/gui/GuiTeleporter.java
+++ b/src/main/java/mekanism/client/gui/GuiTeleporter.java
@@ -92,7 +92,7 @@ public class GuiTeleporter extends GuiMekanismTile<TileEntityTeleporter, Mekanis
             }
             updateButtons();
         }));
-        func_230480_a_(new GuiSlot(SlotType.NORMAL, this, 131, 120));
+        func_230480_a_(new GuiSlot(SlotType.NORMAL, this, 131, 120).setRenderAboveSlots());
         func_230480_a_(new ColorButton(this, getGuiLeft() + 132, getGuiTop() + 121, 16, 16,
               () -> tile.getColor(),
               () -> sendColorUpdate(0),
--- a/src/main/java/mekanism/client/gui/GuiTeleporter.java
+++ b/src/main/java/mekanism/client/gui/GuiTeleporter.java
@@ -94,8 +94,7 @@ public class GuiTeleporter extends GuiMekanismTile<TileEntityTeleporter, Mekanis
         }));
         func_230480_a_(new GuiSlot(SlotType.NORMAL, this, 131, 120).setRenderAboveSlots());
         func_230480_a_(new ColorButton(this, getGuiLeft() + 132, getGuiTop() + 121, 16, 16,
-              //TODO: Replace this with making the SyncableFrequency able to have a custom "compare/hash" check so we can mark it as dirty when the color changes
-              tile::getColor,
+              () -> getFrequency() == null ? null : getFrequency().getColor(),
               () -> sendColorUpdate(0),
               () -> sendColorUpdate(1)));
         func_230480_a_(frequencyField = new GuiTextField(this, 50, 103, 98, 11));
--- a/src/main/java/mekanism/client/gui/GuiTeleporter.java
+++ b/src/main/java/mekanism/client/gui/GuiTeleporter.java
@@ -7,10 +7,13 @@ import java.util.UUID;
 import javax.annotation.Nonnull;
 import mekanism.api.text.EnumColor;
 import mekanism.client.gui.element.bar.GuiVerticalPowerBar;
+import mekanism.client.gui.element.button.ColorButton;
 import mekanism.client.gui.element.button.MekanismButton;
 import mekanism.client.gui.element.button.TranslationButton;
 import mekanism.client.gui.element.custom.GuiTeleporterStatus;
 import mekanism.client.gui.element.scroll.GuiTextScrollList;
+import mekanism.client.gui.element.slot.GuiSlot;
+import mekanism.client.gui.element.slot.SlotType;
 import mekanism.client.gui.element.tab.GuiRedstoneControlTab;
 import mekanism.client.gui.element.tab.GuiSecurityTab;
 import mekanism.client.gui.element.tab.GuiUpgradeTab;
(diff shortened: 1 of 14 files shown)


src/main/java/mekanism/client/gui/GuiTeleporter.java
```diff
@@ -137,6 +149,11 @@ private void updateButtons() {
     @Override
     public void func_231023_e_() {
         super.func_231023_e_();
+        TeleporterFrequency freq = getFrequency();
+        if (!init && getFrequency() != null) {
+            init = true;
+            privateMode = freq.isPrivate();
+        }
```
pupnewfster, line 156: Is there a reason this is being done, it seems kind of pointless given it already gets set in the constructor.

src/main/java/mekanism/client/gui/GuiTeleporter.java
```diff
@@ -85,6 +92,11 @@ public void func_231160_c_() {
             }
             updateButtons();
         }));
+        func_230480_a_(new GuiSlot(SlotType.NORMAL, this, 131, 120));
```
pupnewfster, line 95: This should have a `.setRenderAboveSlots()` on the end of the slot creation like the logistical sorter does. (If it doesn't in your code base just pull in/rebase onto the latest 1.16 branch to see what I mean)

src/main/java/mekanism/client/gui/item/GuiPortableTeleporter.java
```diff
@@ -117,6 +121,11 @@ public double getLevel() {
             }
             updateButtons();
         }));
+        func_230480_a_(new GuiSlot(SlotType.NORMAL, this, 131, 120));
```
pupnewfster, line 124: This should have a `.setRenderAboveSlots()` on the end of the slot creation like the logistical sorter does. (If it doesn't in your code base just pull in/rebase onto the latest 1.16 branch to see what I mean)

src/main/java/mekanism/common/content/teleporter/TeleporterFrequency.java
```diff
@@ -57,4 +71,28 @@ public Coord4D getClosestCoords(Coord4D coord) {
...
+    @Override
+    protected void read(CompoundNBT nbtTags) {
+        super.read(nbtTags);
+        NBTUtils.setEnumIfPresent(nbtTags, NBTConstants.COLOR, EnumColor::byIndexStatic, (value) -> color = value);
+    }
+
+    @Override
+    protected void read(PacketBuffer dataStream) {
+        super.read(dataStream);
+        color = EnumColor.byIndexStatic(dataStream.readVarInt());
```
pupnewfster, line 84: I believe this should be able to be `color = dataStream.readEnum(EnumColor.class);`

src/main/java/mekanism/common/content/teleporter/TeleporterFrequency.java
```diff
@@ -57,4 +71,28 @@ public Coord4D getClosestCoords(Coord4D coord) {
...
+    @Override
+    public void write(CompoundNBT nbtTags) {
+        super.write(nbtTags);
+        nbtTags.putInt(NBTConstants.COLOR, color.ordinal());
+    }
+
+    @Override
+    public void write(PacketBuffer buffer) {
+        super.write(buffer);
+        buffer.writeVarInt(color.ordinal());
```
pupnewfster, line 96: This can be `buffer.writeEnumValue(color);`

src/main/java/mekanism/common/tile/TileEntityTeleporter.java
```diff
@@ -58,6 +60,7 @@
     @Nullable
     private Direction frameDirection;
     private boolean frameRotated;
+    private EnumColor color;
```
pupnewfster, line 63: I am somewhat unsure of having the color also be stored in the tile as well as in the frequency. I am not quite sure on the best solution though so may look into it some myself when I am testing this PR and maybe try to move some of it into the frequency component (or maybe a subclass not sure yet).

src/main/java/mekanism/client/gui/GuiTeleporter.java
ToMe25: Two actually:
1. its done in the QIO too,
2. It fixes some bug, i don't quite remember which one, or whether something else fixes that too tho.

src/main/java/mekanism/client/gui/GuiTeleporter.java
ToMe25: This was just copied from the QIO, but i will look into that.

src/main/java/mekanism/common/content/teleporter/TeleporterFrequency.java
ToMe25: I can change it to that, but i just stole this from the qio.

src/main/java/mekanism/common/content/teleporter/TeleporterFrequency.java
ToMe25: ok, should i change it in the qio too?

src/main/java/mekanism/common/tile/TileEntityTeleporter.java
ToMe25: It is like this in the qio too.

src/main/java/mekanism/client/gui/GuiTeleporter.java
pupnewfster: It may not be strictly needed due to there being no "background" for the slots, but probably is worth doing anyways. (May make sense to adjust the QIO ones to be like it as well).

src/main/java/mekanism/common/content/teleporter/TeleporterFrequency.java
pupnewfster: Yes I think that would be a good idea as it makes it more readable overall.

src/main/java/mekanism/common/tile/TileEntityTeleporter.java
pupnewfster: I will evaluate both at once then, given having it have a more proper home in the frequency component may make sense then.

src/main/java/mekanism/client/gui/GuiTeleporter.java
pupnewfster: Seems odd but fair enough.

src/main/java/mekanism/client/gui/GuiTeleporter.java
ToMe25: done.

src/main/java/mekanism/common/content/teleporter/TeleporterFrequency.java
ToMe25: done.

src/main/java/mekanism/common/content/teleporter/TeleporterFrequency.java
ToMe25: done.

src/main/java/mekanism/client/gui/item/GuiPortableTeleporter.java
ToMe25: done.

PR #6345, opened 2020-07-17
Improve /mek build structures

ToMe25:
 - Turbine: Replace some Electromagnetic Coils with Saturating Condensers.
This makes the turbine able to condense all the steam without decreasing energy production at all.
 - Boiler: Remove second layer of Superheating Elements and move Pressure Dispersers up.
This greatly increases max achievable boil rate, by increasing the water capacity.
 - Fusion Reactor: Remove some useless Casings.
This just makes it look more how most people build it.
 - Added SPS: Builds an SPS without any ports or Supercharged Coils.
 - Fission Reactor: Fill empty columns with air.
This makes it so you no longer have to clean these single columns if you built it into something.

### DIFF
--- a/src/generators/java/mekanism/generators/common/registries/GeneratorsBuilders.java
+++ b/src/generators/java/mekanism/generators/common/registries/GeneratorsBuilders.java
@@ -81,8 +81,8 @@ public class GeneratorsBuilders {
 
         @Override
         protected void build(World world, BlockPos start) {
-            buildWalls(world, start);
             buildPartialFrame(world, start, 1);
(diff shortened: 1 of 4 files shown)


pupnewfster:
I will look at this a bit later, but if you wouldn't mind also adding support for `/mek build sps` if I recall correctly Aidan originally left it out to add a bit of "mystery" and not fully tease all the features. Now that we have released there is no reason not to include it as one of the possible things to build using the command.

ToMe25:
I will look into that, but if its too complex to auto build this way, i wont add it.

PR #6346, opened 2020-07-17
updated mappings

lehjr:
updated mappings to 20200707-1.16.1


pupnewfster:
We are not going to update mappings until K9 (forge bot) supports querying them in both directions, as until it does they are more of a pain than they are worth.

PR #6353, opened 2020-07-19
Added 10x ore processing fueled by antimatter

Witherking25:
10x ore processing by chemical injecting antimatter into ore, doubling it before chemical dissolution chamber


Witherking25:
feel free to rename it from ender
im garbage at naming things

pupnewfster:
We discussed it some and for now we are going to close this as we don't see the point in this being a thing. While personally I don't see a problem with eventually maybe adding more tiers of ore processing, I do not think this is the way to go about it just going from 5x->10x in a single step.

Do note however, that there is nothing stopping you from adding these recipes for your own modpack or making an addon if you really want to play with these recipes.

PR #6382, opened 2020-07-23
Add vein mine and attack damage unit info to HUD

ToMe25:
 - Add enabled/disabled status of extended vein mining from the Vein Mining Unit to the HUD
 - Add damage of the Attack Amplification Unit to the HUD


pupnewfster:
This causes some issues with how many options are being added to the GUI, Eventually we may want to make the config side be able to scroll, but I am not sure if that is currently how we should go about this, or if this PR is just adding too many different options to the GUI.
image

PR #6383, opened 2020-07-23
Make Meka-Tool and Armor lava proof

ToMe25:
 - Make Meka-Tool and Mekasuit fire & lava proof
They make you immune to fire, so why should they burn?

### DIFF
--- a/src/main/java/mekanism/common/registration/impl/ItemDeferredRegister.java
+++ b/src/main/java/mekanism/common/registration/impl/ItemDeferredRegister.java
@@ -33,7 +33,7 @@ public class ItemDeferredRegister extends WrappedDeferredRegister<Item> {
     }
 
     public ItemRegistryObject<Item> registerUnburnable(String name) {
-        return register(name, properties -> new Item(properties.func_234689_a_()));
+        return registerUnburnable(name, Item::new);
     }
 
     public ItemRegistryObject<Item> register(String name, Rarity rarity) {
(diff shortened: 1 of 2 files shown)


src/main/java/mekanism/common/registration/impl/ItemDeferredRegister.java
```diff
@@ -54,6 +54,10 @@ public ITextComponent getDisplayName(@Nonnull ItemStack stack) {
         return register(name, () -> sup.apply(getMekBaseProperties()));
     }
 
+    public <ITEM extends Item> ItemRegistryObject<ITEM> registerUnburnable(String name, Function<Item.Properties, ITEM> sup) {
```
pupnewfster, line 57: The registerUnburnable method that already exists should probably be adjusted to just being `return register(name, Item::new);`

ToMe25: do you mean `return registerUnburnable(name, Item::new);` or did i miss something?

pupnewfster: Yes, sorry I copy pasted and only adjusted the supplier -> method reference forgot to update the method it was calling.

ToMe25: done.

PR #6422, opened 2020-08-04
Replace Atomic Disassembler mode switch with radial selector.

sctprog:
As title says. 

There is an issue with most of the images used showing as black but retaining the intended shape. I do not know why but would be happy to fix, once I understand what is happening.

Available as Aelith on discord.

icons


src/main/java/mekanism/common/item/gear/ItemAtomicDisassembler.java
```diff
@@ -314,18 +349,40 @@ public DisassemblerMode byIndex(int index) {
             return MathUtils.getByIndexMod(MODES, index);
         }
 
+        // This is kinda hacky. A couple duplicates here to allow the different interfaces to play nicely.
```
pupnewfster, line 352: The `IHasTranslationKey` can just be removed it is for use in TextComponentUtils, IHasTextComponent is a more specific "implementation" of it (unless anything is specifically querying the translation key otherwise.

src/main/java/mekanism/common/item/gear/ItemAtomicDisassembler.java
```diff
@@ -314,18 +349,40 @@ public DisassemblerMode byIndex(int index) {
...
+            return this.getTextComponent();
+        }
+
+        // Efficiency of block destruction
         public int getEfficiency() {
             return efficiency;
         }
 
-        @Override
+        // The old interface required an override here. Left it here so that this would be easier to revert if necessary.
```
pupnewfster, line 374: This isn't good, we need to make a "disableable" version of the radial mode system as the atomic disassembler allows for disabling specific modes on the atomic disassembler.

sctprog:
Newest commit now follows the settings set in the config file.
The png issues have also been addressed. My feelings wouldn't be hurt, though, if someone with better art skills made it even nicer.

pupnewfster:
Rewrote various parts of this so that it works a bit better and also doesn't reorder the list of modes.

Issue #6436, opened 2020-08-07
Config option for Meka-Tool to disable extended vein mining


halvors:
**Describe the the feature you'd like**
The Meka-Tool should have an config option for disabling extended vein mining mode like the Atomic Disassembler has.

PR #6453, opened 2020-08-12
Added config option for disabling extended vein mining for the Meka-Tool

halvors:
Added config option for disabling extended vein mining for the Meka-Tool, this also removes HUD and Module Tweaker UI for extended vein mining when disabled.

Think this is the way to best integrate it with the existing code.

### DIFF
--- a/src/main/java/mekanism/client/gui/element/custom/GuiModuleScreen.java
+++ b/src/main/java/mekanism/client/gui/element/custom/GuiModuleScreen.java
@@ -13,6 +13,7 @@ import mekanism.common.MekanismLang;
 import mekanism.common.content.gear.Module;
 import mekanism.common.content.gear.ModuleConfigItem;
 import mekanism.common.content.gear.ModuleConfigItem.BooleanData;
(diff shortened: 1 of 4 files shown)


halvors:
This implements feature requested in issue #6436

pupnewfster:
I need to review this PR further but I would sort of rather it just use the same config option as the atomic disassembler does for extended vein mining (and adjust the comment on that config option slightly)

halvors:
@pupnewfster I see, that is a fairly small change, just not sure under what category in config to place it in, there is no general category in gear config, should it be in some other config file?

halvors:
Should i put it directly under the General category in general.toml or should it still be in gear.toml somehow?

pupnewfster:
Hmm actually on second thought how you have the config now is fine, given a pack dev may want to allow the meka tool (which is a later game item) have generic vein mining but not allow the atomic disassembler.

halvors:
I see, good. That was my initial thought.

pupnewfster:
Ended up having to do some cleanup due to the implementation not being ideal in terms of how data was being handled if values changed or things, but I believe it works fine now.

PR #6459, opened 2020-08-13
Update Development.md

rodrigokiller:
Minecraft version updated.

### DIFF
--- a/docs/Development.md
+++ b/docs/Development.md
@@ -1,6 +1,6 @@
 # Development
 
-Mekanism is developed for Minecraft 1.7.10 using Minecraft Forge.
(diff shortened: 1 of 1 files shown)


pupnewfster:
I am going to leave this open as a reminder to get to it, but I probably won't actually end up merging it as the majority of that file actually should just be rewritten as most of it is invalid as of Mekanism 1.15. I just had not realized that file existed in the first place so never got around to updating it.

rodrigokiller:
I totally understand and agree. I didn't read all the page, and didn't know that most of it is about Mekanism older versions. Maybe it should be as it is, so we could (I suggest) ignore this pull request.

PR #6508, opened 2020-08-26
Added back 1.12 recipes for Applied Energistics 2

### LINKED ISSUE #6479: AE2 dust recipes missing from Mekanism Crusher
*Please use the search functionality before reporting an issue. Also take a look at the closed issues!*
Closed issue: #4881

#### Issue description:
The certus, charged certus and fluix dust from Applied Energistics 2 can't produced in the crusher.
ModTweaker It seems that no one has maintained it for a long time and relying on another mod to solve the problem seems like a bad idea.

#### Steps to reproduce:
1. Place certus quartz or fluix into crusher.

**Forge:** 1.16.1-forge-32.0.108
**Mekanism:** Mekanism-1.16.1-10.0.8.431
**Other relevant version:** appliedenergistics2-8.0.0-alpha.3

shartte:
Fixes #6479

For the record: Not a fan of the recipes that directly convert crystals and dusts into their purified variants.
The recipes that do this for seeds I find less offensive (although the energy requirement might be tweaked if your recipes support it), since it's similar to the good 'ol AE2Stuff crystal grower.

One implementation note: AE2 now exposes the API only in common startup since it's not thread-safe to do so in the Mod constructor. Since this is datagen, common setup never runs, so the public-facing API is never made available. That's the reasoning for grabbing the API instance from the internal class here, which is slightly less stable.
We are however not going to change the item IDs in 1.16.1 anymore anyway.

### DIFF
--- a/src/datagen/main/java/mekanism/common/recipe/compat/AE2RecipeProvider.java
+++ b/src/datagen/main/java/mekanism/common/recipe/compat/AE2RecipeProvider.java
@@ -1,0 +1,109 @@
+package mekanism.common.recipe.compat;
+
+import appeng.api.IAppEngApi;
+import appeng.api.definitions.IBlocks;
+import appeng.api.definitions.IItems;
+import appeng.api.definitions.IMaterials;
+import appeng.core.Api;
+import mekanism.api.datagen.recipe.builder.ItemStackToItemStackRecipeBuilder;
+import mekanism.api.recipes.inputs.ItemStackIngredient;
+import mekanism.common.Mekanism;
+import net.minecraft.data.IFinishedRecipe;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+import java.util.function.Consumer;
+
+@ParametersAreNonnullByDefault
+public class AE2RecipeProvider extends CompatRecipeProvider {
+
+    public AE2RecipeProvider() {
+        super("appliedenergistics2");
+    }
+
+    @Override
+    protected void registerRecipes(Consumer<IFinishedRecipe> consumer, String basePath) {
+        // We cannot use the public API here since data generators do not run the common setup phase,
+        // which would normally expose the API to addons after AE has initialized and registered all of its blocks.
+        IAppEngApi api = Api.instance();
+        if (api == null) {
+            throw new IllegalStateException("AE2 was not initialized");
+        }
+
+        IItems items = api.definitions().items();
+        IMaterials materials = api.definitions().materials();
+        IBlocks blocks = api.definitions().blocks();
+
+        // Certus Crystal -> Certus Dust
+        ItemStackToItemStackRecipeBuilder.crushing(
+                ItemStackIngredient.createMulti(
+                        ItemStackIngredient.from(materials.certusQuartzCrystal()),
+                        ItemStackIngredient.from(materials.certusQuartzCrystalCharged())
+                ),
+                materials.certusQuartzDust().stack(1)
+        ).addCondition(modLoaded)
+                .build(consumer, Mekanism.rl(basePath + "certus_crystal_to_dust"));
+
+        // Fluix Crystal -> Fluix Dust
+        ItemStackToItemStackRecipeBuilder.crushing(
+                ItemStackIngredient.from(materials.fluixCrystal()),
+                materials.fluixDust().stack(1)
+        ).addCondition(modLoaded)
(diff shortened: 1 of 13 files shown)


shartte:
p.s. @pupnewfster Do you have brine as a fluid? Might be more sensible to infuse a crystal with brine to get a purified crystal if you do

src/datagen/main/java/mekanism/common/recipe/compat/AE2RecipeProvider.java
```diff
@@ -0,0 +1,109 @@
...
+                materials.fluixDust().stack(1)
+        ).addCondition(modLoaded)
+                .build(consumer, Mekanism.rl(basePath + "fluix_crystal_to_dust"));
+
+        // Certus Ore -> Certus Crystal
+        ItemStackToItemStackRecipeBuilder.enriching(
+                ItemStackIngredient.from(blocks.quartzOre()),
+                materials.certusQuartzCrystal().stack(4)
+        ).addCondition(modLoaded)
+                .build(consumer, Mekanism.rl(basePath + "certus_ore_to_dust"));
```
thiakil, line 58: bad name - dust != crystal

src/datagen/main/java/mekanism/common/recipe/compat/AE2RecipeProvider.java
```diff
@@ -0,0 +1,109 @@
...
+
+        // Charged Certus Ore -> Charged Certus Crystal
+        ItemStackToItemStackRecipeBuilder.enriching(
+                ItemStackIngredient.from(blocks.quartzOreCharged()),
+                materials.certusQuartzCrystalCharged().stack(4)
+        ).addCondition(modLoaded)
+                .build(consumer, Mekanism.rl(basePath + "charged_certus_ore_to_dust"));
```
thiakil, line 65: bad name - dust != crystal

thiakil (changes requested):

thiakil:
> Not a fan of the recipes that directly convert crystals and dusts into their purified variants.

It doesn't double it, so while it may not be particularly _useful_ it shouldn't break balancing?

shartte:
> > Not a fan of the recipes that directly convert crystals and dusts into their purified variants.
> 
> It doesn't double it, so while it may not be particularly _useful_ it shouldn't break balancing?

It's that pure crystals are **required** for higher-level recipes. This either forces people to manually grow them, or figure out an automation setup, which is not trivial. If they can produce them by dumping them into an easy-to-automate Mekanism machine however, that trivializes that part of the content.

The doubling part is not really troubling to me, I mean the ores give 5 crystals each already :D

src/datagen/main/java/mekanism/common/recipe/compat/AE2RecipeProvider.java
shartte: Fixed. Copypasta :-|

src/datagen/main/java/mekanism/common/recipe/compat/AE2RecipeProvider.java
shartte: Fixed. Copypasta :-|

thiakil:
> If they can produce them by dumping them into an easy-to-automate Mekanism machine however, that trivializes that part of the content.

It's already pretty trivial imo. It's just a trade-off between amount of ore/crystals you throw at it vs more grindy automation.

Also, you could have just left those recipes out and made us add them if we really wanted them (I'm not terribly fussed, though the crystal purification has always seemed unnecessarily grindy in base AE2)

> Don't you have some form of brine fluid that you could make it consume and use the machines with fluid enrichment instead?

Yes, though it would just be a different grindy way as brine isn't super simple to get going

shartte:
@thiakil I think we may have different definitions of grindy. To automate crystals in AE2 you need to understand what a subnet is and actually understand how formation/annihilation planes work.

I would have left them out, but when talking to @pupnewfster, he wanted me to restore the old recipes.

pupnewfster:
If you could update this to 1.16.3 (it seems that alphas are also now being mirrored https://modmaven.k-4u.nl/appeng/appliedenergistics2/8.1.0-alpha.3/) I will go ahead and merge it (otherwise I will probably do so myself sometime before releasing 10.0.10). Also if you two think that a couple recipes should be removed feel free to do so, I mainly said for now just port them all as I don't know what the majority of them are and it is easier to remove than have to relook them up.

shartte:
Rebased & Squashed

PR #6606, opened 2020-10-09
Add some Processing Integration for AE2 Sky Stone

ToMe25:
 * Add an Crusher recipe to convert Sky Stone to Sky Stone Dust
 * Add an Enrichment Chamber recipe to convert Sky Stone Dust to Sky Stone
 * Add an Enrichment Chamber recipe to convert Smooth Sky Stone to Sky Stone

Atleast two of there recipes were so common in the 1.12.2 modpacks i played that i thought they were part of mekanism itself, as it seems like there weren't, this pr is correcting that.
Note that these recipes are mostly only useful in modpacks, as AE2 itself has no use for sky stone dust, and no way besides the grindstone to get it.
If the modpack only nature of two of these recipes makes them not fitting what mekanism should be feel free to close this pr.


thiakil (approved):
Fine by me, ~~though I think the smooth sky stone one is a smelting recipe in AE2?~~

PR #6607, opened 2020-10-09
Fakeplayer Improvements

thiakil:
Lots of changes, might help to go through it commit by commit

- Fixed the usage of `WeakReference`; before it was maintaining a strong reference as a static field
- Moved getting a fake player to a lambda/consumer method, to allow easy & consistent clean up
- Make the Digital Miner pretend to be the owner when using it's fake player!


src/main/java/mekanism/common/util/MinerUtils.java
```diff
@@ -18,7 +19,7 @@
     private MinerUtils() {
     }
 
-    public static List<ItemStack> getDrops(ServerWorld world, BlockPos pos, boolean silk, BlockPos minerPosition) {
+    public static List<ItemStack> getDrops(ServerWorld world, BlockPos pos, boolean silk, BlockPos minerPosition, UUID minerOwner) {
```
pupnewfster, line 22: Part of me wonders if it is at the point that it may make more sense to just make this be a private method in `TileEntityDigitalMiner` rather than having to keep adding more params to this. It originally was in its own class because of how much extra logic was needed for figuring out the drops but with loot contexts it doesn't seem really necessary anymore.

src/main/java/mekanism/common/tile/machine/TileEntityDigitalMiner.java
```diff
@@ -441,10 +442,12 @@ private boolean canMine(BlockPos pos) {
...
         BlockState state = world.getBlockState(pos);
-        PlayerEntity dummy = Objects.requireNonNull(Mekanism.proxy.getDummyPlayer((ServerWorld) world, getPos()).get());
-        BlockEvent.BreakEvent event = new BlockEvent.BreakEvent(world, pos, state, dummy);
-        MinecraftForge.EVENT_BUS.post(event);
-        return !event.isCanceled();
+        return MekFakePlayer.withFakePlayer((ServerWorld) world, this.pos.getX(), this.pos.getY(), this.pos.getZ(), dummy -> {
+            dummy.setEmulatingUUID(getSecurity().getOwnerUUID());//pretend to be the owner
+            BlockEvent.BreakEvent event = new BlockEvent.BreakEvent(world, pos, state, dummy);
+            MinecraftForge.EVENT_BUS.post(event);
+            return !event.isCanceled();
```
pupnewfster, line 449: We might as well actually clean this line up because `post` returns if it is canceled or not for cancellable events. So this could be `return !MinecraftForge.EVENT_BUS.post(new BlockEvent.BreakEvent(world, pos, state, dummy));` (It might have to be inverted I don't remember for certain if it returns true for if it is cancelled or not, the javadocs on `post` should say though)

pupnewfster (commented):
From an initial glance this looks like a decent system. One thing that may make it a bit easier to use is add an extra param that passes a `TileEntityMekanism` and checks `tile.hasSecurity()` and if it does, sets the emulating ID that way if there is an owner in the security component. It also may be worth making the basic laser (even without it having a GUI) have an owner it keeps track of by adding `new AttributeSecurity()` to this [line](https://github.com/mekanism/Mekanism/blob/1.16.x/src/main/java/mekanism/common/registries/MekanismBlockTypes.java#L333) for what attributes it has. At the very least even if we don't do so, adding a helper method/optional param to pass a `TileEntityMekanism` would allow the laser amplifier and laser tractor beam to use their owner's when breaking blocks even if the basic laser doesn't.

kopkalol123:
Pogchamp

thiakil:
Not a fan of adding the TileEntity param and have it magically/force the behaviour, but agreed with the other stuff

PR #6647, opened 2020-10-21
Implement NBTUtils#setResourceLocationIfPresent() the same way as in 1.16, fixes #6644

### LINKED ISSUE #6644: [1.15.2] Oredictionificator crashes dedicated server when placed from an item with existing NBT data
#### Issue description:
When an Oredictionificator that was placed and configured before is placed again, it will crash a dedicated server due to `ResourceLocation#isResourceNameValid() / ResourceLocation#func_217855()` being marked client-only.
Looking through the Mekanism and MC sources, I found that this can be fixed by using the implementation of `NBTUtils#setResourceLocationIfPresent()` from 1.16 because `ResourceLocation#tryCreate()` also exists in MC 1.15.

A very similar issue was reported [here](https://github.com/mekanism/Mekanism/issues/6333) but in that case it seems to not have caused a crash.

#### Steps to reproduce:
1. Join a dedicated server
2. Place an Oredicitionificator, break it and place the dropped item again

#### Version:
**Forge:** 1.15.2-31.2.41
**Mekanism:** 1.15.2-9.10.9.422

#### Crashlog
https://gist.github.com/XFactHD/da1c58bce57f6a0a02f1ec3ef086b115

XFactHD:
As mentioned in #6644 the current implementation of `NBTUtils#setResourceLocationIfPresent()` crashes a dedicated server because `ResourceLocation#isResourceNameValid() / ResourceLocation#func_217855()` is client-only, this PR fixes this by implementing it the same way it is done in 1.16.

Also fixes block placing trying to send a packet to all players on the client and not handling the placer being null.

### DIFF
--- a/src/main/java/mekanism/common/block/BlockMekanism.java
+++ b/src/main/java/mekanism/common/block/BlockMekanism.java
@@ -168,8 +168,14 @@ public abstract class BlockMekanism extends Block {
             ISecurityItem securityItem = (ISecurityItem) item;
             tile.getSecurity().setMode(securityItem.getSecurity(stack));
             UUID ownerUUID = securityItem.getOwnerUUID(stack);
-            tile.getSecurity().setOwnerUUID(ownerUUID == null ? placer.getUniqueID() : ownerUUID);
-            Mekanism.packetHandler.sendToAll(new PacketSecurityUpdate(placer.getUniqueID(), null));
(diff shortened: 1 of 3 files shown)


pupnewfster:
Not sure when (if ever) we will end up doing another 1.15 release but for now I will merge this in so that if we ever do it is included.

PR #6661, opened 2020-10-27
Gradle cleanup

pupnewfster:
- Cleaned up deprecated Gradle calls
- Rewrote gradle calls slightly to remove the majority of IntelliJ warnings on the build.gradle file. The one change I am not too sure about is moving the maven publication to a closure and setting the project for it
- Copy the modified mods.toml file to the intellij build folder in addition to the gradle build folder. This only somewhat helps as the intellij runconfigs have the wrong paths for MOD_CLASSES by default, but now if they get manually fixed it can actually run now. (I didn't bother experimenting with manually specifying the values as I am fine with continuing to just use the gradlle intellij runconfigs
- Moved a lot of the all jar merge logic into a separate file and add support for merging tags that are added by multiple modules. In theory I think this fixes all the remaining merge issues the all jar had.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,23 +59,23 @@ sourceSets {
             //But exclude the cache of the generated data from what gets built
             exclude '.cache'
         }
-        compileClasspath += sourceSets.api.output
-        runtimeClasspath += sourceSets.api.output
+        compileClasspath += api.output
+        runtimeClasspath += api.output
     }
-    additions { setupSourceSet((SourceSet) sourceSets.additions, 'additions') }
-    generators { setupSourceSet((SourceSet) sourceSets.generators, 'generators') }
-    defense { setupSourceSet((SourceSet) sourceSets.defense, 'defense') }
-    tools { setupSourceSet((SourceSet) sourceSets.tools, 'tools') }
+    additions { setupSourceSet(additions, 'additions') }
+    generators { setupSourceSet(generators, 'generators') }
+    defense { setupSourceSet(defense, 'defense') }
+    tools { setupSourceSet(tools, 'tools') }
     test {
-        compileClasspath += sourceSets.api.output + sourceSets.main.output + sourceSets.additions.output + sourceSets.generators.output + sourceSets.defense.output + sourceSets.tools.output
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -643,63 +570,66 @@ task optimizePng {
 def MAVEN_PASSWORD_PROPERTY_NAME = 'mekanismMavenPassword'
 def MAVEN_URL_PROPERTY_NAME = 'mekanismMavenUrl'
 
-publishing {
-    publications {
-        maven(MavenPublication) {
-            from components.java
+def mavenPublicationsClosure = {
```
thiakil, line 573: Not sure what the point of this is?

build.gradle
```diff
@@ -61,23 +62,23 @@ sourceSets {
...
-    datagentools { setupDataGenSourceSet(sourceSets.datagentools, sourceSets.tools, 'tools') }
+    datagenmain { setupDataGenSourceSet((SourceSet) sourceSets.datagenmain, null, 'main') }
+    datagenadditions { setupDataGenSourceSet((SourceSet) sourceSets.datagenadditions, (SourceSet) sourceSets.additions, 'additions') }
+    datagengenerators { setupDataGenSourceSet((SourceSet) sourceSets.datagengenerators, (SourceSet) sourceSets.generators, 'generators') }
+    datagendefense { setupDataGenSourceSet((SourceSet) sourceSets.datagendefense, (SourceSet) sourceSets.defense, 'defense') }
+    datagentools { setupDataGenSourceSet((SourceSet) sourceSets.datagentools, (SourceSet) sourceSets.tools, 'tools') }
 }
 
-//This method sets up an additional sourceset in src/$name and adds a reference to the corresponding
+//This method sets up an additional sourceSet in src/$name and adds a reference to the corresponding
```
thiakil, line 81: In this section, the casts seem to be able to be replaced by a short ref (i.e. instead of `sourceSets.foo`, just `foo`). Seems to import to IDEA just fine

build.gradle
pupnewfster: It mainly was to remove a bunch of IntelliJ's warnings but I think I will revert that as it is the part I am least sure about

PR #6705, opened 2020-11-13
Batter first person looking for flamethrower.

FierceManul:
Batter first person looking for flamethrower.
屏幕截图(184)

### DIFF
--- a/src/main/resources/assets/mekanism/models/item/flamethrower.json
+++ b/src/main/resources/assets/mekanism/models/item/flamethrower.json
@@ -1,48 +1,48 @@
 {
-	"parent": "builtin/entity",
-	"textures": {
-		"particle": "mekanism:block/steel_casing"
-	},
-	"display": {
-		"gui": {
-			"rotation": [ 90, 135, 270 ],
-			"translation": [ -10.5, 6, 0 ],
-			"scale": [ 0.625, 0.625, 0.625 ]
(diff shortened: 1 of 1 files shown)


pupnewfster (changes requested):
Firstly there is no reason for all the spacing changes that are happening in this file (please revert that), and secondly this will not be merged unless you or someone else fixes/moves the location the flame particle hovers while it is idle (that thing that in your picture is in the middle, of the screen and is no longer at the tip of the flamethrower)

FierceManul:
But i don't know how to do. The original version is really ugly. sad ╯︿╰

thiakil:
https://github.com/mekanism/Mekanism/blob/1.16.x/docs/GoogleStyle4Indents.xml
That's our code style, you can import it if you're using IDEA.

pupnewfster:
And this is where the math happens: https://github.com/mekanism/Mekanism/blob/1.16.x/src/main/java/mekanism/client/render/RenderTickHandler.java#L247-L265

pupnewfster:
Went ahead and manually merged this in https://github.com/mekanism/Mekanism/commit/e12a95fec6362cbdd111f9d5f815cc58220b4858 and also adjusted the flame's rendering to be in the correct spot, so this improvement will be included in 10.0.18 once it is released.

Issue #6445, opened 2020-08-10
add computercraft integration to mekanisim v10


ghost:
older versions of mekanisim had computercraft intergation, a fork of computercraft exists for 1.16.1, the forked version of computercraft worked with mekanisim in 1.12.2, however it doesn't list anything in mekanisim v10 as a component.  is this something to do with mekanism or the computercraft fork? it would be awesome to see mekanisim having support for the computercraft fork.

this is the fork i'm referring to: https://www.curseforge.com/minecraft/mc-mods/cc-tweaked

pupnewfster:
This is something I am planning on looking into some eventually for V10.1, though no idea when that will be (hopefully sometime this year)

pupnewfster:
Rough copy paste of what I put on discord in our announcement's channel so that people who find this issue via github can see as well:

For people who use/used ComputerCraft (or OpenComputers for that matter even though it is not updated past 1.12); it would be of use if you could in a structured manner add comments here stating what methods (both in terms of setters and getters) would be helpful for the various tiles (or type of tiles) for Mekanism's integration to have when I look into reimplementing compat with CC for Mek v10.1. Given I haven't really ever bothered using CC or OC, so while I can probably look at some of the things we exposed back in 1.12, I can't really get a sense of what other things might be useful that we didn't used to have. (Such as being able to control the side config be of any use?)

Seniorendi:
Methods for Reactors:
- getEnergyPerTick() 
- activateReactor() 
- stopReactor() 
- getCoolant() 
- getFuel() 
- getWaste() 
- getHeatedCoolant() 
- getTemperatureInCelcius() 
- getTemperatureInKelvin() 
- getHeatingRate() 
- getBurnRate() 
- setBurnRate()
- getBoilEfficiency()

Most of them are for the Fission Reactor(I do not know a lot of the Fusion Reactor).

Energy Cubes:
- getTransferRate()

Teleporter
- setFrequency()
- getFrequency()
- removeFrequency()

Praspinel:
I have found that many mekanism devices support the redstone comparator.
my opinion is that it would be enough to add only a logic adapter for other multiblock structures.
a beautiful example is the reactor logic adapter.
reactor logic adapter only lacks analog mod (0-15 signal strength).
the problem with the turbine's redstone comparator is that it only detects the state of steam.
the analog redstone output of the stored energy signals is very missing.

Lupus590:
[...]

PR #6824, opened 2020-12-28
Added ComputerCraft integration to main multiblocks

TheJebForge:
Was frustrated when I couldn't make an induction matrix power display since normal Forge Energy API runs on ints and advanced cell in induction matrix is larger than int. So, I decided to add computercraft integration on my own, it's not much, supports only 5 multiblocks from main module, but at least that's a start. #6445

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -329,6 +329,8 @@ repositories {
             includeGroup 'mcp.mobius.waila'
         }
     }
+
+    maven { url 'https://squiddev.cc/maven/' }
 }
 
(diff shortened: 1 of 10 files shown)


pupnewfster:
While I do appreciate the help with this I am not going to merge this mostly given I am working on adding CC integration in V10.1, and this implementation sort of goes against how I am designing the integration there. Additionally this PR makes a complete mess of the LazyOptional API making it so that they can't be properly cached.

PR #6938, opened 2021-01-30
Allow flexible teleporter frames up to 10x10.

martica:
This change extends the frame detection of teleporters to allow frames up to 10x10, in all directions and orientations.

Additionally, it prevents teleporters with solid blocks inside the frame from working.

https://i.imgur.com/gkoKxQJ.jpg


Seniorendi:
And does the teleportation work?

thiakil:
Please switch to using our multiblock system rather than have it implement its own structure validation.

Sidenote: your usage of `Optional` seems quite redundant, `null` would probably suffice in many cases.

pupnewfster:
Also by solid blocks I presume things like rails still work fine inside of them?

martica:
> Please switch to using our multiblock system rather than have it implement its own structure validation.

I'll check it out. I hadn't looked that far into the codebase. I was just working from what was already there.

> And does the teleportation work?

Yep. For entities, mobs, players, players on horses, players in minecarts, minecarts alone. I don't think my changes affected that.

> Also by solid blocks I presume things like rails still work fine inside of them?

Yes. It's possible I've made it too restrictive, the old code didn't care what was inside the frame at all. It may be better to allow users more flexibility and if there are blocks in the frame, don't worry about it. I'd still avoid allowing teleporter blocks/teleporter frame blocks inside as nested teleporters could lead to confusing behaviour.

> Sidenote: your usage of Optional seems quite redundant, null would probably suffice in many cases.

Ah. My day job's code style is to avoid generally avoid nulls.

pupnewfster:
> I'll check it out. I hadn't looked that far into the codebase. I was just working from what was already there.

Ya we have a pretty decent/expansive system for multiblocks now in V10 (though I am not sure if some parts of it will need to be adjusted for purposes of the teleporter as it is 2D and in theory the frames can be shared). The main reason we didn't bother switching to it before when we wrote the new system and unified the rest of our multiblock formation code is it seemed a bit overkill for such a small and rigid structure as the teleporter used to be confined to.

> I'd still avoid allowing teleporter blocks/teleporter frame blocks inside as nested teleporters could lead to confusing behaviour.

That is a good point and one thing I didn't realize would be an issue with this, though I think in theory it shouldn't be that hard in the multiblock validation to  blacklist frames/teleporter blocks as invalid "inner" multiblocks.

martica:
I got to the point where teleporters do something again. There are a bunch of rough edges still.

src/main/java/mekanism/common/content/teleporter/TeleporterValidator.java
```diff
@@ -0,0 +1,56 @@
...
+import mekanism.common.registries.MekanismBlockTypes;
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockState;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.chunk.IChunk;
+
+public class TeleporterValidator extends CuboidStructureValidator<TeleporterMultiblockData> {
+
+    public static final VoxelCuboid MIN_BOUNDS = new VoxelCuboid(1, 1, 1);
+    public static final VoxelCuboid MAX_BOUNDS = new VoxelCuboid(17, 18, 17);
```
pupnewfster, line 19: I don't think these bounds are correct at all.

src/main/java/mekanism/common/lib/multiblock/Structure.java
```diff
@@ -149,8 +149,8 @@ public void add(Structure s) {
                         //Otherwise, if the minor map does have an entry, merge them
                         minorPlane.merge(value);
                         //If after merging the planes
-                        if (minorPlane.hasFrame() && minorPlane.length() >= 2 && minorPlane.height() >= 2) {
-                            // the plane has a frame and is at least two by two
+                        if (minorPlane.hasFrame()) {
+                            // the plane has a frame
```
pupnewfster, line 153: I think this change has some significant implications and in fact may entirely invalidate the point/efficiency of having two plane types (though @aidancbrady can probably say better)

src/main/java/mekanism/common/lib/multiblock/StructureHelper.java
```diff
@@ -21,15 +21,16 @@ private StructureHelper() {
      * @param structure structure to check
      * @param minBounds minimum size of the cuboid
      * @param maxBounds maximum size of the cuboid
+     * @param filled if true, the sides must be filled, otherwise, just framed
      *
      * @return found cuboid, or null if it doesn't exist
      */
-    public static VoxelCuboid fetchCuboid(Structure structure, VoxelCuboid minBounds, VoxelCuboid maxBounds) {
+    public static VoxelCuboid fetchCuboid(Structure structure, VoxelCuboid minBounds, VoxelCuboid maxBounds, boolean filled) {
```
pupnewfster, line 28: This method should not be used for this purpose as the entire point is for if it is all sides, the other method is much better used for cases like this as it was designed for non rectangular multiblocks and multiblocks that are not "full" for each face such as the thermal evaporation tower.

src/main/java/mekanism/common/registries/MekanismBlockTypes.java
```diff
@@ -317,8 +318,14 @@ private MekanismBlockTypes() {
           .withEnergyConfig(() -> TELEPORTER_USAGE, MekanismConfig.storage.teleporter)
           .withSupportedUpgrades(EnumSet.of(Upgrade.ANCHOR))
           .without(AttributeStateActive.class, AttributeStateFacing.class, AttributeParticleFX.class)
+          .with(new AttributeSecurity(), new AttributeMultiblock())
```
pupnewfster, line 321: You should rebase as I made changes earlier today to make it so that these attributes only ever have one instance created to save slightly on memory.

src/main/java/mekanism/common/tile/TileEntityTeleporter.java
```diff
@@ -1,14 +1,10 @@
...
+import com.google.common.collect.ImmutableMap;
 import it.unimi.dsi.fastutil.objects.ObjectOpenHashSet;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
+
+import java.util.*;
```
pupnewfster, line 7: We don't use wildcard imports

pupnewfster (changes requested):
I didn't review everything but left a couple comments. I also am not sure if the way this changes it breaks frames being able to be part of multiple teleporters (for example two both vertical ones that are perpendicular to each other). It may be worth making the frames structural and adjusting or just making another multiblock validator that isn't Cuboid not sure. A good part of me also hopes aidan will be able to way in here some as he knows the logic behind the multiblock system a lot better so may be able to better say when parts are breaking things entirely.

src/main/java/mekanism/common/tile/TileEntityTeleporter.java
martica: Ugh. IntelliJ kept doing that. :(

pupnewfster: This is the code formatting stylesheet we use with IntelliJ https://github.com/mekanism/Mekanism/blob/1.16.x/docs/GoogleStyle4Indents.xml

src/main/java/mekanism/common/content/teleporter/TeleporterValidator.java
martica: Yeah. Those were just copied from elsewhere and adjusted while I tried to get it to work.

pupnewfster: Honestly I think a new validator that isn't directly a CuboidValidator may be better/more effective as realistically we want a VoxelPlane not a VoxelCuboid and there isn't a great way to specify the min bounds for a flat thing using the cuboid system.

src/main/java/mekanism/common/tile/TileEntityTeleporter.java
martica: That will certainly help.

src/main/java/mekanism/common/content/teleporter/TeleporterValidator.java
martica: Yeah. I found the structure concept a bit hard to follow so I was trying to get to a working state again before I explored something more fit to purpose.

martica:
> making the frames structural 

Is there an example of this already? A regression in the shared teleporter frames would be ugly.

pupnewfster:
Structural glass is one, take a look at `TileEntityStructuralMultiblock`. I am not sure if there would be side effects such as it allowing using them in other multiblocks that are not the teleporter in place of structural glass. Though I think there may be a flag for what types of multiblocks structural ones are valid in as some only support reactor glass and not structural glass. In general I believe the validator also doesn't allow structural multiblocks as the frame but as we likely will need a custom validator that ends up consisting of a structure check of a single major plane we can have that validator allow it if necessary.

martica:
I've rolled by the WIP commit. If I get time to look at it again, I'll make another attempt.

create-juicey-app:
Welp, we didn't got what we wanted, i dont know why

DBotThePony:
sad

PR #6943, opened 2021-01-31
Mekanism-Feature-Requests#178 Tunable fusion configs

telemenar:
Added some configs to tune fusion the goal here was to enable pack authors to have more options
to tune power generation from fusion without messing around with steam/boilers as that limits
the ability to tune fission.

I tried to be selective here my rough attempt was to just make some the existing statics configurable. 

I chose the 3 that I did like this:
thermocoupleEfficiency -> fusionThermocoupleEfficiency
Makes the passive RF generation tunable.

caseAirConductivity  -> fusionCasingThermalConductivity
This one is almost too broad my main goal for adding this one was to give the ability to limit conductance on the heat capacitor 
so that a pack creator can limit potential of connecting the fusion reactor to a boiler.

caseWaterConductivity  -> fusionWaterHeatingRatio
Limit steam production without limiting the turbine side of steam.


src/generators/java/mekanism/generators/common/config/GeneratorsConfig.java
```diff
@@ -101,6 +106,16 @@
               .defineList("windGenerationDimBlacklist", new ArrayList<>(), o -> o instanceof String && ResourceLocation.tryCreate(((String) o).toLowerCase(Locale.ROOT)) != null));
         builder.pop();
 
+        builder.comment("Fusion Settings").push(FUSION_CATEGORY);
+        fusionThermocoupleEfficiency = CachedDoubleValue.wrap(this, builder.comment("The fraction of the heat dissipated from the case that is converted to RF.")
```
pupnewfster, line 110: Joules not RF, we don't even support RF anymore the other system of energy we can support based on configs is FE.

pupnewfster (changes requested):

src/generators/java/mekanism/generators/common/config/GeneratorsConfig.java
telemenar: I'll just change it to Joules and fix the merge conflict.

pupnewfster (approved):

src/generators/java/mekanism/generators/common/content/fusion/FusionReactorMultiblockData.java
```diff
@@ -345,42 +342,45 @@ protected int getMultiblockRedstoneLevel() {
...
 
     public long getSteamPerTick(boolean current) {
         double temperature = current ? getLastCaseTemp() : getMaxCasingTemperature(true);
-        return MathUtils.clampToLong(HeatUtils.getSteamEnergyEfficiency() * caseWaterConductivity * temperature / HeatUtils.getWaterThermalEnthalpy());
+        return MathUtils.clampToLong(HeatUtils.getSteamEnergyEfficiency() * MekanismGeneratorsConfig.generators.fusionWaterHeatingRatio.get() * temperature / HeatUtils.getWaterThermalEnthalpy());
     }
 
     public static double getInverseConductionCoefficient() {
-        return 1 / caseAirConductivity;
+        return 1 / MekanismGeneratorsConfig.generators.fusionCasingThermalConductivity.get();
```
pupnewfster, line 384: I will do a closer pass tomorrow, but min value for this config option and any others that can be divided by should be something greater than zero and not zero.

pupnewfster (changes requested):

src/generators/java/mekanism/generators/common/content/fusion/FusionReactorMultiblockData.java
telemenar: Ahh good catch. Do you typically just have them set to 0.001 or something more interesting?

pupnewfster: 0.001 should be fine normally we use 1 but normally the cases we have it is an int not a double.

telemenar: I did a quick readthrough with an eye toward looking for div0 cases, I think it was just this configurable that needed tweaking  this way to prevent div0 cases at least in this file. (I didn't think too hard about which functions might return 0 used by other callers.)

src/generators/java/mekanism/generators/common/config/GeneratorsConfig.java
```diff
@@ -101,6 +106,16 @@
               .defineList("windGenerationDimBlacklist", new ArrayList<>(), o -> o instanceof String && ResourceLocation.tryCreate(((String) o).toLowerCase(Locale.ROOT)) != null));
         builder.pop();
 
+        builder.comment("Fusion Settings").push(FUSION_CATEGORY);
+        fusionThermocoupleEfficiency = CachedDoubleValue.wrap(this, builder.comment("The fraction of the heat dissipated from the case that is converted to Joules.")
+               .defineInRange("fusionThermocoupleEfficiency", 0.05D, 0D, 1D));
```
pupnewfster, line 111: The fusion prefix from these config options should be removed for example `thermocoupleEfficiency`, `casingThermalConductivity`, `waterHeatingRatio` as they are already in the fusion category so the config path is evident they are for fusion. The variable names however should be left as is.

pupnewfster (changes requested):

src/generators/java/mekanism/generators/common/config/GeneratorsConfig.java
telemenar: Done

pupnewfster (approved):

Issue #6536, opened 2020-09-08
Bronze Paxel is strongest weapon. Others never rebalanced.


VallenFrostweaver:
#### Issue description:
Bronze, refined obsidian, and steel (among others) were never rebalanced for damage values and bronze accidentally left at high value.
#### Steps to reproduce:
1. Compare damage values of swords/paxels/axes
#### Versions:
**Minecraft:** 1.16.1
**Forge:** 32.0.108
**Mekanism:** 10.0.9.432
**MekanismTools:** 10.0.9.432

An image of the current damage values of swords and paxels. You can see there is extreme unbalance in this and it means people will use earlier materials in favor. I believe this was just an oversite from the previous ports.
![image](https://user-images.githubusercontent.com/13642054/92460800-4dd45f80-f1c0-11ea-87e7-ca7ae750d555.png)
Below is a suggested route to keep the materials more balanced. Axes will need looking at as well but these are just a sample. The effort to get each should fall in line with the appropriate damage levels. At the very least, I would hope that bronze is taken off the top tier for damage values as it's not comparable to Refined Obsidian difficulty to obtain whatsoever.
![image](https://user-images.githubusercontent.com/13642054/92460916-76f4f000-f1c0-11ea-978a-eb1a9b2ea30d.png)

blackwing12344:
Harvest levels need to be adjusted as well, as netherite adds a higher their harvest level.

PR #6954, opened 2021-02-03
Increase Refined Obsidian harvest level to match netherite

blackwing12344:
Increase Refined Obsidian harvest level to match netherite
Even though Refined Obsidian is faster and does more damage than netherite, it has a lower harvest level.


thiakil:
What's the rationale behind raising its havest level, exactly?

blackwing12344:
> What's the rationale behind raising its harvest level, exactly?

There are some other mods that require that higher harvest level,  and refined obsidian is better than netherite.

pupnewfster:
We need to rebalance our various tools but aidan has been busy so we haven't done so yet. Feel free to comment about this on #6536 and we will take into consideration also adjusting the harvest levels.

PR #6963, opened 2021-02-05
Fixed #6961: Entities stuck on pathfinding

### LINKED ISSUE #6161: Can't use machinery
I am not able to use digital miner it gets ready but doesn't show to mine.

#### Issue description:I am not able to use any machinery and can't use jetpacks and I am not even able to put robot.
The mods which I have with it are inside the folder of drive link

https://drive.google.com/folderview?id=1glDILxCZFcErP5CsHIOx7iOQJb0kH3hs
Please help ASAP I want to play with my friends.

I have installed everything of version 1.15.2

### LINKED ISSUE #6961: Mekanism blocks pathfinding issue
*Please use the search functionality before reporting an issue. Also take a look at the closed issues!*

#### Issue description:

Some Mekanism blocks (noticeably cables and pipes, but maybe others) don't properly override `Block#allowsMovement()`.  Specifically, this method is intended to decide whether or not an entity can pathfind through it; the default for land-based and flying entities is to allow pathfinding if the block doesn't have an opaque collision shape.  This is the case for all Mekanism pipes and cables.  Unfortunately, because these blocks have a collision box, entities will get stuck (forever) when trying to pathfind through them.

#### Steps to reproduce:
1. Set up a row of Mek cables like in the screenshot below
2. Put some cows on the other side of the cables
3. Hold some wheat
4. The cows *should* step over or walk around the end of the cables to get to you, but will instead try (and fail) to walk through the cables.

The fix is fortunately very easy: just override `Block#allowsMovement()` to return false for any blocks which are solid but don't have an opaque collision box (i.e. not full cubes).  You'll notice vanilla overrides this method in many of its blocks to do this.

**Forge:** 36.0.0
**Mekanism:** 10.0.19
**Other relevant version:**  MC 1.16.5

[[gist](https://gist.github.com/) / [pastebin](https://pastebin.com/) / etc link here. Please make sure that it isn't set to expire.]

Seniorendi:
Fixes #6961

### DIFF
--- a/src/main/java/mekanism/common/block/transmitter/BlockTransmitter.java
+++ b/src/main/java/mekanism/common/block/transmitter/BlockTransmitter.java
@@ -29,6 +29,7 @@ import net.minecraft.block.material.Material;
 import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.ItemStack;
+import net.minecraft.pathfinding.PathType;
 import net.minecraft.util.ActionResultType;
 import net.minecraft.util.Direction;
 import net.minecraft.util.Hand;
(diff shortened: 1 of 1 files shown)


pupnewfster:
Thanks for this except I think I am going to close it and look into this further as the issue mentioned there are more blocks that have this issue (and we have a lot of blocks that are not full blocks) and I need to look more into how allow movement is actually used and how I want to structure things to reduce duplicate code and make it so we don't forget to add it in the future when we add new blocks.

Seniorendi:
Sure, I understand 👍🏻.

PR #6988, opened 2021-02-15
Math API Division Optimization

Chiefwright:
-Added optimizing methods for commonly used division operations.
-Added test coverage for each of the new methods.

### DIFF
--- a/src/api/java/mekanism/api/math/FloatingLong.java
+++ b/src/api/java/mekanism/api/math/FloatingLong.java
@@ -332,11 +332,53 @@ public class FloatingLong extends Number implements Comparable<FloatingLong> {
         if (toDivide.isZero()) {
             throw new ArithmeticException("Division by zero");
         }
-
         BigDecimal divide = new BigDecimal(toString()).divide(new BigDecimal(toDivide.toString()), DECIMAL_DIGITS, RoundingMode.HALF_EVEN);
         long value = divide.longValue();
         short decimal = parseDecimal(divide.toPlainString());
         return setAndClampValues(value, decimal);
+    }
+
+    /**
+     * Divides this {@link FloatingLong} by the given unsigned long primitive, modifying the current object unless it is a constant in which case it instead returns the
+     * result in a new object. Rounds to the nearest 0.0001
+     *
+     * @param toDivide The value to divide by represented as an unsigned long.
+     *
+     * @return The {@link FloatingLong} representing the value of dividing this {@link FloatingLong} by the given {@link FloatingLong}.
+     *
+     * @throws ArithmeticException if {@code toDivide} is zero.
+     * @apiNote It is recommended to set this to itself to reduce the chance of accidental calls if calling this on a constant {@link FloatingLong}
+     * <br>
+     * {@code value = value.divideEquals(toDivide)}
+     */
+    public FloatingLong divideEquals(long toDivide) {
+        if (toDivide == 0) {
+            throw new ArithmeticException("Division by zero");
+        }
+        if (this.isZero())
+            return this;
+
+        long val = Long.divideUnsigned(this.value, toDivide);
+        long rem = Long.remainderUnsigned(this.value, toDivide);
+
+        //just need to figure out remainder -> decimal
+        long dec;
+
+        //okay, now what if rem * SINGLE_UNIT * 10L will overflow?
+        if (Long.compareUnsigned(rem * 10, MAX_LONG_SHIFT) >= 0) {
+            //if that'll overflow, then todivide also has to be big. let's just lose some denominator precision and use that
+            dec = Long.divideUnsigned(rem, Long.divideUnsigned(toDivide, SINGLE_UNIT * 10L)); //same as multiplying numerator
+        } else {
+            dec = Long.divideUnsigned(rem * SINGLE_UNIT * 10L, toDivide); //trivial case
+            dec += (Long.divideUnsigned(this.decimal * 10L, toDivide)); //need to account for dividing decimal too in case toDivide < 10k
+        }
+
+        //usually will expect to round to nearest, so we have to do that here
+        if (Long.remainderUnsigned(dec, 10) >= 5)
+            dec+=10;
+        dec /= 10;
+
+        return setAndClampValues(val, (short) dec);
     }
 
     /**
@@ -382,10 +424,10 @@ public class FloatingLong extends Number implements Comparable<FloatingLong> {
             if (Long.compareUnsigned(this.value, MAX_LONG_SHIFT) >= 0) {
                 return (this.value / toDivide.decimal) * MAX_DECIMAL //lose some precision here, have to add modulus
                         + (this.value % toDivide.decimal) * MAX_DECIMAL / toDivide.decimal
-                        + this.decimal * MAX_DECIMAL / toDivide.decimal;
+                        + (long)this.decimal * MAX_DECIMAL / toDivide.decimal;
             } else {
                 long d = this.value * MAX_DECIMAL;
-                return d / toDivide.decimal + this.decimal * MAX_DECIMAL / toDivide.decimal; //don't care about modulus since we're returning integers
+                return d / toDivide.decimal + (long)this.decimal * MAX_DECIMAL / toDivide.decimal; //don't care about modulus since we're returning integers
             }
         }
     }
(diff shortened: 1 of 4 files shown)


src/api/java/mekanism/api/math/FloatingLong.java
```diff
@@ -328,12 +332,81 @@ public FloatingLong divideEquals(FloatingLong toDivide) {
...
+     *
+     * @return The {@link FloatingLong} representing the value of dividing this {@link FloatingLong} by the given {@link FloatingLong}.
+     *
+     * @throws ArithmeticException if {@code toDivide} is zero.
+     * @apiNote It is recommended to set this to itself to reduce the chance of accidental calls if calling this on a constant {@link FloatingLong}
+     * <br>
+     * {@code value = value.divideToLong(toDivide)}
+     */
+    public int divideToInt(FloatingLong toDivide) {
+        return divideToLong(toDivide) <= (long)Integer.MAX_VALUE ? (int)divideToLong((toDivide)) : Integer.MAX_VALUE;
```
pupnewfster, line 407: This can just use `MathUtils#clampToInt`

src/api/java/mekanism/api/math/FloatingLong.java
```diff
@@ -328,12 +332,81 @@ public FloatingLong divideEquals(FloatingLong toDivide) {
         if (toDivide.isZero()) {
             throw new ArithmeticException("Division by zero");
         }
+
```
pupnewfster, line 335: Unneeded whitespace change.

src/api/java/mekanism/api/math/FloatingLong.java
```diff
@@ -482,7 +555,33 @@ public FloatingLong divide(FloatingLong toDivide) {
      * @throws ArithmeticException if {@code toDivide} is zero.
      */
     public FloatingLong divide(long toDivide) {
```
pupnewfster, line 557: This method should instead do a similar thing to the divide(FloatingLong) (in that it does the copy) and have the actual logic added below be done in the divideEquals that takes a long as a parameter.

src/api/java/mekanism/api/math/FloatingLong.java
```diff
@@ -328,12 +332,81 @@ public FloatingLong divideEquals(FloatingLong toDivide) {
...
+    /**
+     * Divides this {@link FloatingLong} by the given {@link FloatingLong} rounded down to an integer value, modifying the current object unless it is a constant in which case it instead returns the
+     * result in a new object. This gets clamped at the upper bound of {@link Long#MAX_VALUE} rather than overflowing.
+     *
+             * @param toDivide The {@link FloatingLong} to divide by.
+     *
+             * @return The {@link FloatingLong} representing the value of dividing this {@link FloatingLong} by the given {@link FloatingLong}.
+            *
+            * @throws ArithmeticException if {@code toDivide} is zero.
+            * @apiNote It is recommended to set this to itself to reduce the chance of accidental calls if calling this on a constant {@link FloatingLong}
```
pupnewfster, line 351: Formatting is a bit wonky though not the end of the world.

src/api/java/mekanism/api/math/FloatingLong.java
Chiefwright: Fixed.

src/api/java/mekanism/api/math/FloatingLong.java
Chiefwright: Fixed.

src/api/java/mekanism/api/math/FloatingLong.java
Chiefwright: Done, but could use a review to make sure.

src/test/java/mekanism/api/math/FloatingLongTest.java
```diff
@@ -78,6 +78,14 @@ void testDivisionLargeNumerator() {
         Assertions.assertEquals(FloatingLong.create((long) 649_657 * 337), a.divide(b));
     }
 
+    @Test
+    @DisplayName("Test division with a very large denominator")
+    void testDivisionLargeDenominator() {
+        FloatingLong a = FloatingLong.create(922355340224119L);
+        FloatingLong b = FloatingLong.create(-1L);
+        Assertions.assertEquals(FloatingLong.create(0L, (short)1), a.divide(b));
```
BlueAgent, line 86: Denominator is the part under a fraction (a.k.a. divisor) but it's being used as the Numerator in this division. I might be missing something here though?

pupnewfster: FloatingLong uses unsigned longs internally so -1 is largest whole number FloatingLong

BlueAgent: Ah I see, sorry for wasting time.

Chiefwright: Sorry not sure I understand, is it the -1L that is confusing? For this test to give useful information I had to also include a decently large numerator (>= (2^64-1) / 10,000)?, otherwise the result will just be zero due to the minimum nonzero value of 0.0001. 
It's really the -1L (maximum unsigned long value) that I wanted to test. Let me know if I'm understanding your question incorrectly :)

BlueAgent: I misinteperated `a` as being the larger number.

Chiefwright: No worries! Happy to help/explain

src/api/java/mekanism/api/math/FloatingLong.java
```diff
@@ -334,6 +338,117 @@ public FloatingLong divideEquals(FloatingLong toDivide) {
...
+        long val = Long.divideUnsigned(this.value, toDivide);
+        long rem = Long.remainderUnsigned(this.value, toDivide);
+
+        //just need to figure out remainder -> decimal
+        long dec;
+
+        //okay, now what if rem * SINGLE_UNIT * 10L will overflow?
+        if (Long.compareUnsigned(rem * 10, MAX_LONG_SHIFT) >= 0) {
+            //if that'll overflow, then todivide also has to be big. let's just lose some denominator precision and use that
+            dec = Long.divideUnsigned(rem, Long.divideUnsigned(toDivide, SINGLE_UNIT * 10L)); //same as multiplying numerator
```
BlueAgent, line 370: This crashes with divide by zero when `Long.divideUnsigned(toDivide, SINGLE_UNIT * 10L)` evaluates to zero which happens when `toDivide < SINGLE_UNIT * 10L`.
Example of an input that causes this to happen is.
```java
FloatingLong.create(0, (short) 1).divide(2L);
```

I think one of the reasons why this is happening is because of:
```java
private static final long MAX_LONG_SHIFT = -1L / SINGLE_UNIT;
```
Evaluating is to zero because it's operating on a signed long.
Which means the condition:
```java
if (Long.compareUnsigned(rem * 10, MAX_LONG_SHIFT) >= 0)
```
Is often evaluating to true (any remainder that is non-zero).
Didn't seem to get caught by the Property tests since the randomly generated samples might have not have run into the situation?

Chiefwright: Thanks for catching the MAX_LONG_SHIFT error, I definitely will fix that. But I do not think there is any case of divide_by_zero in the `if (Long.compareUnsigned(rem * 10, MAX_LONG_SHIFT) >= 0)` case you mentioned. 

This is because the above if statement guarantees that the remainder (rem) is very large (larger than MAX_LONG_SHIFT / 10). Therefore, toDivide must also be very large (> rem+1). In this case, it cannot go to zero with the following
`Long.divideUnsigned(toDivide, SINGLE_UNIT * 10L)`

If you think it's hard to read through the logic, I'm all for changing it to add readability. Just let me know if you think it's required and I have no issues :)

Regarding the discrepancy, I will add a fix for the discrepancies, I agree they should be equal.

BlueAgent: Yes, sorry, it doesn't crash with that change to using unsigned division added. I should have made it clearer by seperating out the second issue (which I ran into after switching to unsigned division in testing but it was not related).

Chiefwright: Resolved with https://github.com/mekanism/Mekanism/pull/6988/commits/e49f3bf423269668e3dda01a49cf51ac281ce31e

PR #7106, opened 2021-04-13
Update feature request template

Seniorendi:
Fixed a funny typo after 15 months.

removed one "the"


thiakil:
Would you mind rewriting it to point people to https://github.com/mekanism/Mekanism-Feature-Requests instead?

pupnewfster:
I think github's "new" (like maybe a year old) format selector actually even has a way we could make it so it is opening it on the other repo when they press the button, but I haven't looked into how to set it up yet.

Seniorendi:
Well, I was too lazy 😄

PR #7124, opened 2021-04-21
Change Resistive Heater efficiency from 60% to 100%

tarbaII:
Changing the Resistive Heater efficiency from 60% to 100%

In real life, resistive heaters are 100% efficient at converting electric energy to heat energy, and I saw in the Mekanism configuration that the default is 60%. Since this mod is to some extent realistic, I thought that this might be a welcome change.


Naheulf:
In theory it's 100%. However in real case it's not the case because of Joule loss in transmission lines and thermal conduction loss. That why I think you should use something between 95% and 98%. (But I agree : 60% efficiency is not realist)

alyxferrari:
@Naheulf To my knowledge, the Mekanism cables don't take into account heat loss through the resistance of the wire, so considering that loss isn't accounted for in other parts of the mod, maybe it shouldn't be here too. Just a thought.

PR #7125, opened 2021-04-24
Update Forge Maven

Naheulf:
Minecraft forge maven URI has changed. I follow steps given by forge-bot on the Forge Discord:
How to fix your gradle build
Forge has recently updated maven endpoints, which has caused a bug in Gradle because the old one now redirects. To fix this, follow the steps below.
1. Update all endpoints in your build.gradle
You should change all references of https://files.minecraftforge.net/maven in your build.gradle to the newer endpoint at https://maven.minecraftforge.net/.
2. Update your ForgeGradle version
The maven endpoint used in ForgeGradle is hardcoded. A fix has been pushed for FG 3 and 4, while FG 1 and 2 are unsupported. Please run gradlew --refresh-dependencies in a terminal inside the project to force Gradle to use the latest ForgeGradle version.

Source: https://discord.com/channels/313125603924639766/796838585797050428/835278480761421845

Note : The step 2 update local cache. So there no impact on the repo itself

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
     repositories {
         maven {
             name "forge"
-            url "https://files.minecraftforge.net/maven"
+            url "https://maven.minecraftforge.net/"
         }
     }
     dependencies {
@@ -282,7 +282,7 @@ minecraft {
 repositories {
     maven {
         name 'ForgeFS'
-        url 'https://files.minecraftforge.net/maven'
+        url 'https://maven.minecraftforge.net/'
         content {
             includeGroupByRegex 'net\\.minecraftforge.*'
             includeGroup 'de.oceanlabs.mcp'


pupnewfster:
There isn't really much of a reason to do this at least on this branch as the active development branch at the moment is `V10.1`. Which I suppose if you want you could PR the one line change to (one of the two references doesn't exist in the 10.1 build.gradle) or I will do it in my next commit. I had thought I already updated it in the 10.1 branch but I updated a different spot relating to our content filtering.

Naheulf:
I send this PR because the last jar available on curse came from this branch.

Naheulf:
And because i'ts the default branch on this repo

pupnewfster:
Yes I am aware this is the default branch on the repo. And the last jar available on curse came from the master branch. The reason we haven't merged 10.1 into this branch yet is in case there is some absolutely massive gamebreaking bug reported that we have to do a hotfix for before 10.1. Forge updating their maven is not a gamebreaking bug.

pupnewfster:
Done as part of https://github.com/mekanism/Mekanism/commit/771daad364b011cf0ef542f2a60de0b465e9168d for the 10.1 branch

PR #7128, opened 2021-04-24
Implement Fill Held toggle for Electrolytic Breathing Unit

beholderface:
This adds a toggle to the electrolytic breathing unit that controls whether it will try to fill held items with gas, as I described in https://github.com/mekanism/Mekanism-Feature-Requests/issues/230. 
That second commit in the pull request is just undoing the auto-formatting that I told Eclipse to do before reading the contribution guidelines.

### DIFF
--- a/src/datagen/generated/mekanism/.cache/cache
+++ b/src/datagen/generated/mekanism/.cache/cache
@@ -23,8 +23,8 @@ d5bd6c407c9250a90e22e1f2c8805738b3830abf assets/mekanism/blockstates/lithium.jso
 c1be0ed56a542f711719367cf1e508f10063c695 assets/mekanism/blockstates/sulfuric_acid.json
 f5e3eb6799a15d0b72969547614dfc51dac53419 assets/mekanism/blockstates/tin_ore.json
 246c1904e7699b01a3a20f2a2b79b35b5d99aec9 assets/mekanism/blockstates/uranium_ore.json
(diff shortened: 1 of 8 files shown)


beholderface:
The fourth commit is me fixing a major error I made when removing the autoformatting. It has been tested, unlike the second and third commits.

pupnewfster:
1. Please target the `v10.1` branch over the default 1.16.x branch as that is the one that is currently under development. The main reason it is in a separate branch still is in case some gamebreaking bug comes up that we need to fix before 10.1.
2. Lang changes should only be done to `en_us` and `en_ud` other lang entries are done via crowdin as needed/if they are different than english. Lang changes also should be made for en_us via data generators (`MekanismLangProvider` and running runData), as otherwise they will just be deleted the next time the data generator is ran.
3. The spacing for the changes in ModuleMekaSuit seems rather screwed up.

beholderface:
ok I believe I have fixed the issues with the lang files but I don't really know what to do about spacing being screwed up

beholderface:
I've started working on redoing my code based on the v10.1 branch

beholderface:
I've finished redoing my code but I don't see any option to put it in this pull request (something to do with the branch it started with maybe, idk), is there a way to do that or should I close this pull request and make a new one with the redone code?

pupnewfster:
You in theory could force push to your branch, though you may find it simpler to just close and make a new one that started from the 10.1 branch just due to the mass number of changes between the two branches.

PR #7150, opened 2021-05-01
Created the Geiger Unit

beholderface:
It displays ambient radiation levels on the GUI, as described in [feature request 185](https://github.com/mekanism/Mekanism-Feature-Requests/issues/185)

The # 185 bit on the commit list was unintentional, I did not know it would try to link to something (especially the wrong thing).


src/main/java/mekanism/common/content/gear/mekasuit/ModuleMekaSuit.java
```diff
@@ -398,4 +405,20 @@ public void addHUDElements(List<HUDElement> list) {
...
     }
+    
+    public static class ModuleGeigerUnit extends ModuleMekaSuit {
+        private static final ResourceLocation icon = MekanismUtils.getResource(ResourceType.GUI_HUD, "geiger_counter.png");
+        @Override
+        public void addHUDElements(List<HUDElement> list) {
+            if (!isEnabled()) {
+                return;
+            }
+            else {
```
pupnewfster, line 416: Having the explicit else here isn't really needed.

pupnewfster (commented):
A recipe should probably also be added for this module, probably can just add another set of this: https://github.com/mekanism/Mekanism/blob/v10.1/src/datagen/main/java/mekanism/common/recipe/impl/MekanismRecipeProvider.java#L1206-L1216 to our provider and switch out the dosimeter in it for a geiger counter.

beholderface:
I had a recipe while I was testing this but I guess it didn't occur to me that it was something that had to be done through datagen

src/main/java/mekanism/common/content/gear/mekasuit/ModuleMekaSuit.java
```diff
@@ -398,4 +404,19 @@ public void addHUDElements(List<HUDElement> list) {
...
+        @Override
+        public void addHUDElements(List<HUDElement> list) {
+            if (!isEnabled()) {
+                return;
+            }
+            {
+                double magnitude = MekanismAPI.getRadiationManager().getRadiationLevel(Minecraft.getInstance().player);HUDElement e = HUDElement.of(icon, UnitDisplayUtils.getDisplayShort(magnitude, RadiationUnit.SV, 3));
+                HUDColor color = magnitude < RadiationManager.MIN_MAGNITUDE ? HUDColor.REGULAR : (magnitude < 0.1 ? HUDColor.WARNING : HUDColor.DANGER);
+                list.add(e.color(color));
+            }
```
pupnewfster, line 419: When I said the else isn't needed I also meant the brackets that go with it aren't needed either. Also `HUDElement e` should be on the next line and not sharing the same line with `double magnitude`

src/main/java/mekanism/common/content/gear/mekasuit/ModuleMekaSuit.java
```diff
@@ -398,4 +404,18 @@ public void addHUDElements(List<HUDElement> list) {
...
+    
+    public static class ModuleGeigerUnit extends ModuleMekaSuit {
+        private static final ResourceLocation icon = MekanismUtils.getResource(ResourceType.GUI_HUD, "geiger_counter.png");
+        @Override
+        public void addHUDElements(List<HUDElement> list) {
+            if (!isEnabled()) {
+                return;
+            }
+                double magnitude = MekanismAPI.getRadiationManager().getRadiationLevel(Minecraft.getInstance().player);
+                HUDElement e = HUDElement.of(icon, UnitDisplayUtils.getDisplayShort(magnitude, RadiationUnit.SV, 3));
```
pupnewfster, line 416: Overall looks good and seems to work quite well so thank you for the PR. One last minor thing though is that this `, 3` should be changed to a `, 2` to match how we show the dosimeter's level on the HUD as the extra decimal point of precision isn't needed on the HUD.

Seniorendi:
Nice work!

PR #7241, opened 2021-06-18
No summon robit

heipiao233:


pupnewfster:
Why?

heipiao233:
If not, when we summoning it, Minecraft will crash at both side!

pupnewfster:
Oh right so this controls the `/summon` command (which spawns the entity).I was thinking this was controlling `tp` for some reason (probably due to how much stuff I used to do with bukkit and there being an overload command called `summon` to teleport a specific thing to you).

PR #7250, opened 2021-06-27
Avoid VisionEnhancement night vision flashing

milomai:
Change VisionEnhancement night vision effect to 15s (300 tick) to avoid flashing.


pupnewfster:
Hmm looking at it apparently it is when there is 10 or less seconds left of Night Vision it flickers, so setting it to say 11 seconds may make more sense than 15, probably would also help to adjust the comment the line above to mention that it is 220 not 200, and is set to that to make sure that it doesn't cause any blinking.

milomai:
Fine, I fixed it.

PR #7379, opened 2022-01-02
[1.16.x] Nameable tiles 

### LINKED ISSUE #6371: Personal chest naming does not save when placed and picked-up
#### Issue description:
When a personal chest is named from an anvil and placed, once it is picked up, the item loses its custom name. This problem does not occur when using normal Minecraft chests, for reference.

#### Steps to reproduce:
1. Name a personal chest using an anvil.
2. Place the personal chest.
3. Pick up the personal chest.
4. The name should be reverted to "Personal chest".

#### Version: 1.16.1
**Forge: 1.16.1-forge-32.0.69** 
**Mekanism: Mekanism-1.16.1-10.0.2.425**

Matyrobbrt:
This PR adds support for tile entities to have custom names (through anvil remaining), providing an `isNameable` method which can be implemented in order to not let a tile have a name. (true by default, since this behaviour is wanted on most tile entities)
This also means that the loot tables need to be changed in order to have the `copy_name` function. I changed the data generation code in order to support that, but when I try to runData I get some class not found exceptions related to ProjectE. I will look into this issue and try to fix it  tommorow.

If you think that other tile entities should have naming disabled let me know so that I can do it. 

This PR also fixes #6371


Matyrobbrt:
I just noticed the fact that my IDE replaced a few imports with wildcard ones. I will fix that tomorrow

pupnewfster:
I am going to close this and focus on #7398 instead as I would rather not deal with merging in and fixing all the issues with it in 1.16 so close to the 1.16 branch being EOL (as it is EOL except for a few fixes and one more merge of translations).

PR #7389, opened 2022-01-10
Update zh_cn.json

horeaper:
fix to a wrong translation.


toddlersizedsharkcage:
https://crowdin.com/project/mekanism is used for translations. Please contribute there if you like. These are not done with pull requests.

PR #7398, opened 2022-01-22
1.18.x nameable tiles

### LINKED ISSUE #6371: Personal chest naming does not save when placed and picked-up

Matyrobbrt:
1.18 version of #7379
Fixes #6371


src/main/java/mekanism/common/tile/base/TileEntityMekanism.java
```diff
@@ -738,6 +783,10 @@ public CompoundTag getReducedUpdateTag() {
             component.addToUpdateTag(updateTag);
         }
         updateTag.putFloat(NBTConstants.RADIATION, radiationScale);
+        // Sync the name with clients
+        if (this.customName != null && isNameable()) {
+            updateTag.putString(NBTConstants.CUSTOM_NAME, Component.Serializer.toJson(this.customName));
+        }
```
pupnewfster, line 789: What is the use of syncing this stuff in the update tag? (Especially after the changes I made yesterday). Is it just so if a mod calls the methods on the client they have results, given I am honestly not quite sure how much that matters.

src/main/java/mekanism/common/tile/base/TileEntityMekanism.java
```diff
@@ -431,8 +438,38 @@ public void addComponent(ITileComponent component) {
...
     public Component getName() {
-        return TextComponentUtil.translate(Util.makeDescriptionId("container", getBlockType().getRegistryName()));
+        return hasCustomName() ? getCustomName() : TextComponentUtil.build(getBlockType());
+    }
+
+    @Nonnull
+    @Override
+    @SuppressWarnings("ConstantConditions")
+    public Component getDisplayName() {
+        return hasCustomName() ? getCustomName() : TextComponentUtil.translate(Util.makeDescriptionId("container", getBlockType().getRegistryName()));
```
pupnewfster, line 451: I added this as a different implementation than getName just for more "accurate" results at least for purposes as we use display name for the container, but I am wondering if this should maybe instead be:
```java
if (isNameable()) {
//current return
}
return TextComponentUtil.build(getBlockType());
```
Thoughts? Alternatively maybe move the logic of this return to AttributeGUI and remove the getDisplayName override allowing it to be the default in case mods make assumptions and want the actual name not the name for displaying in containers.

src/main/java/mekanism/common/tile/TileEntityBoundingBlock.java
```diff
@@ -27,7 +30,7 @@
 /**
  * Multi-block used by wind turbines, solar panels, and other machines
  */
-public class TileEntityBoundingBlock extends TileEntityUpdateable implements IUpgradeTile, ISecurityTile {
+public class TileEntityBoundingBlock extends TileEntityUpdateable implements IUpgradeTile, ISecurityTile, Nameable {
```
pupnewfster, line 33: Is there really much use in making the bounding blocks be Nameable, or is this just in case other mods might want to query information via them? I suspect at least for Jade and TOP this may not be needed due to redirects I already do, but if you think it is worth leaving them in we can.

src/main/java/mekanism/common/tile/base/TileEntityMekanism.java
Matyrobbrt: I think that we should wrap that return inside the `isNameable`, because if it isn't showing the custom name is useless.

src/main/java/mekanism/common/tile/base/TileEntityMekanism.java
Matyrobbrt: I'm not sure if the syncing is needed anymore. Let me see 🤔

src/main/java/mekanism/common/tile/TileEntityBoundingBlock.java
Matyrobbrt: From my testing, Jade needed that in order to pickup the custom name, if it exists. I'd leave it there in case any other mod needs the block's name for whatever reason.

src/main/java/mekanism/common/tile/base/TileEntityMekanism.java
```diff
@@ -431,8 +438,46 @@ public void addComponent(ITileComponent component) {
...
     @Nonnull
+    @Override
+    @SuppressWarnings("ConstantConditions")
     public Component getName() {
-        return TextComponentUtil.translate(Util.makeDescriptionId("container", getBlockType().getRegistryName()));
+        final var defaultName = TextComponentUtil.build(getBlockType());
+        if (isNameable()) {
+            return hasCustomName() ? getCustomName() : defaultName;
+        }
+        return defaultName;
```
pupnewfster, line 448: Given getCustomName returns null if it isn't nameable, this can stay as it was before `return hasCustomName() ? getCustomName() : TextComponentUtil.build(getBlockType());`

src/main/java/mekanism/common/tile/base/TileEntityMekanism.java
```diff
@@ -431,8 +438,46 @@ public void addComponent(ITileComponent component) {
...
+    }
+
+    @Nonnull
+    @Override
+    @SuppressWarnings("ConstantConditions")
+    public Component getDisplayName() {
+        final var defaultName = TextComponentUtil.translate(Util.makeDescriptionId("container", getBlockType().getRegistryName()));
```
pupnewfster, line 455: Two things:
1. For now I think I am going to hold off in a big way from using var in Mekanism
2. The reason I was mentioning wrapping in isNameable, is not all tiles have this translation key now. Which is why I was saying to use a different fallback `TextComponentUtil.build(getBlockType())` for if it is not nameable, but use this one for if it is nameable and just doesn't have a custom name.

PR #7430, opened 2022-03-19
Add instructions on building and moving over ProjectE files

JustinHuPrime:

The build documentation is updated with instructions on building and copying over ProjectE files. At the moment, this is the only part of the build process that must be done manually and not through Gradle.

I would have found it helpful if these instructions were available when I tried to build Mekanism.


pupnewfster:
The reason I decided against updating the docs is 1.18 was not really ready to be built so I didn't want to just add it and then forget to remove it. I will leave this PR open for now just so people can see it easier, but given it hopefully will only be like a week or so more that this is needed I don't see much of a point in merging it just to remove it shortly afterward.

Issue #3907, opened 2016-10-19
[suggestion] Nitrogen chemistry for fertiliser and explosives


momerathe:
Chemical production lines are cool. Here are a couple of suggestions for some new ones.

Fertiliser:

For the source of nitrogen. I tried to think of how to do this without a new machine, but failed. 
Step 1: "Gas seperator" converts air into Nitrogen & oxygen
Step 2: Nitrogen + hydrogen in a chemical infuser -> Ammonia gas
Step 3: Ammonia gas in a rotary condensentrator -> Liquid Ammonia
Urea (a common fertiliser) is created by reacting ammonia with carbon dioxide. I think this can be finessed to work with the Pressurized Reaction Chamber.
Step 4: Liquid Ammonia + coal (carbon) + oxygen in a PRC -> "Chemical fertiliser" + water vapour.

"Chemical fertiliser" obviously works like bonemeal for crops, but can't be used for dyeing.

Explosives:

Same as before up to ammonia gas:
Step 3: Ammonia gas + oxygen in a chemical infuser -> Nitric oxide
Step 4: Nitric oxide + oxygen in a chemical infuser -> Nitrogen dioxide
The next step is really two steps in one, creating nitric acid with water and then nitrating an organic compound.
Step 5: Nitrogen dioxide + Water + Substrate in PRC -> explosive pellet + nitric oxide

explosives could be oredict'd with gunpowder, or used for later recipes.

DovahOfKiin:
I'm actually trying to begin work on a mod that adds nitrogen related chemistry (and depends on Mekanism) but because of this issue, I cannot: https://github.com/aidancbrady/Mekanism/issues/3896

PR #7432, opened 2022-03-23
Mekanism: Chemistry Module

JustinHuPrime:
This is a proposed module for Mekanism involving industrial chemical processes. Currently, features are being implemented to fulfill #3907 (which hasn't seen any action for a few years, so I figure the original attempt was abandoned). This might tie in to [greenhouses and farming](https://github.com/mekanism/Mekanism/projects/2#card-21409126) if the greenhouse or farming machine requires fertilizer produced by this module. I've got this as a draft for now; help working on the fractionating distiller multiblock would be appreciated.

## Changes Made

- A new block (model and texture not yet implemented) called the Air Compressor. Consumes energy, producing the gas air, similar to an electric pump.
- A new multiblock (texture and gui not yet implemented) called the Fractionating Distiller. Requires low levels of heating and liquid air, producing liquid oxygen and liquid nitrogen.
- A new chemical infuser recipe combining one unit nitrogen and one unit hydrogen into one unit of ammonia (might want to change this to 3:1 hydrogen:nitrogen)
- A new chemical crystallizer recipe crystallizing ammonia into ammonium (might change this to be a PRC reaction w/ liquid ammonia + oxygen + coal/charcoal in a PRC)
- A new shapeless crafting recipe converting ammonium and sulfur dust into fertilizer
- Condensentrator recipes for nitrogen and ammonia.

## Changes To Be Made

- The multiblock's GUI and actual functionality are not complete - I think there's bugs in how I implement the GUI, the internal fluid tanks, and recipe operations.
- None of the art is complete; I've got obvious programmer art in place for now, but I don't think I could match the existing art style at all.
- Nitrogen-based explosives need to be implemented - I think this is pretty quick; maybe an alternate crafting recipe for TNT using ammonia or something like that.
- Greenhouses/industrial farms need to be implemented - maybe as multiblock farms similar to Forestry, but that could also be its own project in and of itself.
- Petrochemical processing would be a good long-term goal.


pupnewfster:
I will try to review this a little closer and try to answer your questions/comments in more detail, especially pertaining to the gui, fluid tanks, and recipe operations. But from a brief glance through things, it looks like in general the correct systems of ours are being used though which is definitely a good start.

JustinHuPrime:
I've added recipes for the machines, as well as a production chain from nitrogen to TNT. Currently, I think the actual fractionating distiller itself and maybe the greenhouse need to be implemented and art assets added before I'd be happy with making this PR ready for review.

PR #7433, opened 2022-03-26
Dimensional stabilizer

JustinHuPrime:
This PR implements a dedicated chunkloader block for the base Mekanism module (see [this feature request](https://github.com/mekanism/Mekanism/projects/1#card-3896069)).

## Changes Made
- A new block called the Dimensional Stabilizer (model and texture not yet implemented). Consumes energy (1 kJ/t) and chunkloads up to a 5x5 area centered on the block, with toggles for each chunk.
- Minor changes to the Electrolytic Separator's EnergyContainer to reduce duplicated code

## Changes To Be Made
- I can't get the state of the per-chunk loading toggles to sync - I looked at the Rotary Condensentrator and tried to copy what it's doing for its toggle.
- This lack of sync seems to extend to saving - I may have an obvious bug with how I write my NBT data.
- As before, I only have programmer art.
- Currently many of the subfeatures requested (visual area of effect indicator, player detector, max owned per player, and configurable maximum range) are not implemented. Then again, I think it's useable as-is if the existing bugs get fixed.
- I haven't gotten around to adding in the calculation for per-chunk energy usage, but that should be a simple fix once I get around to it.
- The GUI could stand some improvement; labels and such on the toggles or axes to indicate which side is north and what chunk the toggle corresponds to
- It would also be nice if the buttons didn't use the generic switch and instead used some sort of on/off colour change, similar to side configuration.


pupnewfster:
I kind of held off from implementing a chunk loader as I had some ideas of making a multiblock chunk loader that uses resources and the shape/configuration of the multiblock determines which chunks the chunk loader loads, but given it doesn't seem like I will get to this any time soon it seems like this may be a reasonable thing to implement.

In terms of
> I can't get the state of the per-chunk loading toggles to sync - I looked at the Rotary Condensentrator and tried to copy what it's doing for its toggle.
This lack of sync seems to extend to saving - I may have an obvious bug with how I write my NBT data.

I believe saving might be working properly but it is the sync that isn't working so when you re-open it after reloading your game it looks like it isn't updated. The issue is that you aren't currently syncing the data from the server to the client, and you need to add something like this to the tile class:
```java
@Override
public void addContainerTrackers(MekanismContainer container) {
    super.addContainerTrackers(container);
    container.trackArray(loadingChunks);
}
```
and then in `MekanismContainer` by the existing trackArray stuff add another helper that can handle `boolean[][]`. You can do it in the same way that we do it [here](https://github.com/mekanism/Mekanism/blob/1.18.x/src/main/java/mekanism/common/tile/factory/TileEntityFactory.java#L763-L769) and then remove that private helper from the factory and just make it call the helper that you add in `MekanismContainer`

JustinHuPrime:
That seems to have fixed syncing, and I've made energy per-chunk-loaded-per-tick.

I've gone ahead and implemented trackArray for two-level boolean arrays, and changed the Factory stuff to use the two-level array helper directly.

I think this is the todo-list for this PR:
- [ ] Actual art assets
- [ ] Nicer GUI, maybe showing which chunk goes with which toggle, or direction labels
- [ ] Green/grey toggle buttons

JustinHuPrime:
Okay - I think the only items remaining are art assets

src/main/java/mekanism/client/gui/element/button/ColorToggleButton.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public ColorToggleButton(IGuiWrapper gui, int x, int y, int size, EnumColor color, BooleanSupplier toggled, Runnable onLeftClick, @Nullable GuiElement.IHoverable onHover) {
+        super(gui, x, y, size, size, TextComponent.EMPTY, onLeftClick, onHover);
+        this.toggled = toggled;
+        this.color = color;
+    }
+
+    @Override
+    public void drawBackground(@Nonnull PoseStack matrix, int mouseX, int mouseY, float partialTicks) {
+        if (toggled.getAsBoolean()) {
```
pupnewfster, line 33: This value should be stored in a local variable so as to not have to potentially calculate the supplier twice in one draw call if it has more logic than a simple true/false getter.

src/main/java/mekanism/common/tile/TileEntityDimensionalStabilizer.java
```diff
@@ -0,0 +1,211 @@
...
+            } else {
+                setChunkloading(false);
+            }
+        }
+    }
+
+    private void setChunkloading(boolean value) {
+        boolean old = chunkloading;
+        chunkloading = value;
+        if (old != chunkloading) {
```
pupnewfster, line 106:
```java
if (chunkloading != value) {
    chunkloading = value;
```

src/main/java/mekanism/common/tile/factory/TileEntityFactory.java
```diff
@@ -760,11 +760,9 @@ public ErrorTracker(Map<RecipeError, Boolean> errorTypes, int processes) {
...
-                //Note: We don't need a special syncable boolean for two depth arrays
-                // as long as our arrays are in the same order on client and server
-                for (int j = 0; j < processTrackedErrors.length; j++) {
-                    container.track(SyncableBoolean.create(processTrackedErrors, j));
+            for (int i = 0; i < trackedErrors.length; i++) {
+                for (int j = 0; j < trackedErrors[i].length; j++) {
+                    container.track(SyncableBoolean.create(trackedErrors, i, j));
                 }
             }
         }
```
pupnewfster, line 768: Is there a reason this method still exists instead of being replaced with a call to `container.trackArray(trackedErrors)`

src/main/java/mekanism/common/tile/TileEntityDimensionalStabilizer.java
```diff
@@ -0,0 +1,211 @@
...
+        loadingChunks = new boolean[MAX_LOAD_DIAMETER][MAX_LOAD_DIAMETER];
+        loadingChunks[MAX_LOAD_RADIUS][MAX_LOAD_RADIUS] = true;
+    }
+
+    public boolean isChunkloadingAt(int x, int z) {
+        return loadingChunks[x][z];
+    }
+
+    public void toggleChunkloadingAt(int x, int z) {
+        loadingChunks[x][z] = !loadingChunks[x][z];
```
pupnewfster, line 62: Given this method is called via a packet, we should validate the `x` and `z` values are actually within the proper ranges so we don't let modified clients crash the server with an IOBE by sending invalid values,

src/main/java/mekanism/common/tile/TileEntityDimensionalStabilizer.java
```diff
@@ -0,0 +1,211 @@
...
+
+    private void setChunkloading(boolean value) {
+        boolean old = chunkloading;
+        chunkloading = value;
+        if (old != chunkloading) {
+            setChanged(true);
+        }
+    }
+
+    private int getChunksLoaded() {
```
pupnewfster, line 111: This method might also be worth exposing to computers, though I wonder if the one exposed to them should be wrapped basically as:
```java
return chunkLoading ? getChunksLoaded() : 0;
```

pupnewfster (commented):
I haven't actually pulled this locally to look at it in an IDE yet so once I do I may have more comments, but from a glance it is looking pretty good, just a few comments here and there.

pupnewfster:
It looks like the merge/rebase on this PR went a bit crazy like the free runners one initially did.

Issue #5221, opened 2018-12-18
Armored Version of the Free Runners


ToMe25:
The Free Runners are a great way to stop getting Falldamage, but it would be nice if they wouldn't reduce my protection by don't having a Armor value.

So Please add a Armored Version of the Free Runners.

PR #7434, opened 2022-03-26
Armored Free Runners

JustinHuPrime:
This small PR implements the armoured free runners from the legacy enhancements list (specifically #5221).

## Changes Made
- Armored free runners are a version of the free runners, like the armored jetpack is to the jetpack.
- Has similar crafting recipe as Armored Jetpack
- Has diamond armour stats.

## Changes To Be Made
- Missing model and art assets; currently re-uses same art assets as plain free runners


src/main/java/mekanism/common/config/GearConfig.java
```diff
@@ -21,8 +21,9 @@
     private static final String ENERGY_TABLET_CATEGORY = "energy_tablet";
     private static final String FLAMETHROWER_CATEGORY = "flamethrower";
     private static final String FREE_RUNNER_CATEGORY = "free_runner";
+    private static final String ARMORED_FREE_RUNNER_SUBCATEGORY = "armored_free_runner";
     private static final String JETPACK_CATEGORY = "jetpack";
-    private static final String ARMORED_JETPACK_SUBCATEGORY = "armored";
+    private static final String ARMORED_JETPACK_SUBCATEGORY = "armored_jetpack";
```
pupnewfster, line 26: This doesn't need to be changed as it is a subcategory to the jetpack category already so it basically is currently `jetpack.armored`

pupnewfster: Though I definitely think it may be worth renaming the constant to `ARMORED_SUBCATEGORY` and then just using it for both the free runner one and the jetpack one.

JustinHuPrime: Ah, right. Fixed.

pupnewfster:
I will try to look into this and your other PRs in more detail after we get the initial 1.18 release out, but if you could rebase them to 1.18.2 now that datagen can be ran again easily in 1.18.2 that would be great!

pupnewfster:
Something seems to have gone weird with your merge/rebase as it is now showing all the changes made upstream in the diff.

JustinHuPrime:
Right - does that fix things?

pupnewfster:
From a glance it looks like it did, I will look closer next week as that should be after I get a release out.

pupnewfster (commented):
Looks good for the most part (I left a couple tiny comments). I reached out to one of our artists regarding potentially making a model that looks slightly more armored. One other thing I will need to discuss with the other devs is whether the recipe makes sense or if it should include some amount of steel similar to how the armored jetpack requires a steel block (I am currently undecided given I do feel like a full steel block may be overkill, but maybe we want to make it take a couple steel ingots; not sure yet).

PR #7435, opened 2022-03-26
Add creative laser block

JustinHuPrime:
Minor change. Adds a creative laser version of a regular laser. It's 1000x more powerful than a regular laser and doesn't require energy. I've found it to be useful for testing fusion reactors.

Partially addresses https://github.com/mekanism/Mekanism-Feature-Requests/issues/322

Might be useful to add a gui to adjust laser power up or down, but the primary use of lasers at the moment is to start up the fusion reactor.


pupnewfster:
I honestly don't really see much of a point in this given as of Mekanism 10.0.0 you have been able to right click the laser focus matrix with a configurator while in creative to activate the fusion reactor without having to go through a laser if you have access to creative.

JustinHuPrime:
I, ah, wasn't aware of that. Closed with much embarassment.

PR #7446, opened 2022-04-11
[1.18.x] Curio jetpack

Matyrobbrt:
This PR makes the jetpack work in a curio slot.
**The jetpack is not added by default to any curio slot, that is something a pack author has to do, if they want to enable that functionality.**


src/main/java/mekanism/client/render/MekanismCurioRenderer.java
```diff
@@ -0,0 +1,24 @@
...
+import net.minecraft.client.renderer.texture.OverlayTexture;
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.item.ItemStack;
+import top.theillusivec4.curios.api.SlotContext;
+import top.theillusivec4.curios.api.client.ICurioRenderer;
+
+public record MekanismCurioRenderer(ICustomArmor model) implements ICurioRenderer {
+
+	@Override
+	public <T extends LivingEntity, M extends EntityModel<T>> void render(ItemStack stack, SlotContext slotContext, PoseStack matrixStack, RenderLayerParent<T, M> renderLayerParent, MultiBufferSource renderTypeBuffer, int light, float limbSwing, float limbSwingAmount, float partialTicks, float ageInTicks, float netHeadYaw, float headPitch) {
```
pupnewfster, line 18: Switch to spaces from tabs

src/main/java/mekanism/client/render/hud/MekaSuitHUD.java
```diff
@@ -31,28 +34,46 @@ public void render(ForgeIngameGui gui, PoseStack poseStack, float partialTicks,
         Minecraft minecraft = Minecraft.getInstance();
         if (!minecraft.options.hideGui && !minecraft.player.isSpectator() && MekanismConfig.client.enableHUD.get() && MekanismClient.renderHUD) {
             int count = 0;
-            Map<EquipmentSlot, List<Component>> renderStrings = new LinkedHashMap<>();
+            final List<List<Component>> renderStrings = new LinkedList<>();
```
pupnewfster, line 37: Any reason for using a LinkedList over just using an ArrayList? And at this point it probably should just be a `List<Component>` with using `addAll(list)` below rather than `add(list)`.

src/main/java/mekanism/client/render/hud/MekaSuitHUD.java
```diff
@@ -31,28 +34,46 @@ public void render(ForgeIngameGui gui, PoseStack poseStack, float partialTicks,
...
+                        ItemStack stack = inv.getStackInSlot(i);
+                        if (stack.getItem() instanceof IItemHUDProvider hudProvider) {
+                            List<Component> list = new ArrayList<>();
+                            hudProvider.addCurioHUDStrings(list, minecraft.player, stack);
+                            int size = list.size();
+                            if (size > 0) {
+                                renderStrings.add(list);
+                                count += size;
+                            }
+                        }
```
pupnewfster, line 64: This could just be a helper method that returns the size to increment count by and then reduces the duplicate code between this and the normal armor spots.

src/datagen/main/java/mekanism/client/lang/MekanismLangProvider.java
```diff
@@ -666,7 +666,8 @@ private void addMisc() {
         add(MekanismLang.PROCESS_RATE_MB, "Process Rate: %1$s mB/t");
         add(MekanismLang.TICKS_REQUIRED, "Ticks Required: %1$s");
         add(MekanismLang.DECAY_IMMUNE, "Will not decay inside a Radioactive Waste Barrel");
-        //Gui stuff
+        add(MekanismLang.CURIO_NOT_USABLE, "This item is usable as a curio, but it has not been added to a Curios tag");
+		//Gui stuff
```
pupnewfster, line 670: Weird random indentation

src/main/java/mekanism/common/CommonPlayerTickHandler.java
```diff
@@ -179,21 +181,22 @@ public static boolean handleJetpackMotion(Player player, JetpackMode mode, Boole
         return true;
     }
 
-    private static boolean isJetpackOn(Player player, ItemStack chest) {
+    private static ItemStack getJetpackIfOn(Player player) {
+        final ItemStack chest = ItemJetpack.getJetpack(player);
```
pupnewfster, line 185: Why the change to this? The idea of it taking the chest stack as a parameter was to not have to relook up the stored item.

src/main/java/mekanism/common/integration/MekanismHooks.java
```diff
@@ -44,12 +42,15 @@ public final class MekanismHooks {
     public void hookConstructor() {
         ModList modList = ModList.get();
         CraftTweakerLoaded = modList.isLoaded(CRAFTTWEAKER_MOD_ID);
+        CuriosLoaded = modList.isLoaded(CURIOS_MODID);
+        if (CuriosLoaded) {
+            CuriosIntegration.addListeners(FMLJavaModLoadingContext.get().getModEventBus());
```
pupnewfster, line 47: If you are adding events like this might as well just modify `hookConstructor` to accept the mod event bus as a parameter.

src/main/java/mekanism/common/integration/curios/CuriosIntegration.java
```diff
@@ -0,0 +1,46 @@
...
+import top.theillusivec4.curios.api.SlotTypePreset;
+import top.theillusivec4.curios.api.client.CuriosRendererRegistry;
+
+import javax.annotation.Nonnull;
+import java.util.Optional;
+import java.util.function.Predicate;
+
+public class CuriosIntegration {
+
+	public static void sendIMC() {
```
pupnewfster, line 24: Tabs to spaces

src/main/java/mekanism/common/item/gear/ItemJetpack.java
```diff
@@ -21,156 +22,190 @@
...
-    @ParametersAreNonnullByDefault
-    @MethodsReturnNonnullByDefault
-    protected static class JetpackMaterial extends BaseSpecialArmorMaterial {
-
-        @Override
-        public String getName() {
-            return Mekanism.MODID + ":jetpack";
-        }
-    }
+	private static final JetpackMaterial JETPACK_MATERIAL = new JetpackMaterial();
```
pupnewfster, line 51: Remove all these formatting changes of making spaces be tabs

src/main/java/mekanism/common/item/interfaces/IItemHUDProvider.java
```diff
@@ -9,4 +9,8 @@
 public interface IItemHUDProvider {
 
     void addHUDStrings(List<Component> list, Player player, ItemStack stack, EquipmentSlot slotType);
+
+    default void addCurioHUDStrings(List<Component> list, Player player, ItemStack stack) {
+
```
pupnewfster, line 14: Blank line for the body can be removed (keep the closing brace on its own line though)

src/main/java/mekanism/client/render/hud/MekaSuitHUD.java
Matyrobbrt: I thought of that as well, but each component list needs to have a different amount of space between 2 components from the same list (9 vs 2)

pupnewfster: Ahh missed that part because I didn't expand github's view and missed seeing there is an extra spacer between groups.

src/main/java/mekanism/common/CommonPlayerTickHandler.java
```diff
@@ -179,21 +182,24 @@ public static boolean handleJetpackMotion(Player player, JetpackMode mode, Boole
         return true;
     }
 
-    private static boolean isJetpackOn(Player player, ItemStack chest) {
+    private static ItemStack getJetpackIfOn(ItemStack chest, Player player) {
```
pupnewfster, line 185: No reason to reorder the parameters here.

src/main/java/mekanism/client/render/hud/MekaSuitHUD.java
```diff
@@ -2,12 +2,15 @@
 
 import com.mojang.blaze3d.vertex.PoseStack;
 import java.util.ArrayList;
-import java.util.LinkedHashMap;
+import java.util.LinkedList;
```
pupnewfster, line 5: Unused import

src/main/java/mekanism/client/render/hud/MekaSuitHUD.java
```diff
@@ -51,8 +61,8 @@ public void render(ForgeIngameGui gui, PoseStack poseStack, float partialTicks,
                 int yScale = (int) ((1 / hudScale) * height);
                 poseStack.pushPose();
                 poseStack.scale(hudScale, hudScale, hudScale);
-                for (Map.Entry<EquipmentSlot, List<Component>> entry : renderStrings.entrySet()) {
-                    for (Component text : entry.getValue()) {
+                for (List<Component> entry : renderStrings) {
```
pupnewfster, line 64: Potentially rename entry to group or something?

src/main/java/mekanism/client/render/hud/MekaSuitHUD.java
```diff
@@ -66,6 +76,17 @@ public void render(ForgeIngameGui gui, PoseStack poseStack, float partialTicks,
         }
     }
 
+    private int makeComponent(Consumer<List<Component>> adder, List<List<Component>> initial) {
+        List<Component> list = new ArrayList<>();
+        adder.accept(list);
+        int size = list.size();
+        if (size > 0) {
+            initial.add(list);
+            return size;
```
pupnewfster, line 85: Remove this extra return statement and just change the `return 0;` at the bottom to `return size;`

src/main/java/mekanism/common/CommonPlayerTickHandler.java
```diff
@@ -179,21 +182,24 @@ public static boolean handleJetpackMotion(Player player, JetpackMode mode, Boole
...
+        if (chest.isEmpty() && Mekanism.hooks.CuriosLoaded) {
+            chest = CuriosIntegration.findFirstCurio(player, s -> s.getItem() instanceof ItemJetpack).orElse(ItemStack.EMPTY);
```
pupnewfster, line 187: I am unsure whether or not we should only be doing this if the chest stack is empty or should it be when it isn't a jetpack or something?  Either way I know the findFirstCurio is to handle there potentially being multiple jetpacks in the curio slots, but do we want to have better handling if say two jetpacks are equipped but only one has hydrogen that we then find the one that has hydrogen?

src/main/java/mekanism/common/Mekanism.java
```diff
@@ -249,7 +249,7 @@ public Mekanism() {
         versionNumber = new Version(ModLoadingContext.get().getActiveContainer());
         packetHandler = new PacketHandler();
         //Super early hooks, only reliable thing is for checking dependencies that we declare we are after
-        hooks.hookConstructor();
+        hooks.hookConstructor(FMLJavaModLoadingContext.get().getModEventBus());
```
pupnewfster, line 252: Just pass `modEventBus`, we already have it in a local variable.

src/main/java/mekanism/common/MekanismLang.java
```diff
@@ -177,6 +177,8 @@ public enum MekanismLang implements ILangEntry {
     FLAMETHROWER_STORED("tooltip", "stored.flamethrower"),
     JETPACK_STORED("tooltip", "stored.jetpack"),
     DECAY_IMMUNE("tooltip", "decay_immune"),
+    CURIO_NOT_USABLE("tooltip", "curio_not_usable"),
+
```
pupnewfster, line 181: Extra blank line can probably be removed

src/main/java/mekanism/common/integration/curios/CuriosIntegration.java
```diff
@@ -0,0 +1,46 @@
...
+            CuriosRendererRegistry.register(MekanismItems.ARMORED_JETPACK.asItem(), () -> new MekanismCurioRenderer(JetpackArmor.ARMORED_JETPACK));
+        });
+    }
+
+    public static Optional<? extends IItemHandler> getCuriosInventory(LivingEntity living) {
+        return CuriosApi.getCuriosHelper().getEquippedCurios(living).resolve();
+    }
+
+    public static Optional<ItemStack> findFirstCurio(@Nonnull LivingEntity livingEntity, Predicate<ItemStack> filter) {
+        return CuriosApi.getCuriosHelper().findFirstCurio(livingEntity, filter).map(SlotResult::stack);
```
pupnewfster, line 40: `return findFirstCurioAsResult(livingEntity, filter).map(SlotResult::stack);`

src/main/java/mekanism/common/item/gear/ItemGasArmor.java
```diff
@@ -13,6 +13,7 @@
 import mekanism.common.item.interfaces.IGasItem;
 import mekanism.common.util.ChemicalUtil;
 import mekanism.common.util.StorageUtils;
+import net.minecraft.client.gui.screens.Screen;
```
pupnewfster, line 16: ??

src/main/java/mekanism/common/item/gear/ItemJetpack.java
```diff
@@ -89,18 +98,27 @@ public void setMode(ItemStack stack, JetpackMode mode) {
...
-                if (gasHandlerItem.getTanks() > 0) {
-                    stored = gasHandlerItem.getChemicalInTank(0);
-                }
+            addHUD(list, stack);
+        }
+    }
+
+    @Override
+    public void addCurioHUDStrings(List<Component> list, Player player, ItemStack stack) {
+        addHUD(list, stack);
```
pupnewfster, line 107: It would be cleaner (especially from a diff perspective) to just leave `addHUDStrings` alone and then make the implementation of this method just call `addHUDStrings(list, player, stack, getSlot())`

src/main/java/mekanism/common/CommonPlayerTickHandler.java
Matyrobbrt: Oh ups.

src/main/java/mekanism/common/item/gear/ItemGasArmor.java
Matyrobbrt: Ups, I thought I changed the use of `Scree#hasShiftDown` to `TooltipFlag#isAdvanced`. Fixed.

src/main/java/mekanism/common/CommonPlayerTickHandler.java
Matyrobbrt: That's a fair point. It should get the first jetpack with hydrogen. That will probably need fixing in the key binding to add that "check" as well.

build.gradle
```diff
@@ -423,6 +424,7 @@ dependencies {
     //TODO - 1.18: Support flux networks again when it ports
     //compileOnly fg.deobf("curse.maven:flux-networks-248020:${flux_networks_id}")
     compileOnly fg.deobf("top.theillusivec4.curios:curios-forge:${minecraft_version}-${curios_version}:api")
+    // runtimeOnly fg.deobf("top.theillusivec4.curios:curios-forge:${minecraft_version}-${curios_version}")
```
pupnewfster, line 427: Unneeded easy enough to add if/when we need to test in dev

gradle.properties
```diff
@@ -44,6 +44,7 @@ wildfire_gender_mod_id=3706808
 crafttweaker_version_range=[9.1.90,)
 jei_version_range=[9.7.0.180,)
 wildfire_gender_version_range=[2.9,)
+curios_version_range=[1.18.2-5.0.6.3,)
```
pupnewfster, line 47: This should go between the crafttweaker and JEI version ranges as this block is alphabetical

src/main/java/mekanism/common/item/gear/ItemJetpack.java
```diff
@@ -128,6 +143,31 @@ public int getDefaultTooltipHideFlags(@Nonnull ItemStack stack) {
...
+     * Gets the first found jetpack from an entity, if one is worn. <br>
+     * If Curios is loaded, the curio slots will be checked as well.
+     *
+     * @param entity the entity on which to look for the jetpack
+     * @return the jetpack stack if present, otherwise an empty stack
+     */
+    @Nonnull
+    public static ItemStack getJetpack(final LivingEntity entity) {
+        final ItemStack chest = entity.getItemBySlot(EquipmentSlot.CHEST);
+        if (chest.getItem() instanceof ItemJetpack) return chest;
```
pupnewfster, line 156: This should have brackets and then to keep it shorter you can just do `} else if (Mekanism.hooks.CuriosLoaded) {`

src/main/java/mekanism/common/item/gear/ItemGasArmor.java
pupnewfster, line 16: Remove this import?

src/main/java/mekanism/common/item/gear/ItemJetpack.java
```diff
@@ -76,6 +83,9 @@ protected IGasProvider getGasType() {
     public void appendHoverText(@Nonnull ItemStack stack, @Nullable Level world, @Nonnull List<Component> tooltip, @Nonnull TooltipFlag flag) {
         super.appendHoverText(stack, world, tooltip, flag);
         tooltip.add(MekanismLang.MODE.translateColored(EnumColor.GRAY, getMode(stack).getTextComponent()));
+        if (flag.isAdvanced() && Mekanism.hooks.CuriosLoaded && !(stack.is(MekanismTags.Items.CURIOS_BODY) || stack.is(MekanismTags.Items.CURIOS_BACK))) {
+            tooltip.add(MekanismLang.CURIO_NOT_USABLE.translateColored(EnumColor.DARK_GRAY));
+        }
```
pupnewfster, line 88: Having this feels super weird especially based on the wording of the lang entry. I think adding this tooltip should just be removed as if it can go in the curio slots, then curios will add which slots it can go in. And then we also don't have to define these two tags (as currently you are defining them as non-optional even if they are semi optional just due to current data validation.

src/main/java/mekanism/common/item/gear/ItemJetpack.java
```diff
@@ -89,12 +99,12 @@ public void setMode(ItemStack stack, JetpackMode mode) {
...
         if (slotType == getSlot()) {
-            ItemJetpack jetpack = (ItemJetpack) stack.getItem();
+            final ItemJetpack jetpack = (ItemJetpack) stack.getItem();
             list.add(MekanismLang.JETPACK_MODE.translateColored(EnumColor.DARK_GRAY, jetpack.getMode(stack)));
             GasStack stored = GasStack.EMPTY;
-            Optional<IGasHandler> capability = stack.getCapability(Capabilities.GAS_HANDLER_CAPABILITY).resolve();
+            final Optional<IGasHandler> capability = stack.getCapability(Capabilities.GAS_HANDLER_CAPABILITY).resolve();
             if (capability.isPresent()) {
-                IGasHandler gasHandlerItem = capability.get();
+                final IGasHandler gasHandlerItem = capability.get();
```
pupnewfster, line 107: No need to change these lines to have final given you aren't touching them otherwise.

src/main/java/mekanism/common/item/gear/ItemJetpack.java
```diff
@@ -103,6 +113,11 @@ public void addHUDStrings(List<Component> list, Player player, ItemStack stack,
         }
     }
 
+    @Override
+    public void addCurioHUDStrings(List<Component> list, Player player, ItemStack stack) {
+        addHUDStrings(list, player, stack, getSlot());
+    }
```
pupnewfster, line 119: I feel like this impl should be removed from here and instead moved to `IItemHUDProvider` and have the default be:
```java
if (this instanceof ArmorItem armor) {
    addHUDStrings(list, player, stack, armor.getSlot());
}
```

src/main/java/mekanism/client/render/MekanismCurioRenderer.java
```diff
@@ -0,0 +1,24 @@
...
+import net.minecraft.client.renderer.texture.OverlayTexture;
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.item.ItemStack;
+import top.theillusivec4.curios.api.SlotContext;
+import top.theillusivec4.curios.api.client.ICurioRenderer;
+
+public record MekanismCurioRenderer(ICustomArmor model) implements ICurioRenderer {
+
+    @Override
+    public <T extends LivingEntity, M extends EntityModel<T>> void render(ItemStack stack, SlotContext slotContext, PoseStack matrixStack, RenderLayerParent<T, M> renderLayerParent, MultiBufferSource renderTypeBuffer, int light, float limbSwing, float limbSwingAmount, float partialTicks, float ageInTicks, float netHeadYaw, float headPitch) {
```
pupnewfster, line 18: Line break before `RenderLayerParent<....`

src/main/java/mekanism/client/render/MekanismCurioRenderer.java
```diff
@@ -0,0 +1,24 @@
...
+        final M model = renderLayerParent.getModel();
+        if (model instanceof HumanoidModel<? extends LivingEntity> humanoidModel) {
```
pupnewfster, line 20: No need for a local variable here other than humanoidModel this can just be:
```java
if (renderLayerParent.getModel() instanceof HumanoidModel<?> humanoid) {
    model.render(humanoid, ....);
}
```

src/main/java/mekanism/common/integration/curios/CuriosIntegration.java
```diff
@@ -0,0 +1,51 @@
...
+public class CuriosIntegration {
+
+    public static void sendIMC() {
+        InterModComms.sendTo(MekanismHooks.CURIOS_MODID, "register_type", () -> SlotTypePreset.BODY.getMessageBuilder().build());
+    }
+
+    public static void addListeners(final IEventBus bus) {
+        bus.addListener((FMLClientSetupEvent event) -> {
+            CuriosRendererRegistry.register(MekanismItems.JETPACK.asItem(), () -> new MekanismCurioRenderer(JetpackArmor.JETPACK));
+            CuriosRendererRegistry.register(MekanismItems.ARMORED_JETPACK.asItem(), () -> new MekanismCurioRenderer(JetpackArmor.ARMORED_JETPACK));
```
pupnewfster, line 32: I think you should add a helper method that basically is:
```java
private static void registerRenderers(IItemProvider... providers) {
    for (IItemProvider provider : providers) {
        if (provider.asItem() instanceof ArmorItem armor && RenderProperties.get(armor) instanceof ISpecialGear gear) {
            CuriosRendererRegistry.register(armor, () -> new MekanismCurioRenderer(gear.getGearModel(armor.getSlot())));
        } else {
            Mekanism.logger.warn("Attempted to register curios renderer for a non-special gear item.");
        }
    }
}
```

and then you could just make the `addListener` method basically be:
```java
bus.addListener((FMLClientSetupEvent event) -> registerRenderers(MekanismItems.JETPACK, MekanismItems.ARMORED_JETPACK));
```

src/main/java/mekanism/client/render/hud/MekaSuitHUD.java
```diff
@@ -31,16 +34,22 @@ public void render(ForgeIngameGui gui, PoseStack poseStack, float partialTicks,
...
-                    hudProvider.addHUDStrings(list, minecraft.player, stack, slotType);
-                    int size = list.size();
-                    if (size > 0) {
-                        renderStrings.put(slotType, list);
-                        count += size;
+                    count += makeComponent(list -> hudProvider.addHUDStrings(list, minecraft.player, stack, slotType), renderStrings);
+                }
+            }
+            if (Mekanism.hooks.CuriosLoaded) {
+                final var invOptional = CuriosIntegration.getCuriosInventory(minecraft.player);
```
pupnewfster, line 45: Switch this to not use var.

src/main/java/mekanism/client/render/hud/MekaSuitHUD.java
```diff
@@ -31,16 +34,22 @@ public void render(ForgeIngameGui gui, PoseStack poseStack, float partialTicks,
...
+                if (invOptional.isPresent()) { // Can't use ifPresent due to count not being final
+                    final IItemHandler inv = invOptional.get();
+                    for (int i = 0; i < inv.getSlots(); i++) {
```
pupnewfster, line 48:
```java
for (int i = 0, slots = inv.getSlots(); i < slots; i++) {
```

pupnewfster (changes requested):

src/main/java/mekanism/common/item/gear/ItemJetpack.java
Matyrobbrt: I can keep the tooltip without needing to add those 2 tags as I discovered there's a method which gets the curio tags that an item has. But I can just remove the tooltip if you want.

pupnewfster: I feel like it should just be removed given the wording of it is a bit odd and also while I don't have a major issue with having optional support for it if people want it I feel like having that message is more of a nag telling people they should heavily consider adding it to a curio slot tag.

build.gradle
```diff
@@ -146,9 +146,7 @@ static void extendConfigurations(Configuration base, Configuration... configurat
 
 ext {
     versionProperties = ["version"           : mod_version, "mc_version": minecraft_version_range, "forge_version": forge_version_range,
-                         "loader_version"    : loader_version_range, "crafttweaker_version": crafttweaker_version_range, "jei_version": jei_version_range,
-                         "wildfire_gender_version": wildfire_gender_version_range,
-                         "curios_version": curios_version_range]
+                         "loader_version"    : loader_version_range, "crafttweaker_version": crafttweaker_version_range, "jei_version": jei_version_range, "wildfire_gender_version": wildfire_gender_version_range, "curios_version": curios_version_range]
```
pupnewfster, line 149: I meant the wildfire_gender_version and the curios_version should be on the same line. That way it still fits on the screen just fine.

src/main/java/mekanism/client/render/MekanismCurioRenderer.java
```diff
@@ -0,0 +1,24 @@
...
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.item.ItemStack;
+import top.theillusivec4.curios.api.SlotContext;
+import top.theillusivec4.curios.api.client.ICurioRenderer;
+
+public record MekanismCurioRenderer(ICustomArmor model) implements ICurioRenderer {
+
+    @Override
+    public <T extends LivingEntity, M extends EntityModel<T>> void render(ItemStack stack, SlotContext slotContext, PoseStack matrixStack,
+                  RenderLayerParent<T, M> renderLayerParent, MultiBufferSource renderTypeBuffer, int light, float limbSwing, float limbSwingAmount, float partialTicks, float ageInTicks, float netHeadYaw, float headPitch) {
```
pupnewfster, line 19: I didn't realize quite how long this line is but probably worth adding another line break before the `float partialTicks`

src/main/java/mekanism/common/item/gear/ItemJetpack.java
Matyrobbrt: My though process was that no-one would've known that it works in a Curio slot. Anyways, I'll remove it

src/main/java/mekanism/common/tags/MekanismTags.java
```diff
@@ -10,6 +10,7 @@
 import mekanism.api.chemical.infuse.InfuseType;
 import mekanism.api.chemical.slurry.Slurry;
 import mekanism.common.Mekanism;
+import mekanism.common.integration.MekanismHooks;
```
pupnewfster, line 13: No need to add this import.

src/main/java/mekanism/common/item/interfaces/IItemHUDProvider.java
```diff
@@ -4,9 +4,16 @@
...
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ArmorItem;
 import net.minecraft.world.item.ItemStack;
 
 public interface IItemHUDProvider {
 
     void addHUDStrings(List<Component> list, Player player, ItemStack stack, EquipmentSlot slotType);
+
+    default void addCurioHUDStrings(List<Component> list, Player player, ItemStack stack) {
+        if (this instanceof ArmorItem armour) {
```
pupnewfster, line 15: `armour` -> `armor`

src/main/java/mekanism/client/key/MekanismKeyHandler.java
```diff
@@ -63,9 +66,27 @@ public static void registerKeybindings() {
...
+            } else if (Mekanism.hooks.CuriosLoaded) {
+                CuriosIntegration.findCurio(player, s -> s.canEquip(slot, player))
+                        .stream()
+                        .filter(r -> {
+                            final boolean isMode = IModeItem.isModeItem(r.stack(), slot);
+                            if (r.stack().getItem() instanceof IGasItem gas) {
+                                return isMode && !gas.getGas(r.stack()).isEmpty();
+                            }
+                            return isMode;
+                        })
```
pupnewfster, line 82: I feel like this may be more readable as two filter calls?
```java
.filter(r -> IModeItem.isModeItem(r.stack(), slot))
.filter(r -> !(r.stack().getItem() instanceof IGasItem item) || !item.getGas(r.stack()).isEmpty())
```

Matyrobbrt: That would however create 2 streams, so it'd be less efficient

pupnewfster: True though it also isn't the most performance critical spot as it is only ran on keybinding press, but even so you could rewrite it to be:
```java
.filter(r -> {
    ItemStack stack = r.stack();
    if (IModeItem.isModeItem(stack, slot)) {
        return !(stack.getItem() instanceof IGasItem item) || !item.getGas(stack).isEmpty();
    }
    return false;
})
```
Which also would have the benefit of removing the performance impact of the instanceof call if it isn't a mode item.

build.gradle
```diff
@@ -146,8 +146,7 @@ static void extendConfigurations(Configuration base, Configuration... configurat
 
 ext {
     versionProperties = ["version"           : mod_version, "mc_version": minecraft_version_range, "forge_version": forge_version_range,
-                         "loader_version"    : loader_version_range, "crafttweaker_version": crafttweaker_version_range, "jei_version": jei_version_range,
-                         "wildfire_gender_version": wildfire_gender_version_range]
+                         "loader_version"    : loader_version_range, "crafttweaker_version": crafttweaker_version_range, "jei_version": jei_version_range, "wildfire_gender_version": wildfire_gender_version_range, "curios_version": curios_version_range]
```
pupnewfster, line 149: ??? Why do you keep marking this as resolved, after `jei_version_range,` leave the new line

src/main/java/mekanism/client/key/MekanismKeyHandler.java
```diff
@@ -63,9 +67,33 @@ public static void registerKeybindings() {
...
+                                return isMode && !gas.getGas(r.stack()).isEmpty();
+                            }
+                            return isMode;
+                        }).filter(r -> {
+							ItemStack stack = r.stack();
+							if (IModeItem.isModeItem(stack, slot)) {
+								return !(stack.getItem() instanceof IGasItem item) || !item.getGas(stack).isEmpty();
+							}
+							return false;
+						})
```
pupnewfster, line 89: Spacing is super weird here (due to it using tabs rather than spaces). Also like you have the filter taking place twice instead of replacing the existing one you had?

src/main/java/mekanism/client/key/MekanismKeyHandler.java
```diff
@@ -63,9 +67,33 @@ public static void registerKeybindings() {
...
+                        .findFirst()
+                        .ifPresent(result -> {
+                            Mekanism.packetHandler().sendToServer(new PacketModeChangeCurios(result.slotContext().index(), player.isShiftKeyDown()));
+                            SoundHandler.playSound(MekanismSounds.HYDRAULIC);
+                        });
+            }
         }
     }
+
```
pupnewfster, line 98: No need to add this blank line

src/main/java/mekanism/common/item/interfaces/IItemHUDProvider.java
```diff
@@ -4,9 +4,16 @@
...
+import net.minecraft.world.item.ArmorItem;
 import net.minecraft.world.item.ItemStack;
 
 public interface IItemHUDProvider {
 
     void addHUDStrings(List<Component> list, Player player, ItemStack stack, EquipmentSlot slotType);
+
+    default void addCurioHUDStrings(List<Component> list, Player player, ItemStack stack) {
+        if (this instanceof ArmorItem armor) {
+            addHUDStrings(list, player, stack, armour.getSlot());
```
pupnewfster, line 16: Uhhh I feel like you didn't test if this compiles after changing the variable name.

pupnewfster (changes requested):

src/main/java/mekanism/common/item/interfaces/IItemHUDProvider.java
Matyrobbrt: Oh my bad

build.gradle
Matyrobbrt: Should be fixed now

build.gradle
```diff
@@ -146,8 +146,8 @@ static void extendConfigurations(Configuration base, Configuration... configurat
 
 ext {
     versionProperties = ["version"           : mod_version, "mc_version": minecraft_version_range, "forge_version": forge_version_range,
-                         "loader_version"    : loader_version_range, "crafttweaker_version": crafttweaker_version_range, "jei_version": jei_version_range,
-                         "wildfire_gender_version": wildfire_gender_version_range]
+                         "loader_version"    : loader_version_range, "crafttweaker_version": crafttweaker_version_range, "jei_version": jei_version_range,
```
pupnewfster, line 149: Remove added space at end of line.

src/main/java/mekanism/client/key/MekanismKeyHandler.java
```diff
@@ -63,9 +67,25 @@ public static void registerKeybindings() {
...
+                Mekanism.packetHandler().sendToServer(new PacketModeChange(slot, player.isShiftKeyDown()));
+                SoundHandler.playSound(MekanismSounds.HYDRAULIC);
+            } else if (Mekanism.hooks.CuriosLoaded) {
+                CuriosIntegration.findCurio(player, s -> s.canEquip(slot, player))
+                        .stream()
+                        .filter(r -> {
+                            if (IModeItem.isModeItem(stack, slot)) {
+                                return !(stack.getItem() instanceof IGasItem item) || !item.getGas(stack).isEmpty();
+                            }
+                            return false;
```
pupnewfster, line 81: Does not compile.

src/main/java/mekanism/client/render/MekanismCurioRenderer.java
```diff
@@ -0,0 +1,25 @@
...
+import top.theillusivec4.curios.api.SlotContext;
+import top.theillusivec4.curios.api.client.ICurioRenderer;
+
+public record MekanismCurioRenderer(ICustomArmor model) implements ICurioRenderer {
+
+    @Override
+    public <T extends LivingEntity, M extends EntityModel<T>> void render(ItemStack stack, SlotContext slotContext, PoseStack matrixStack,
+                  RenderLayerParent<T, M> renderLayerParent, MultiBufferSource renderTypeBuffer, int light, float limbSwing, float limbSwingAmount, 
+				  float partialTicks, float ageInTicks, float netHeadYaw, float headPitch) {
+        if (renderLayerParent.getModel() instanceof HumanoidModel<? extends LivingEntity> humanoidModel) {
```
pupnewfster, line 21: This can just be `instanceof HumanoidModel<?>` as all `HumanoidModel`s must have the generic extend `LivingEntity`

src/main/java/mekanism/common/integration/curios/CuriosIntegration.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    public static Optional<? extends IItemHandler> getCuriosInventory(LivingEntity living) {
+        return CuriosApi.getCuriosHelper().getEquippedCurios(living).resolve();
+    }
+
+    public static Optional<ItemStack> findFirstCurio(@Nonnull LivingEntity livingEntity, Predicate<ItemStack> filter) {
+        return findFirstCurioAsResult(livingEntity, filter).map(SlotResult::stack);
+    }
```
pupnewfster, line 54: This method seems to be unused? If so might as well just remove it as if we ever need a method like that we can add it at that time.

src/main/java/mekanism/common/item/gear/ItemJetpack.java
```diff
@@ -128,6 +130,30 @@ public int getDefaultTooltipHideFlags(@Nonnull ItemStack stack) {
...
+    @Nonnull
+    public static ItemStack getJetpack(LivingEntity entity) {
+        return getJetpack(entity, entity.getItemBySlot(EquipmentSlot.CHEST));
+    }
+
+    @Nonnull
+    public static ItemStack getJetpack(LivingEntity entity, ItemStack chest) {
+        if (chest.getItem() instanceof ItemJetpack) {
+            return chest;
+        } else if (Mekanism.hooks.CuriosLoaded) {
```
mewacaser, line 151: We're missing a case here for the mekasuit. I was messing around with something with the jetpack which I was thinking about making a pr for. But I'm not actually sure it's something which will get approved. Do you want me to put that in with it, and potentially refactoring the jetpack logic to an interface?

pupnewfster: Having an interface might be useful but the big thing about jetpack vs mekasuit (which is why the places that call this also special case handle checking the mekasuit (which the mekasuit jetpack doesn't work in the curio slots by design) is that the mekasuit is conditional on whether the module is even enabled.

mewacaser: yeah, I only noticed this as it's not working currently, adding an interface is turning into a small scale refactor so it might just be faster to hotfix this and rebase my changes on that.

PR #7455, opened 2022-04-20
Add deepslate certus ore->dust recipe

JustinHuPrime:
Add a deepslate certus ore to certus quartz dust recipe similar to the existing regular certus ore recipe


pupnewfster:
Does AE2 not tag both their ores in a single tag? And if not I feel like instead of a second recipe just using an ingredient that matches multiple items.

JustinHuPrime:
There's either the option of using the `forge:ores/certus_quartz` tag, or just listing the two ore items. Since the certus crystal->dust  recipe just lists the items despite there being the `forge:gems/certus_quartz` tag, I just listed the two ore items.

pupnewfster:
I think it may make more sense to use the tag and change the crystal one to using tags as well assuming AE2 adds both tags. The reason it didn't use tags was that when I was porting the recipes to 1.18 and looked at AE2's repo they didn't have any tags on it. Looking again I am guessing that is because they don't include their generated resources on the repo and only include them for builds(?).

PR #7468, opened 2022-05-06
Added raw block recipes and organized existing recipes

mewacaser:
With raw ore needing 3 ore for its processing recipe I thought it would useful to have a bulk recipe for the raw ore block that does 3 times as much but only had a single item input. This would also be nice if you didn't want to use a sorter to input into your processing line.

I experimented with it only requiring twice as much secondary input as a bulk efficiency thing, but I think it would probably be better to switch to 3 as the speed/power efficiency is probably enough of an advantage already.

I also organized the other processing recipes so I could make sure I wasn't missing any. I hope that's ok, if not I'm fine with putting them back where they were.


pupnewfster:
I will try and look into this a bit closer sometime before 10.2.1 is released but one thing that stands out from a very brief glance is the use of tabs in recipe provider file. Our standard formatting for Mekanism uses four spaces rather than a tab.

mewacaser:
Sorry, fixed that, had show whitespace off. Amended so tabs won't be in the repo at all.

PR #7469, opened 2022-05-06
Added gyroscopic stabilization unit

mewacaser:
Added a module which gets rid of the mining speed decrease from being under water or in the air.

With the jetpack you are in the air a lot more and it's nice to be able to mine full speed when hovering.

One potential issue is that I don't know how it would react to other mods attempts to get rid of the in the air slowdown. It might additionally multiply if they have enchants like "Air Affinity" or "Stable Footing" since those go on slots other than the chest. But I feel like that's pretty minor as the mekatool can almost instamine anything anyway, and it's unlikely they'd not be using the mekasuit helmet if they're using the chestplate.


Seniorendi:
I like that

src/main/java/mekanism/common/registries/MekanismModules.java
```diff
@@ -89,6 +90,8 @@ private MekanismModules() {
                 .rendersHUD().exclusive());
     public static final ModuleRegistryObject<?> ELYTRA_UNIT = MODULES.registerMarker("elytra_unit", () -> MekanismItems.MODULE_ELYTRA.asItem(),
           builder -> builder.rarity(Rarity.EPIC));
+    public static final ModuleRegistryObject<ModuleGyroscopicStabilizationUnit> GYROSCOPIC_STABILIZATION_UNIT = MODULES.register("gyroscopic_stabilization_unit",
```
pupnewfster, line 93: Use `registerMarker` instead of using `register` and creating an empty dummy class implementation.

src/main/java/mekanism/common/CommonPlayerTickHandler.java
```diff
@@ -382,4 +386,21 @@ private FallEnergyInfo getFallAbsorptionEnergyInfo(LivingEntity base) {
 
     private record FallEnergyInfo(@Nullable IEnergyContainer container, FloatSupplier damageRatio, FloatingLongSupplier energyCost) {
     }
+
+    @SubscribeEvent
+    public void getBreakSpeed(BreakSpeed event) {
+        Player player = event.getPlayer();
+        IModule<ModuleGyroscopicStabilizationUnit> module = MekanismAPI.getModuleHelper().load(player.getItemBySlot(EquipmentSlot.CHEST), MekanismModules.GYROSCOPIC_STABILIZATION_UNIT);
+        if (module != null && module.isEnabled()) {
```
pupnewfster, line 394:
```java
ItemStack chest = player.getItemBySlot(EquipmentSlot.CHEST);
if (!chest.isEmpty() && MekanismAPI.getModuleHelper().isEnabled(chest, MekanismModules.GYROSCOPIC_STABILIZATION_UNIT)) {
```

src/main/java/mekanism/common/CommonPlayerTickHandler.java
```diff
@@ -382,4 +386,21 @@ private FallEnergyInfo getFallAbsorptionEnergyInfo(LivingEntity base) {
...
+            float speed = event.getOriginalSpeed();
```
pupnewfster, line 395: Do we maybe want to compare original and new speed to estimate if another mod already is doing this? (Such as if new is 5x original just don't do anything?)

Also I feel like if our calculated speed is less than the new speed we may not want to be updating the break speed. Though maybe we should just be operating on the new speed in the first place? That way we don't cause some other thing from mods to get overridden because they happened before us.

pupnewfster (commented):
Did a very quick pass looking at this and it seems like a nice idea. Part of me is leaning towards the module being for the leggings rather than the chestplate though as we already have so many chestplate modules and it sort of makes sense that the legs would have more to do with stabilization than the chest. Any thoughts?

mewacaser:
I'm fine with switching it to legs. I was thinking if it was real it would be at the center of mass, but also head feels normal because of aqua affinity, and boots wouldn't be that weird for something that's making you more stable so I wasn't sure which slot to go with. But I like adding more for the legs too. Speaking of which I was thinking of a module to increase water movement speed that would also make sense on the legs, or maybe just prevent flowing water from slowing you down as it feels really lame to go from full speed to a crawl. Wasn't sure where to go with that or if it should just be Depth Strider module, but I don't want to make the first module that stacks to 3 either...

src/main/java/mekanism/common/registries/MekanismModules.java
mewacaser: I even looked for a way to do that, apparently not hard enough though.

pupnewfster:
That sounds like a decent idea for a module, and honestly having a module that only stacks to three isn't something that we are opposed to if you end up going the route of using depth strider more directly.

src/main/java/mekanism/common/CommonPlayerTickHandler.java
mewacaser: I don't really like the idea of comparing new and original to figure out if they've already done it as they might get called after us, but if they also do that it might work. I could see just using the new speed, it might result in really fast break speeds but only when you're in the air... and most likely with a tool which can have 128 efficiency anyway, because of the edge case nature I think I'm honestly ok with whatever.

Seniorendi:
gasm

PR #7470, opened 2022-05-06
Advancements for mekanism

Tyrannicodin:
Preliminary creation of advancements for Mekanism let me know if you think anything should be changed, updated or added.
### Multiblock and gas production 
Currently multiblocks can't be detected as Minecraft has no way of creating custom advancement triggers, and gases can only be detected in the inventory, and as most gases are just put through pressurised tubes and other machines, so most likely wont be taken in gas containers.


pupnewfster:
If these were actually generated (which would be great) then it looks like the providers and the cache file differences were missed when committing. If they weren't then the things should probably not be in the datagen folder or the next time `runData` is ran they will all be deleted. Though I probably can help with moving advancements over to datagen after I get a chance to review things.

Tyrannicodin:
To be honest, I didn't really know where to put them and that seemed to worked as it was with all the other data, if you let me know where they should really be I'll move them there.

Tyrannicodin:
Figured out how to use data gen looking by looking at [this mod](https://github.com/Andromander/Gaia-Dimension/blob/1.18.x/src/main/java/androsa/gaiadimension/data/GaiaAdvancements.java), will try to implement soon.

src/datagen/generated/mekanism/assets/mekanism/lang/en_us.json
```diff
@@ -1,4 +1,64 @@
...
+  "advancements.mekanism.robit.robit.title":"Welcome to the world",
+  "advancements.mekanism.robit.robit.description":"Hello friend",
+  "advancements.mekanism.teleports.portableteleporter.title":"Portable portals",
+  "advancements.mekanism.teleports.portableteleporter.description":"Instant travel from anywhere",
+  "advancements.mekanism.teleports.quantumentangloporter.title":"Quantum entanglement",
+  "advancements.mekanism.teleports.quantumentangloporter.description":"Instant resource transport",
+  "advancements.mekanism.teleports.teleportcore.title":"Instant teleportation",
+  "advancements.mekanism.teleports.teleportcore.description":"Numerous applications!",
+  "advancements.mekanism.teleports.teleporter.title":"Portals:online",
+  "advancements.mekanism.teleports.teleporter.description":"Instant travel",
```
pupnewfster, line 61: Where are these being added from? They should be somewhere in the [lang provider](https://github.com/mekanism/Mekanism/blob/1.18.x/src/datagen/main/java/mekanism/client/lang/MekanismLangProvider.java)

pupnewfster (commented):
I think you forgot to rerun datagen after moving the rest of the advancements over (or you forgot to commit the added files).

src/datagen/generated/mekanism/assets/mekanism/lang/en_us.json
Tyrannicodin: I haven't added them yet, just trying to get nbt to be working.

pupnewfster: Ah I didn't read the sub message in your commit. And in regards to creating a stack with (if I understand properly) maxed upgrades the way I would do it is probably via a helper method like this:
```java
private ItemStack getMaxedGear(ItemRegistryObject<? extends IModuleContainerItem> item) {
    ItemStack stack = item.getItemStack();
    if (stack.getItem() instanceof IModuleContainerItem container) {
        for (ModuleData<?> module : MekanismAPI.getModuleHelper().getSupported(stack)) {
            container.addModule(stack, module);
            ModuleHelper.INSTANCE.load(stack, module).setInstalledCount(module.getMaxStackSize());
        }
    }
    return stack;
}
```

And then I think you can pass that stack as the display stack to the display method instead of just passing the item.

Tyrannicodin: This should work, thanks

Tyrannicodin: hmm, .getSupported(stack) return an empty array

pupnewfster:
Pushed a couple changes to your branch to clean things up slightly and also get the NBT working for the supported modules. I probably will end up at some point trying to just make a custom advancement trigger so that we can just flag it as requiring all modules and then it calculates it at runtime based on all the available modules (in case mods add any).

Tyrannicodin:
This should all be fixed and is now generated data.

pupnewfster:
Pushed a bit of cleanup, and also in theory got it so that the check for a full canteen properly obeys the configured max storage based on the config and the fully upgraded mekasuit properly checks only installed types and should support custom modules.

A couple things I noticed were:
1. Why were some descriptions obfuscated? I feel like that most likely doesn't fit (so I removed them for now but with TODOs to add back if there is a reasoning)
2. It seems there was some copy pasting errors so there is no title or description set for the digital miner advancement (see MekanismLangProvider and the todo I left in the addAdvancements section)

Beyond that I haven't looked in too great detail yet at the actual advancements yet, but figured I would give an update on the state of this.

pupnewfster:
Also I am thinking of targeting this PR for 1.19 and potentially adding some misc advancements + making the other modules like generators have some basic advancements that get added into the Mekanism tree. So odds are I will merge this into the 1.19 branch before actually starting the port and then I might ask you for suggestions on discord for some of the other modules.

Tyrannicodin:
> Why were some descriptions obfuscated?

I wasn't really sure what to put, and it was supposed to be like the antimatter was glitching everything after it was created.
> Copy-pasting errors

I think I know where I went wrong with there, as there was a double robit which I removed, that was probably an error in my code that replaced the digital miner with it, I'll take a look

Other than that, all sounds good lmk if you want me to do anything else

Tyrannicodin:
Could create a discord channel for advancement suggestions as I wasn't sure what else to add, and I'm sure people in the discord have loads of ideas. They could also probably improve the actual names and descriptions for the titles if I make a tree diagram with that sort of stuff, depends on how long the release will be I guess.

pupnewfster:
On second thoughts odds are this will be a couple minor versions into 1.19, but I will rebase and port this PR to 1.19 as needed when the time comes.

pupnewfster:
I went ahead and rebased this onto the 1.19 branch, and will likely start looking at it in more depth later today as well as figuring out which other things we may want advancements for, both for base mek and the various modules.

PR #7471, opened 2022-05-06
MekaTool radial profiles kinda working, more GUI work needed

mewacaser:
So I got this idea from Mekanism-Feature-Requests#282

It is mostly working and it is awesome to use.
Currently limited to 6 profiles and while it has to be hardcoded to be in the enum there is no reason there couldn't be more, and potentially even disable them based on a numeric property

That being said there are issues with how I did it that probably need to be fixed.
The server sync kinda just piggybacks off of the module.save currently which works, but isn't ideal.
It's currently dealing with the NBT directly instead of using the config.set methods so that the module.save doesn't overwrite it, I have some potential solutions in mind, but I'm not sure which way would be best.
It only saves an option which you have changed while on that profile. Which if there was a way to clear a profile could actually be a feature.
The save button in the GUI doesn't do anything right now.
The profile names can be changed with the GUI, but it's static so it's the same for every MekaTool.

This also kinda makes the handle mode key option on the tool modules obsolete. So we might want to remove that on them.


Tyrannicodin:
6 seems a good number to me, and we could maybe use the handle mode key to open the menu instead of switching modes if it'll just be obsolete

mewacaser:
To clarify the handle mode key option previously decided if trying to change the mode would change the mode on that module, the key itself is still being used for the radial menu, but it can change the mode for any config. What I was suggesting possibly removing was just the boolean on the module to say whether it handled the key.

The NBT profiles are currently on the itemstacks, which is also where the module data is, if you switched to a different mekatool it could have different modules/configs/profiles. I don't think we'd want that to be global, even though the profile names are currently.

Tyrannicodin:
I see what you mean now, and I definitely agree. You're probably right about radial wheels being NBT not global as well

Seniorendi:
> It is mostly working and it is awesome to use.

Show me <:

mewacaser:
https://user-images.githubusercontent.com/105073512/167268429-7d912db0-e122-43bc-a1ce-e8f89f45221e.mp4

src/main/java/mekanism/common/item/gear/ItemMekaTool.java
```diff
@@ -342,18 +360,50 @@ public boolean canApplyAtEnchantingTable(ItemStack stack, Enchantment enchantmen
...
-    public boolean supportsSlotType(ItemStack stack, @Nonnull EquipmentSlot slotType) {
-        return IModeItem.super.supportsSlotType(stack, slotType) && getModules(stack).stream().anyMatch(Module::handlesModeChange);
-    }
+    public enum MekaToolMode implements IRadialSelectorEnum<MekaToolMode>, IHasTextComponent {
+        ONE("mode_1"),
+        TWO("mode_2"),
+        THREE("mode_3"),
+        FOUR("mode_4"),
+        FIVE("mode_5"),
+        SIX("mode_6");
```
Matyrobbrt, line 369: In the state that this enum currently is, I feel like the `key` parameter shouldn't exist, and `getKey` should just return `"mode_" + (ordinal() + 1)`.

src/main/java/mekanism/common/item/gear/ItemMekaTool.java
```diff
@@ -342,18 +360,50 @@ public boolean canApplyAtEnchantingTable(ItemStack stack, Enchantment enchantmen
...
+
+        public String getKey() {
+            return key;
+        }
+
+        public boolean isEnabled() {
+            return true;
+        }
+
+        public static MekaToolMode byIndexStatic(int index) {
```
Matyrobbrt, line 385: What is the point of having a static and non-static method for doing the same thing?

wacaser2: So I'm not sure since I'm not at my pc right now and I copied some of this from the atomic disassembler implementation. But I think that the instance method is an override for the IRadialModeEnum, but the static one is the only one used I think.

mewacaser: So I'm not sure since I'm not at my pc right now and I copied some of this from the atomic disassembler implementation. But I think that the instance method is an override for the IRadialModeEnum, but the static one is the only one used I think.

src/main/java/mekanism/common/item/gear/ItemMekaTool.java
mewacaser: Valid point, I added it because I was constructing it in multiple places but it doesn't have to be defined I guess.

src/main/java/mekanism/common/item/gear/ItemMekaTool.java
mewacaser: Double checked and pretty much every mode object has this exact implementation as well, including the valve blocks. I can't figure out how the instance one could be used where the static wouldn't. Maybe if you wanted the next mode but didn't know which enum you had?

mewacaser:
I think everything is resolved on this except for the static mode names. I don't see a way to fix that so I think it's done unless someone else can think of something.

Peekofwar:
> 6 seems a good number to me, and we could maybe use the handle mode key to open the menu instead of switching modes if it'll just be obsolete. As for nbt storage, we might want to store it globally for all meka-tools, because if you're using on meka-tool and switch to another (no clue why you'd do that), you would still want to be able to use your configs.

What if you wanted to have specialized Meka Tools with their own configs? Or maybe you could use the data configuration card to transfer configs. Also, how would this effect using another player's meka tool?

PR #7475, opened 2022-05-11
Jetpack Thrust Orientation

mewacaser:
Changed the Regular jetpack mode to provide thrust towards the view's up vector instead of the world's.
Just a fun change really, I kept the speed to the same cap (actually lower as the total motion delta is clamped now and not just the vertical) so it's not giving super speed but you can maneuver by where you're facing now which is cool.

Also extracted the jetpack logic to an interface, still some more optimization that can be done though I think. For example ClientTickHandler is calling the getJetpack method (now getActiveJetpack) 3 times in its tick method. But it is working with the interface and it should allow jetpacks which don't use gas so a jetpack could be added which used the existing functionality and used a liquid, energy, or maybe even a cooldown.


pupnewfster (commented):
Overall definitely think changing to an interface makes things cleaner though I left a few comments. Unsure about the change to how the motion is handled though as I haven't gone ahead and tested it yet to see how it feels so after I do there is a *chance* I may end up having you revert it.

src/main/java/mekanism/client/ClientTickHandler.java
```diff
@@ -86,15 +87,13 @@ public static boolean isJetpackActive(Player player) {
...
+            ItemStack activeJetpack = IJetpackItem.getActiveJetpack(player);
+            if (!activeJetpack.isEmpty()) {
+                JetpackMode mode = ((IJetpackItem) activeJetpack.getItem()).getJetpackMode(activeJetpack);
                 if (mode == JetpackMode.NORMAL) {
                     return minecraft.screen == null && minecraft.player.input.jumping;
                 } else if (mode == JetpackMode.HOVER) {
-                    boolean ascending = minecraft.player.input.jumping;
-                    boolean descending = minecraft.player.input.shiftKeyDown;
-                    if (!ascending || descending || minecraft.screen != null) {
+                    if (!minecraft.player.input.jumping || minecraft.player.input.shiftKeyDown || minecraft.screen != null) {
```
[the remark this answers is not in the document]
mewacaser: I changed it as it wasn't doing that for the block below, but I'll revert and maybe add it to that one

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -0,0 +1,130 @@
...
+     * If Curios is loaded, the curio slots will be checked as well.
+     *
+     * @param entity the entity on which to look for the jetpack
+     *
+     * @return the jetpack stack if present, otherwise an empty stack
+     */
+    @Nonnull
+    static ItemStack getActiveJetpack(LivingEntity entity) {
+        ItemStack chest = entity.getItemBySlot(EquipmentSlot.CHEST);
+        if (chest.getItem() instanceof IJetpackItem jetpackItem && jetpackItem.isActive(chest)) {
```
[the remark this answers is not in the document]
mewacaser: So I know it is different than it was before. My thought was that if you ran out of fuel on the current one it should switch to a backup in curios. But you're right that this would be unintuitive with disabling the worn one. Maybe we switch isActive to not check the mode?

src/main/java/mekanism/client/ClientTickHandler.java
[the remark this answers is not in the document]
pupnewfster: Main reason I didn't bother checking it presuming you mean for normal mode that is mainly because it shows what it corresponds to two lines below (where that variable is declared).

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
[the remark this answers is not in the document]
pupnewfster: Personally I think it shouldn't use curios at all if the mekasuit is being worn (nor should the mekasuit work in curios), making the normal jetpack fallback to curios if the one in the chestplate is empty seems like it may be reasonable though.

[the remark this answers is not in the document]
mewacaser: It's less likely for the mekasuit to run out of fuel, but I don't know why we'd prevent it from defaulting to a curio jetpack if it did. 
I feel like we shouldn't be doing anything different for IJetpackItems though, so we'd have to put the check for whether the mekasuit is in curios in the mekasuit implementation I think, and set isActive to always true so that it won't go look at curios if that's what you decide.

src/main/java/mekanism/client/ClientTickHandler.java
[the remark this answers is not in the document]
mewacaser: You're right, for some reason I thought the normal block was below, not above.

mewacaser:
I was also wondering if IJetpackItem (and/or IBlastingItem) should be put in the api in case someone wanted to implement it?

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -95,7 +96,9 @@ static ItemStack getActiveJetpack(LivingEntity entity) {
      */
     static boolean handleJetpackMotion(Player player, JetpackMode mode, BooleanSupplier ascendingSupplier) {
         Vec3 motion = player.getDeltaMovement();
-        if (mode == JetpackMode.NORMAL) {
+        if (mode == JetpackMode.ASCEND) {
+            player.setDeltaMovement(motion.x(), Math.min(motion.y() + 0.15D, 0.5D), motion.z());
```
pupnewfster, line 100: I think this is kind of pointless and shouldn't be added tbh.

omeranha:
hi @mewacaser I don't think this is the appropriate location, but, can I talk with you in private?

mewacaser:
> hi @mewacaser I don't think this is the appropriate location, but, can I talk with you in private?

Sure

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
mewacaser: So I've been playing around with oriented for a while and in general I think it's an improvement. But I have found some situations where I'd rather it just be vertical so I can look up or down, like shafts. Also landing on single floating blocks is much easier and controlled when going up doesn't also mean changing your horizontal position.
Anyway, that's my opinion. I'm willing to change it however you feel best at the end of the day.
Or are you saying that oriented is pointless?

pupnewfster: Ah, I haven't looked yet at the changes to orientation so forgot that this might just be the old version.

omeranha:
> > hi @mewacaser I don't think this is the appropriate location, but, can I talk with you in private?
> 
> Sure

contact me in discord, please: oMeranha#8071

pupnewfster (changes requested):
Okay, so I finally got a chance to test this some and at least personally I don't feel like the changes to regular and then addition of ascend mode really fit that well in Mekanism (plus the movement in what is regular in the PR felt somewhat jerky and the like and was making me minorly nauseous. If you revert those changes though I will look into this PR again for purposes of it being more of a cleanup PR by use of creating the `IJetpackItem` interface.

mewacaser:
I agree that the refactor is entirely separate so I'll make a different pr for it. There's still a bunch to do on it though I think.

I'm still hoping you'll reconsider the thrust orientation though, everyone I had test it really liked it. So I'm gonna keep that branch around (but can close the pr if you want) on the off chance you come back around or decide it can be behind a config or something.

mewacaser:
I split the refactor into a separate pr.

Is there a good place I can get "pre-approval" on features before implementing them? I have more ideas, but some are iffy like this one on whether you would want them.

pupnewfster:
> So I'm gonna keep that branch around (but can close the pr if you want) on the off chance you come back around or decide it can be behind a config or something.

I will think about this feature more for 1.19 but I highly doubt it will make it into 1.18

> Is there a good place I can get "pre-approval" on features before implementing them? I have more ideas, but some are iffy like this one on whether you would want them.

Uh one of the easier spots to discuss things like that probably would be discord (especially in the testers channel as then it is easier for me to not miss the message if there is lots of chat going on)

mewacaser:
I finally got around to updating and I started messing around with different thrust levels for the jetpack. This actually revealed a real defect that was really hard to notice without drastically increased thrust (at least double). That is just for elytra flight though and wasn't really noticeable at normal thrust.

I also changed out the y velocity cap for diminishing returns which I like much better. This was pretty much necessary for the higher thrust levels that the mekasuit can now get with the changes to the jetpack module stack size but I still think it is better even with the standard jetpack.

I'm still very much in favor of the vector mode. All of the people (4) who played with it on a server we ran on a homebrew jar seemed to prefer it over the regular mode for most things. I could see making it a config option like the atomic disassembler vein mining though if you still don't like it. Also I think it did get a bit smoother with the thrust changes but I might just be imagining that.

pupnewfster:
For some reason I didn't notice the notification when you updated it. Would you mind retargeting this against the 1.20.4 branch, and then I will try and look into it and do some testing myself?

mewacaser:
I just did a quick rebase without testing to make sure it builds. But it seems low risk that it would break so just let me know if I need to fix it.

pupnewfster:
I will try it more tomorrow, rather than now when I am tired, but from a very initial impression, some times vector feels nicer when it comes to going forwards, and sometimes it seems to send you in very random directions. For example if you look up enough then it makes you start going backwards?

src/main/java/mekanism/common/content/gear/mekasuit/ModuleJetpackUnit.java
```diff
@@ -55,4 +61,34 @@ public void changeMode(IModule<ModuleJetpackUnit> module, Player player, ItemSta
...
+        return thrustMultiplier.get().getMultiplier();
+    }
+
+    @NothingNullByDefault
+    public enum ThrustMultiplier implements IHasTextComponent {
+        HALF(.5f),
+        NORMAL(1f),
+        FAST(2f),
+        FASTER(4f),
+        FASTEST(8f);
```
pupnewfster: I feel like rather than making this be based on how many are installed (or maybe in addition to doing so), we may want to increase the fuel usage for higher than normal thrust values.

src/main/java/mekanism/common/CommonPlayerTickHandler.java
```diff
@@ -126,8 +126,9 @@ private void tickEnd(Player player) {
             if (!primaryJetpack.isEmpty()) {
                 JetpackMode primaryMode = ((IJetpackItem) primaryJetpack.getItem()).getJetpackMode(primaryJetpack);
                 JetpackMode mode = IJetpackItem.getPlayerJetpackMode(player, primaryMode, () -> Mekanism.keyMap.has(player.getUUID(), KeySync.ASCEND));
+                double jetpackThrust = ((IJetpackItem) primaryJetpack.getItem()).getJetpackThrust(primaryJetpack);
```
pupnewfster: This can be moved past the disabled check unless the intention was to also consider it disabled if thrust is zero

src/main/java/mekanism/common/item/gear/ItemJetpack.java
```diff
@@ -87,6 +87,11 @@ public JetpackMode getJetpackMode(ItemStack stack) {
         return getMode(stack);
     }
 
+    @Override
+    public double getJetpackThrust(ItemStack stack) {
+        return 0.1D;
```
pupnewfster: I believe this used to be `0.15`, is the reason for this change just to balance it out a bit more with the vector mode or was it accidental?

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -113,33 +117,38 @@ private static ItemStack getJetpack(LivingEntity entity, Predicate<ItemStack> ma
...
-                      .add(forward.scale(1.5).subtract(motion).scale(0.5));
-                player.setDeltaMovement(motion.add(delta));
-                return false;
-            } else {
-                player.setDeltaMovement(motion.x(), Math.min(motion.y() + 0.15D, 0.5D), motion.z());
-            }
+        if ((mode == JetpackMode.NORMAL || mode == JetpackMode.VECTOR) && player.isFallFlying()) {
+            Vec3 forward = player.getLookAngle();
+            Vec3 drag = forward.scale(1.5).subtract(motion).scale(0.5);
+            Vec3 delta = forward.scale(thrust).add(drag);
```
pupnewfster, line 131: You mentioned on discord a bug your discovered with the elytra interaction. Was the issue basically the `forward.multiply(forward.scale(thrust))` vs `forward.scale(thrust)`? And if that is the case, what was wrong with how it was before?

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -113,33 +117,38 @@ private static ItemStack getJetpack(LivingEntity entity, Predicate<ItemStack> ma
...
+            player.setDeltaMovement(motion.add(delta));
```
pupnewfster: I believe this can be `player.addDeltaMovement(delta)`

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -113,33 +117,38 @@ private static ItemStack getJetpack(LivingEntity entity, Predicate<ItemStack> ma
...
+            return false;
+        } else if (mode == JetpackMode.NORMAL) {
+            player.setDeltaMovement(motion.x(), motion.y() + thrust * Math.min(1, Math.exp(-motion.y())), motion.z());
+        } else if (mode == JetpackMode.VECTOR) {
+            Vec3 forward = player.getLookAngle();
+            float theta = player.getYRot() * ((float) Math.PI / 180F);
```
pupnewfster: This can use the constant `Mth.DEG_TO_RAD`

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -113,33 +117,38 @@ private static ItemStack getJetpack(LivingEntity entity, Predicate<ItemStack> ma
...
+            Vec3 left = new Vec3(Mth.cos(theta), 0, Mth.sin(theta));
+            Vec3 up = forward.cross(left);
+            Vec3 thrustVec = up.scale(thrust);
+            player.setDeltaMovement(motion.add(new Vec3(thrustVec.x, thrustVec.y * Math.min(1, Math.exp(-motion.y())), thrustVec.z)));
```
pupnewfster: I believe this can be `player.addDeltaMovement(delta)`

pupnewfster (changes requested):
On a bit more testing it seems that vector definitely feels nicer in some situations, though behaves weird in some. For example:
- looking straight down
- activating it (pressing space) while falling feels very disorienting in regards to which directions it causes you to go
- Still feels a bit odd when if you look too high vertically it starts making you go backwards, but I think I sort of understand why it does

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
mewacaser: I was always confused about what it was doing before but it seemed to be working so I didn't mess with it. When the thrust was increased though it became much more obvious. At first I figured it had to do with when I switched to the vector operations but those are actually equivalent to what you originally had. I actually had to go back to the vanilla code for firework rockets to make sure I wasn't missing anything. I think when you first brought that code over you made temp variables with the forward component already multiplied into the thrust and then used them instead of the thrust? When the thrust is high and you're elytra flying it is really obvious though. It'll hold you in place depending on what direction you're looking at.

src/main/java/mekanism/common/content/gear/mekasuit/ModuleJetpackUnit.java
mewacaser: I like that idea. A potential issue there is that increasing the size of the hydrogen tank based on the number of modules isn't trivial. I know it's not necessary since at the same speed it would be able to go as far, but it does feel weird not to increase if it'll affect the time at max speed. (not that it will ever actually come up with the breathing unit refilling it)

src/main/java/mekanism/common/item/gear/ItemJetpack.java
mewacaser: This was an intentional change. This didn't really have anything to do with the vector mode, it was more for balancing with the exponential fall off change that also affects normal mode. I'm open to changing it to whatever you want but now that the y velocity isn't capped this has more relevance. Previously changing the thrust only changed if you reached the max y velocity in 2 vs. 3 ticks. So not a big change even it was reduced a lot.

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
mewacaser: I think that was already there unless I did that when I vectorized it. But I can swap it. Wasn't sure if there were other side effects.

pupnewfster: Yeah it was already like that, I just missed it when I converted some use cases into `addDeltaMovement` from doing a manual impl of that method.

src/main/java/mekanism/common/content/gear/mekasuit/ModuleJetpackUnit.java
mewacaser: I added the dynamic fuel usage. Since it also affected hover mode I changed hover to increase/decrease the ascension speed based on the thrust. I was surprised how fast the 8x used up the fuel so we might just want the usage to increase by 1 each tier instead of doubling with the thrust values if we can't increase the tank size.

pupnewfster: The `gasTankSpec` takes a `LongSupplier`. It should be simple enough to change that to a `ToLongFunction` that takes an `ItemStack`, and then in `ChemicalTankSpec#createTank` turn it into a long supplier using the stack that is passed in.

pupnewfster: I went ahead and pushed a commit that makes the tank size scale with the number of installed modules, and handles the case where the thrust usage would be higher than the amount of remaining gas (even if it is only a single usage so doesn't fully matter)

pupnewfster:
Posting this here mainly so I remember to look into it, but the slider kind of is breaking down for the module tweaker

image

pupnewfster:
I uhh set the thrust to 8 for the jetpack on the mekasuit (on regular mode) pressed space very briefly and ended up at over y = 8,000. That doesn't seem correct. Testing again, tapping it takes me to like 3,000 and then momentum carries me up to about 8,000 without any input from me.

Pressing once on hover takes me to 1,600, and tapping sneak instantly brings me to the ground?

Edit: Ignore me, I just messed up an if statement for clamping usage when there isn't enough stored.

pupnewfster:
I reverted the base thrust to `0.15` as the jetpack felt super slugish. This also made it so the vector mode didn't feel as off putting to me as I believe the issue was that it just didn't have enough thrust to reasonably be fighting gravity. Though I am a little unsure about the max thrust multiplier on the jetpack unit given it almost seems like it might be too much.

I also still need to do testing to see how all of this interacts with the elytra

pupnewfster:
I made the max thrust values a bit more linear and the maximum speed feels a bit more appropriate now. Main thing left I think is to figure out how to best handle the rendering of the modes in the module tweaker screen

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
pupnewfster: Yup, and seems like this also was related to #8015 which is convenient, as it saves me from having to look into the cause of that issue.

pupnewfster:
image

PR #7477, opened 2022-05-12
Added the Hydrostatic Repulsor Unit

mewacaser:
Added a module to the legs which stacks to 4 and increases speed in water.
First 3 levels give depth strider, last level gives toggleable swim boost (Dophin's Grace)


src/main/java/mekanism/client/ClientTickHandler.java
```diff
@@ -123,6 +125,11 @@ public static boolean isVisionEnhancementOn(Player player) {
         return module != null && module.isEnabled() && module.getContainerEnergy().greaterThan(MekanismConfig.gear.mekaSuitEnergyUsageVisionEnhancement.get());
     }
 
+    public static boolean isHydroEnhancementOn(Player player) {
+        IModule<ModuleHydrostaticRepulsorUnit> module = MekanismAPI.getModuleHelper().load(player.getItemBySlot(EquipmentSlot.LEGS), MekanismModules.HYDROSTATIC_REPULSOR_UNIT);
+        return module != null && module.isEnabled() && module.getInstalledCount() == 4 && module.getCustomInstance().isSwimBoost() && module.getContainerEnergy().greaterThan(MekanismConfig.gear.mekaSuitEnergyUsageHydrostaticRepulsion.get());
```
pupnewfster, line 130: I feel like the check for the installed count could potentially be moved into the `isSwimBoost` method

src/main/java/mekanism/client/ClientTickHandler.java
```diff
@@ -241,6 +248,19 @@ public void tickStart() {
                 }
             }
 
+            if (isHydroEnhancementOn(minecraft.player)) {
+                hydroEnhancement = true;
+                // adds if it doesn't exist, otherwise tops off duration to 220. equal or less than 200 will make it flicker
+                minecraft.player.addEffect(new MobEffectInstance(MobEffects.DOLPHINS_GRACE, 220, 0, false, false, false));
```
pupnewfster, line 254: This can probably just be `200` like the night vision module used to be and remove the comment? The reason why night vision is at 220 is that 200 corresponds to when night vision starts flickering to warn that the potion is running out.

src/main/java/mekanism/client/ClientTickHandler.java
```diff
@@ -241,6 +248,19 @@ public void tickStart() {
                 }
             }
 
+            if (isHydroEnhancementOn(minecraft.player)) {
```
pupnewfster, line 251: I haven't tested yet but part of me feels like this might not work quite right especially on dedicated servers as the effect is only being added client side. (Honestly I am slightly unsure why Aidan made it so the night vision potion is only client side instead of being fully server side controlled unless it is purely given the fact it doesn't really matter for that one given if a client modifies it to adjust how things behave there is a lot of the night vision logic that is client side anyway due to adjusting fog and color rendering)

pupnewfster (commented):
Left a few comments but one thing based on the PR description I am unsure if you are aware and intended is that the fourth level gives:
- depth strider **4** and dolphins grace
If this is not intended then you likely need to add into `EnchantmentBasedModule` a method like this:
```java
protected int getEnchantmentLevel(IModule<MODULE> module) {
    return module.getInstalledCount();
}
```
And then override it to be `return Math.min(super.getEnchantmentLevel(module), getEnchantment().getMaxLevel());`

Though part of me feels like adding a method like that would need a lot more care for ensuring it properly handles adding and removing and also docs stating that it expects it to be at least 1 for the returned thing.

src/main/java/mekanism/client/ClientTickHandler.java
mewacaser: I didn't want to add the module param to isSwimBoost and it was having trouble finding the getInstalledCount method on itself so I just left it. If you figure out the right way to call it from its own instance I'm definitely willing to make that change though.

src/main/java/mekanism/client/ClientTickHandler.java
mewacaser: I wasn't sure if it would flicker the effect if it was lower, but if it didn't I wasn't sure why not to go down to 1 tick? or maybe 2 depending on the order of clearing/updating effects.

src/main/java/mekanism/client/ClientTickHandler.java
mewacaser: So this is actually very similar in that all movement is controlled by the client. The server will kick if it thinks the movement is "flying" but as this just affects swim speed I haven't had it have issues yet, and I've tested on a server.

src/main/java/mekanism/common/content/gear/mekasuit/ModuleHydrostaticRepulsorUnit.java
```diff
@@ -0,0 +1,43 @@
...
+import net.minecraft.world.item.enchantment.Enchantment;
+import net.minecraft.world.item.enchantment.Enchantments;
+
+public class ModuleHydrostaticRepulsorUnit extends EnchantmentBasedModule<ModuleHydrostaticRepulsorUnit> {
+
+    private IModuleConfigItem<Boolean> swimBoost;
+
+    @Override
+    public void init(IModule<ModuleHydrostaticRepulsorUnit> module, ModuleConfigItemCreator configItemCreator) {
+        swimBoost = configItemCreator.createDisableableConfigItem("swim_boost", MekanismLang.MODULE_SWIM_BOOST, true, () -> module.getInstalledCount() == 4);
```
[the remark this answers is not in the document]
mewacaser: I'll switch to the constant and change to >= 4

mewacaser:
I was aware that it was giving depth strider IV, but I didn't think it was worth adding any additional checks on it. Vanilla functionality currently caps to 3 so it won't be any different for its intended purpose. I don't know how other mods treat it though so it's possible it would give even higher swim speeds with the right mods, but I don't think that is necessarily an issue either.

src/main/java/mekanism/client/ClientTickHandler.java
pupnewfster: Ahh you probably need to do `module.getCustomInstance().isSwimBoost(module)` similar to how most methods in `ICustomModule` take an `IModule` param

src/main/java/mekanism/client/ClientTickHandler.java
pupnewfster: Mostly because vanilla is kind of strange/then it accounts for a bit of lag by giving it an extra second as it is permanent when active and instantly removed when turned off so it isn't like a couple ticks vs a second make a difference there.

src/main/java/mekanism/client/ClientTickHandler.java
mewacaser: Yup, just realized the Enchant Module class doesn't extend or implement Module in any way... That was confusing. You still think it's worth it though, even with adding the module param? It does clean up the stack size checks.

src/main/java/mekanism/client/ClientTickHandler.java
mewacaser: Would it be better to extract a constant like PERM_POTION_TICKS or something and keep them consistent?

src/main/java/mekanism/client/ClientTickHandler.java
pupnewfster: Yeah probably

src/main/java/mekanism/client/ClientTickHandler.java
pupnewfster: Hmm probably. I also feel like for dolphins grace though (even though client side controls movement) it should be added server side instead of client side (and let MC sync it to the client presuming it updates it properly when we reapply it). That way if some anticheat type mod is installed it is less likely to bug out.

Though actually thinking about it maybe instead of adding an effect like this we want to use forge's swim speed attribute (`ForgeMod.SWIM_SPEED.get()`)? Then it could even stack when the player is near a dolphin. (See `PlayerState#updateStepAssist` for roughly how it could probably be added/applied properly)

mewacaser: I like that, I had initially discounted using it because it didn't affect the water push back like depth strider. But I like it much more than the potion effect

mewacaser: I switched to the attribute, and it looks much nicer.

To compare to the speed of the locomotion unit I turned the attribute addition to 5 or 6x normal swim speed. So I looked into how that unit is working. As far as I can tell it is literally adding the boost value as meters per tick to the player's movement. Do we want to also change that to use the MOVEMENT_SPEED attribute while we're at it?

pupnewfster: Maybe, I definitely think it would be cleaner if it used the attribute though we will need to tweak it to make sure it properly still uses energy and also look at some of the interactions as I am not sure what places check the attribute. (Basically our can function/our boost throttling for when in the air or in water and then also the jump throttling in `CommonPlayerTickHandler` for if it is active and the hydraulic propulsion unit is also active)

mewacaser:
Something I've been kicking around that's a bit related to this is the idea of a matching helmet/legs for the jetpack and free runner. I like how they match and work so well together, but it would be pretty cool if there was a whole starter gear set which upgraded into the modules for the mekasuit. And they could all have armored versions I guess in case someone didn't want to go to fission for some reason. If this is something you like I'm willing to do it, but my artistic ability caps out at the module items (and this one specifically is questionable comparing it to the inhalation unit).
I'm thinking the legs could be this or even better the locomotion effect. That would synergize pretty well I think with both the jetpack and free runners.
The obvious helmet choice is night vision I think, but even limiting to just one module's effect it maybe wouldn't be a big enough difference as these are all kinda sample versions of the mekasuit. Also it doesn't really synergize with the whole mobility aspect of the others. We could also do the water breathing, but I think it would have to not fill the jetpack and the scuba stuff already does that (could also add leggings for water speed as part of the scuba set). So maybe night vision is the way to go, and I think it also improves that module recipe.

pupnewfster:
Not currently sure on the rest of the gear for having something about at the same tier as free runners and the jetpack. I think it might make sense (and something we could probably get our artists to work on when they have time) but before creating any we would need to come up with better ideas of what we actually would want the pieces of gear to do.

mewacaser:
Let me know if you want to move this discussion somewhere else like an issue.

I like the locomotion boost idea for the legs, and I think it replacing diamond leggings improves that recipe.

Another idea for helmet could be reach distance, although I was thinking that might make more sense as a mekatool module.

pupnewfster:
Yeah it might be worth opening a feature request issue for tracking the idea of adding a corresponding legging and helmet piece and potential bonuses.

src/datagen/main/java/mekanism/client/lang/MekanismLangProvider.java
```diff
@@ -1296,6 +1297,7 @@ private void addMisc() {
         add(MekanismModules.GRAVITATIONAL_MODULATING_UNIT, "Gravitational Modulating Unit", "Using experimental technologies and the tremendous energy of antimatter, allows the user to defy gravity.");
         add(MekanismModules.ELYTRA_UNIT, "Elytra Unit", "Applies an HDPE Reinforced Elytra to the MekaSuit.");
         add(MekanismModules.GYROSCOPIC_STABILIZATION_UNIT, "Gyroscopic Stabilization Unit", "Allows the user to act as though they are on solid ground.");
+        add(MekanismModules.HYDROSTATIC_REPULSOR_UNIT, "Hydrostatic Repulsor Unit", "Allows the user to move through water with less resistance.");
```
pupnewfster, line 1300: Thoughts on if this might be better as a description?
> Uses advanced technology to repel water lowering the resistance felt while moving through it.

src/main/java/mekanism/client/ClientTickHandler.java
```diff
@@ -77,6 +78,7 @@ public class ClientTickHandler {
     public boolean shouldReset = false;
     public static boolean firstTick = true;
     public static boolean visionEnhancement = false;
+    public static boolean hydroEnhancement = false;
```
pupnewfster, line 81: Unused boolean

src/main/java/mekanism/client/ClientTickHandler.java
```diff
@@ -19,6 +19,7 @@
 import mekanism.common.base.HolidayManager;
 import mekanism.common.base.KeySync;
 import mekanism.common.config.MekanismConfig;
+import mekanism.common.content.gear.mekasuit.ModuleHydrostaticRepulsorUnit;
```
pupnewfster, line 22: Unused import

src/main/java/mekanism/common/content/gear/mekasuit/ModuleHydrostaticRepulsorUnit.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public void tickServer(IModule<ModuleHydrostaticRepulsorUnit> module, Player player) {
+        if (isSwimBoost(module)) {
+            module.useEnergy(player, MekanismConfig.gear.mekaSuitEnergyUsageHydrostaticRepulsion.get());
+        }
+    }
+
+    public Boolean isSwimBoost(IModule<ModuleHydrostaticRepulsorUnit> module) {
```
pupnewfster, line 42: This should return the primitive `boolean`

src/datagen/main/java/mekanism/common/recipe/impl/MekanismRecipeProvider.java
```diff
@@ -1465,6 +1465,17 @@ private void addGearModuleRecipes(Consumer<FinishedRecipe> consumer) {
               .key(Pattern.CONSTANT, Items.OBSIDIAN)
               .key(Pattern.HDPE_CHAR, MekanismItems.POLONIUM_PELLET)
               .build(consumer);
+        //Hydrostatic Repulsion Unit
+        ExtendedShapedRecipeBuilder.shapedRecipe(MekanismItems.MODULE_HYDROSTATIC_REPULSOR)
+              .pattern(RecipePattern.createPattern(
+                    TripleLine.of(Pattern.ALLOY, Pattern.CONSTANT, Pattern.ALLOY),
+                    TripleLine.of(Pattern.ALLOY, Pattern.PREVIOUS, Pattern.ALLOY),
+                    TripleLine.of(Pattern.HDPE_CHAR, Pattern.HDPE_CHAR, Pattern.HDPE_CHAR))
```
pupnewfster, line 1473: This can be `.pattern(BASIC_MODULE)`

pupnewfster (commented):
Seems pretty good, though to make the texture further apart from the inhalation purification unit and also go more with my proposed description change, I think the top part of the module's icon (the upper wave) should be flipped so as to *mirror* the "wave" below it instead of being the same wave just shifted up. (Basically move the low pixels of the upper wave up one and the high pixels down one)

src/main/java/mekanism/client/ClientTickHandler.java
mewacaser: Thought I'd backed out all the changes in ClientTickHandler, I'm squashing to make sure I didn't miss any this time.

mewacaser:
I changed the image just a little more than you suggested, actually mirroring on the full panel (moved the top one up a pixel).
Kinda makes me think of coils now too instead of just waves.

PR #7479, opened 2022-05-14
Fix floating cuboid in Atomic Disassembler model

hakanai:
I found this cuboid floating out in the middle of nowhere, but given its name, it seemed it was meant to be up at the same height as the other part with a similar name.

Where it was:
image

Where it is now:
image

- Fixes floating cuboid in rendering of Atomic Disassembler


pupnewfster:
Thanks for this, looks like [this call](https://github.com/mekanism/Mekanism/blob/1.16.x/src/main/java/mekanism/client/model/ModelAtomicDisassembler.java#L99) got lost in the port to 1.18. And it seems that at least in this case due to no rotations this PR is equivalent to using PartPose to apply the offset.

PR #7493, opened 2022-05-29
Jetpack Interface

mewacaser:
Made an interface for the jetpack to clean up some of the logic and simplify usage.

I think there is still a decent amount of cleanup to do on it. Currently it is finding the active jetpack several times a tick, at least on the client side, and some of those are embedded in the sound handling so it's not a straightforward refactor. I also haven't looked close enough yet, but think we have an issue when the gravitational module is active and a jetpack curio is equipped.


mewacaser:
I cleaned up a bunch of the remaining concerns. Some of the stuff it was doing I couldn't figure out why, so I changed it and made sure to test in dedicated server env and it worked. Also fixed the curio issues with gravitational module and it now uses the first jetpack's mode when there is a valid backup jetpack. We can wrap the primaryJetpack mode logic in a curio flag if you want though as currently it's impossible for it to be different than the active one without curios. But I think it's more performant now than before even without doing that.

Also wondering if we want to have the chest mode switch try to change a curio jetpack mode if it can't find a mode switch handler on the chest?

mewacaser:
If you're ok with the changes I made to the isJetpackOn where it just checks the state considering it refreshes the state at the start of each tick, would we also want to change the other ones with similar logic rather than recheck them on the client each time?

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -0,0 +1,163 @@
...
+     * If Curios is loaded, the curio slots will be checked as well.
+     *
+     * @param entity the entity on which to look for the jetpack
+     *
+     * @return the jetpack stack if present, otherwise an empty stack
+     */
+    @Nonnull
+    static ItemStack getPrimaryJetpack(LivingEntity entity) {
+        ItemStack chest = entity.getItemBySlot(EquipmentSlot.CHEST);
+        if (chest.getItem() instanceof IJetpackItem jetpackItem) {
```
pupnewfster, line 105: Unused variable can just be an instanceof

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -0,0 +1,163 @@
...
+            return chest;
+        } else if (Mekanism.hooks.CuriosLoaded) {
+            return CuriosIntegration.findFirstCurio(entity, s -> s.getItem() instanceof IJetpackItem jetpackItem);
```
pupnewfster, line 108: Unused variable can just be an instanceof

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -0,0 +1,163 @@
...
+    @Nonnull
+    static ItemStack getActiveJetpack(LivingEntity entity) {
+        ItemStack chest = entity.getItemBySlot(EquipmentSlot.CHEST);
+        if (chest.getItem() instanceof IJetpackItem jetpackItem && jetpackItem.isActive(chest)) {
+            return chest;
+        } else if (Mekanism.hooks.CuriosLoaded) {
+            return CuriosIntegration.findFirstCurio(entity, s -> s.getItem() instanceof IJetpackItem jetpackItem && jetpackItem.isActive(s));
+        }
+        return ItemStack.EMPTY;
+    }
```
pupnewfster, line 91: I feel like this method and the one below could have a private helper that they use like this:
```java
private static ItemStack getJetpack(LivingEntity entity, Predicate<ItemStack> matcher) {
    ItemStack chest = entity.getItemBySlot(EquipmentSlot.CHEST);
    if (matcher.test(chest)) {
        return chest;
    } else if (Mekanism.hooks.CuriosLoaded) {
        return CuriosIntegration.findFirstCurio(entity, matcher);
    }
    return ItemStack.EMPTY;
}
```

src/main/java/mekanism/client/ClientTickHandler.java
```diff
@@ -205,17 +208,12 @@ public void tickStart() {
...
             }
 
-            ItemStack chestStack = minecraft.player.getItemBySlot(EquipmentSlot.CHEST);
-            IModule<ModuleJetpackUnit> jetpackModule = MekanismAPI.getModuleHelper().load(chestStack, MekanismModules.JETPACK_UNIT);
-
-            ItemStack jetpack = ItemJetpack.getJetpack(minecraft.player, chestStack);
-            if (!jetpack.isEmpty() || jetpackModule != null) {
+            if (!jetpack.isEmpty()) {
+                ItemStack primaryJetpack = IJetpackItem.getPrimaryJetpack(minecraft.player);
+                JetpackMode primaryMode = ((IJetpackItem) primaryJetpack.getItem()).getJetpackMode(primaryJetpack);
```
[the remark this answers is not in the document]
mewacaser: it doesn't get into here if there isn't an active jetpack (usable), in which case there will always be at least a first jetpack

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -0,0 +1,163 @@
...
+
+public interface IJetpackItem {
+
+    boolean canUseJetpack(ItemStack stack);
+
+    JetpackMode getJetpackMode(ItemStack stack);
+
+    void useJetpackFuel(ItemStack stack);
+
+    default boolean isActive(ItemStack stack) {
```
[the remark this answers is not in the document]
mewacaser: so my initial thought was to use isActive to include stuff like the gravity module check, but that's now in canUse so I might just merge them until we find a reason to split them again

src/main/java/mekanism/client/ClientTickHandler.java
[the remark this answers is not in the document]
pupnewfster: Yes I get that and if you look at the issue I linked you will see that quite a few people were running into issues where somehow the getStackInSlot(chest) was changing partway through which is why even though it shouldn't be possible I would rather have an extra check here.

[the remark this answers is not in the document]
mewacaser: oh, didn't look close enough at the issue to see that. That is awful, but I'll dummy proof it...

src/main/java/mekanism/common/CommonPlayerTickHandler.java
```diff
@@ -124,23 +119,24 @@ private void tickEnd(Player player) {
...
-                player.fallDistance = 0.0F;
-                if (player instanceof ServerPlayer serverPlayer) {
-                    serverPlayer.connection.aboveGroundTickCount = 0;
+        ItemStack jetpack = IJetpackItem.getActiveJetpack(player);
+        if (!jetpack.isEmpty()) {
+            ItemStack primaryJetpack = IJetpackItem.getPrimaryJetpack(player);
+            JetpackMode primaryMode = ((IJetpackItem) primaryJetpack.getItem()).getJetpackMode(primaryJetpack);
+            JetpackMode mode = IJetpackItem.getPlayerJetpackMode(player, primaryMode);
+            if (mode != JetpackMode.DISABLED) {
+                // Client handles motion and sets fallDistance in ClientTickHandler at the tick start
```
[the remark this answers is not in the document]
mewacaser: This is just condensed return logic for the handle motion call, so if we need to change this back it should just be replacing the method out. I almost left it just so it wouldn't be a hassle to find if the logic changed for some reason.

Speaking of though, I think I finally figured out why vanilla doesn't let you stop using an elytra mid flight (bedrock does and running into something also does it automatically). I was curious how much fall damage or energy would be used falling 1k+ meters in the mekasuit and it turns out not much since it disconnects you every 200 meters (resetting your fall damage) as flying is not enabled on the server.

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
```diff
@@ -0,0 +1,157 @@
...
+    static ItemStack getPrimaryJetpack(LivingEntity entity) {
+        return getJetpack(entity, stack -> stack.getItem() instanceof IJetpackItem);
+    }
+
+    private static ItemStack getJetpack(LivingEntity entity, Predicate<ItemStack> matcher) {
+        ItemStack chest = entity.getItemBySlot(EquipmentSlot.CHEST);
+        if (matcher.test(chest)) {
+            return chest;
+        } else if (Mekanism.hooks.CuriosLoaded) {
+            return CuriosIntegration.findFirstCurio(entity, s -> matcher.test(s));
```
pupnewfster, line 102: You should be able to just pass matcher directly as a parameter to `findFirstcurio` or is there something I am overlooking?

src/main/java/mekanism/common/CommonPlayerTickHandler.java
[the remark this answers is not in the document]
pupnewfster: I think I would definitely prefer you calling the method on the server as well as I recall with some things related to boosting the server having issues at times due to the moved wrongly/too quickly thing and this way the server would know they are supposed to be moving quicker.

I am not quite sure what you are talking about the elytra and being kicked for flying though?

src/main/java/mekanism/common/item/interfaces/IJetpackItem.java
mewacaser: nope, just missed it

src/main/java/mekanism/common/CommonPlayerTickHandler.java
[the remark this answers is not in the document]
mewacaser: I left the comment as I think it still explains why we are only doing this for the server.

The flying/kick thing isn't really something we can fix so I guess not especially relevant, was more just stating that as you can stop elytra flight with the mekasuit it is easy to fall 200m and disconnect. But I guess that problem exists with just the jetpack too.

src/main/java/mekanism/common/CommonPlayerTickHandler.java
```diff
@@ -124,23 +119,24 @@ private void tickEnd(Player player) {
...
+        ItemStack jetpack = IJetpackItem.getActiveJetpack(player);
+        if (!jetpack.isEmpty()) {
+            ItemStack primaryJetpack = IJetpackItem.getPrimaryJetpack(player);
+            if (!primaryJetpack.isEmpty()) {
+                JetpackMode primaryMode = ((IJetpackItem) primaryJetpack.getItem()).getJetpackMode(primaryJetpack);
+                JetpackMode mode = IJetpackItem.getPlayerJetpackMode(player, primaryMode, () -> Mekanism.keyMap.has(player.getUUID(), KeySync.ASCEND));
+                if (mode != JetpackMode.DISABLED) {
+                    // Client handles motion and sets fallDistance in ClientTickHandler at the tick start
+                    if (player instanceof ServerPlayer serverPlayer && IJetpackItem.handleJetpackMotion(player, mode, () -> Mekanism.keyMap.has(player.getUUID(), KeySync.ASCEND))) {
+                        player.fallDistance = 0.0F;
```
pupnewfster, line 131: This should reset fall distance regardless of the player is a server player (even though I am not sure there are any cases that will happen). See the code flow before this PR:
handle -> fall distance -> if server player -> above ground tick count

mewacaser: I know I changed it, but client already calls handleJetpackMotion and conditionally resets fall distance every tick in ClientTickHandler. Is there any reason we want to have it call it at tick start and tick end? Or should the ClientTickHandler stuff be moved to common as they're basically doing the same thing. I didn't take it that far as I wasn't sure if it was doing it on tick start for the client intentionally

pupnewfster: I just meant do this except do the fallDistance = 0 regardless of whether `player instanceof ServerPlayer` in this check. Also this `tickEnd` check only runs on the server side, unlike the client stuff which is on the client side.

src/main/java/mekanism/common/item/gear/ItemMekaSuitArmor.java
```diff
@@ -329,6 +330,25 @@ public boolean elytraFlightTick(ItemStack stack, LivingEntity entity, int flight
         return true;
     }
 
+    @Override
+    public boolean canUseJetpack(ItemStack stack) {
+        return slot == EquipmentSlot.CHEST && (isModuleEnabled(stack, MekanismModules.JETPACK_UNIT) && ChemicalUtil.hasChemical(stack, MekanismGases.HYDROGEN.get()) || isModuleEnabled(stack, MekanismModules.GRAVITATIONAL_MODULATING_UNIT));
```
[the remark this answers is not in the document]
mewacaser: I added this as a comment so you could review before the other pr merges.

src/main/java/mekanism/common/CommonPlayerTickHandler.java
mewacaser: ok, it's all reverted

src/main/java/mekanism/common/item/gear/ItemMekaSuitArmor.java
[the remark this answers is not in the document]
pupnewfster: I think the comment you added looks fine for the implementation of it afterwards other than I think there is a missing check related to validating the jetpack has fuel? Which likely would need to be shoved in `ItemMekaSuitArmor#getJetpackMode` and pretend to be disabled if there is no fuel present.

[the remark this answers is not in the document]
mewacaser: my intention was actually to leave the jetpack check alone and just replace the gravity one with that. I'm still a fan of allowing a curio backup jetpack if the mekasuit ever ran out of fuel.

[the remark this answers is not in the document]
pupnewfster: Sounds good

pupnewfster:
Okay, once this is rebased and uses the new exclusive module stuff I will give this one last quick test and then merge it.

src/main/java/mekanism/common/item/gear/ItemMekaSuitArmor.java
```diff
@@ -329,6 +331,26 @@ public boolean elytraFlightTick(ItemStack stack, LivingEntity entity, int flight
         return true;
     }
 
+    @Override
+    public boolean canUseJetpack(ItemStack stack) {
+        return slot == EquipmentSlot.CHEST && (isModuleEnabled(stack, MekanismModules.JETPACK_UNIT) && ChemicalUtil.hasChemical(stack, MekanismGases.HYDROGEN.get())
```
pupnewfster, line 336: Upon closer inspection I think this long line needs to be split into multiple if statements because if the jetpack is enabled but there is no chemical THEN it will check the OR which will return true because the jetpack is enabled and exclusive for overriding jump.

mewacaser: I fixed it, but should we consider adding a canUse function to module? It could be used for a couple things which require energy or something else, including the canUseJetpack method. It would put the logic in one place in case it ever changes too.

pupnewfster: I think it might be worth at least considering, but I think that will end up waiting for 1.19

PR #7497, opened 2022-05-31
Exclusive Module channels

mewacaser:
Refactored the exclusive module logic to allow different channels of exclusivity.


src/main/java/mekanism/common/registries/MekanismModules.java
```diff
@@ -54,19 +55,17 @@ private MekanismModules() {
...
                 .rendersHUD());
     public static final ModuleRegistryObject<ModuleFarmingUnit> FARMING_UNIT = MODULES.register("farming_unit", ModuleFarmingUnit::new,
-          () -> MekanismItems.MODULE_FARMING.asItem(), builder -> builder.maxStackSize(4).rarity(Rarity.UNCOMMON).exclusive());
-    //TODO: Eventually we may want to come up with a better exclusive method given realistically the shearing unit and farming unit don't need to be
-    // exclusive of each other, but they both should be exclusive in regards to the teleportation unit
+          () -> MekanismItems.MODULE_FARMING.asItem(), builder -> builder.maxStackSize(4).rarity(Rarity.UNCOMMON).exclusive(ExclusiveFlag.getFlags(ExclusiveFlag.INTERACT_BLOCK)));
     public static final ModuleRegistryObject<ModuleShearingUnit> SHEARING_UNIT = MODULES.register("shearing_unit", ModuleShearingUnit::new,
-          () -> MekanismItems.MODULE_SHEARING.asItem(), builder -> builder.rarity(Rarity.UNCOMMON).exclusive());
+          () -> MekanismItems.MODULE_SHEARING.asItem(), builder -> builder.rarity(Rarity.UNCOMMON).exclusive(ExclusiveFlag.getFlags(ExclusiveFlag.INTERACT_ENTITY)));
```
pupnewfster, line 60: Might be worth a comment that while you can shear blocks in general they don't overlap with anything that supports the farming unit right click.

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -86,11 +87,18 @@ public final int getMaxStackSize() {
...
      *
      * @return {@code true} if this module type is exclusive.
      */
-    public final boolean isExclusive() {
+    public final boolean isExclusive(int flags) {
+        return (exclusive & flags) != 0;
+    }
+
+    /**
+     * Gets the exclusive flags for this module type.
```
pupnewfster, line 99: Add a see tag:
`@see ExclusiveFlag`

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -86,11 +87,18 @@ public final int getMaxStackSize() {
     }
 
     /**
-     * Exclusive modules only work one-at-a-time; when one is enabled, others will be automatically disabled.
+     * Exclusive modules only work one-at-a-time; when one is enabled, incompatible others will be automatically disabled.
      *
      * @return {@code true} if this module type is exclusive.
      */
-    public final boolean isExclusive() {
+    public final boolean isExclusive(int flags) {
```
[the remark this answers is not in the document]
mewacaser: I could actually see just having the no param call, but it does make the calls less explicit so if you don't like that we don't have to and we could set it deprecated. Or it can wait I guess, just let me know which you prefer.

src/main/java/mekanism/common/registries/MekanismModules.java
mewacaser: With that in mind though, do we want to split the INTERACT_BLOCK flag to clarify that?

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -275,4 +283,20 @@ public ModuleDataBuilder<MODULE> disabledByDefault() {
...
+
+    public enum ExclusiveFlag {
+        INTERACT_EMPTY,
+        INTERACT_ENTITY,
+        INTERACT_BLOCK,
+        OVERRIDE_JUMP;
+
+        // This will work for 32 flags, if we go past that we will have to switch it to a long. But that is a lot of flags!
+        public static int getFlags(ExclusiveFlag... flags) {
+            return Arrays.stream(flags).mapToInt(Enum::ordinal).reduce(0, (result, flag) -> result | (1 << flag));
```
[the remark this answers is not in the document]
mewacaser: Are you talking about for performance concerns or readability? As this is only called for registration, I don't know that performance is really an issue. I've done flags explicitly and with constants and never liked how it looked, and it was always annoying to add/swap orders on the items. There are a couple things about this implementation that I really like, it's clear what the flags are and how to add another, getting compound masks is also easy without messing with bitwise yourself and we can also change the data type relatively easily.

But you're right adding a getMask function is a good idea, we just don't need to add it explicitly to the enum values.

src/api/java/mekanism/api/gear/ModuleData.java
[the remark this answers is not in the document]
pupnewfster: Might as well just deprecate it in case anyone ends up making 1.18 addons that add custom modules that they would find this of use.

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -86,11 +87,18 @@ public final int getMaxStackSize() {
     }
 
     /**
-     * Exclusive modules only work one-at-a-time; when one is enabled, others will be automatically disabled.
+     * Exclusive modules only work one-at-a-time; when one is enabled, incompatible others will be automatically disabled.
```
pupnewfster, line 90: `incompatible modules will be automatically disabled`

src/api/java/mekanism/api/gear/ModuleData.java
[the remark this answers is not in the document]
pupnewfster: Readability in terms of seeing what corresponds to which. Also given the fact this is exposed as API (so should also have javadocs, the specific flags should also have javadocs briefly explaining them) you cannot state this is only used for initialization as it may be used at any time.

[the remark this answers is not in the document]
mewacaser: True, I only meant that it's currently only called then. But it's not like it's expensive anyway.

With the getMask function added what do you think about keeping at as a basic enum?

src/main/java/mekanism/common/registries/MekanismModules.java
```diff
@@ -85,12 +86,12 @@ private MekanismModules() {
           ModuleGeigerUnit::new, () -> MekanismItems.MODULE_GEIGER.asItem(), builder -> builder.rarity(Rarity.UNCOMMON).rendersHUD());
     public static final ModuleRegistryObject<ModuleJetpackUnit> JETPACK_UNIT = MODULES.register("jetpack_unit",
           ModuleJetpackUnit::new, () -> MekanismItems.MODULE_JETPACK.asItem(), builder -> builder.rarity(Rarity.RARE).handlesModeChange().rendersHUD()
-                .exclusive(ExclusiveFlag.getFlags(ExclusiveFlag.OVERRIDE_JUMP)));
+                .exclusive(ExclusiveFlag.OVERRIDE_JUMP.getMask()));
```
pupnewfster, line 89: Thoughts on a helper in the builder to just take a vararg of exclusive flag?

mewacaser: I like it, I'm thinking we just have it as a wrapper for the other call though. That also takes care of the deprecation I think.

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -86,11 +87,30 @@ public final int getMaxStackSize() {
...
      */
+    @Deprecated(forRemoval = true, since = "10.2.3")
     public final boolean isExclusive() {
+        return isExclusive(ExclusiveFlag.ANY);
+    }
+
+    /**
+     * Exclusive modules only work one-at-a-time; when one is enabled, incompatible modules will be automatically disabled.
+     *
+     * @param mask {@link ExclusiveFlag} mask
```
pupnewfster, line 102: `@param mask Mask of all {@link ExclusiveFlag flags} to check exclusivity against.`

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -86,11 +87,30 @@ public final int getMaxStackSize() {
...
+     *
+     * @return {@code true} if this module type is exclusive.
```
pupnewfster, line 104: `... is exclusive of the given flags.`

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -275,4 +314,61 @@ public ModuleDataBuilder<MODULE> disabledByDefault() {
             return this;
         }
     }
+
+    /**
+     * Enum of flags for module exclusivity channels
+     */
+    public enum ExclusiveFlag {
+
```
pupnewfster, line 322: I don't think the line breaks between the enum values (and also this initial one between the enum declaration and the first javadoc) are needed or all that beneficial.

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -275,4 +314,61 @@ public ModuleDataBuilder<MODULE> disabledByDefault() {
...
+         */
+        INTERACT_BLOCK,
+
+        /**
+         * This flag indicates that this module changes what pressing jump does
+         */
+        OVERRIDE_JUMP;
+
+        // This will work for 32 flags, if we go past that we will have to switch it to a long. But that is a lot of flags!
+
```
pupnewfster, line 344: Spacing around this comment is a bit odd, feels like it got somewhat lost due to it not being a comment on anything in particular

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -275,4 +314,61 @@ public ModuleDataBuilder<MODULE> disabledByDefault() {
...
+        /**
+         * Gets the mask for this flag
+         */
+        public int getMask() {
+            return 1 << ordinal();
+        }
+
+        /**
+         * Gets the mask for the combination of all the input flags
```
pupnewfster, line 353:
```
Helper to get the mask of the combination of the given input flags

@param flags {@link ExclusiveFlag Flags} to combine into a mask.

@returns Mask representing all the given {@link ExclusiveFlag flags}.
```

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -275,4 +314,61 @@ public ModuleDataBuilder<MODULE> disabledByDefault() {
...
+         */
+        public static int getCompoundMask(ExclusiveFlag... flags) {
+            return Arrays.stream(flags).mapToInt(ExclusiveFlag::getMask).reduce(0, (result, mask) -> result | mask);
```
pupnewfster, line 356: I feel like it would be slightly more readable as `reduce(NONE, ...`

src/api/java/mekanism/api/gear/ModuleData.java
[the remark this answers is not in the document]
pupnewfster: If you mean keeping it as a basic enum vs just making it a bunch of static ints.. I am unsure part of me feels like ints would be a bit less expensive to read than getting the ordinal and doing a bit shift each time you want to see if something conflicts with it, but I am open to either for now at least until there is some use case where it ends up as a hotspot.

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -224,16 +244,35 @@ public ModuleDataBuilder<MODULE> maxStackSize(int maxStackSize) {
...
+            exclusive = mask;
             return this;
         }
 
         /**
-         * Marks this module type as being able to handle mode changes. In addition to using this method {@link ICustomModule#changeMode(IModule, Player, ItemStack, int,
-         * boolean)} should be implemented.
+         * Marks this module type as exclusive. Exclusive modules only work one-at-a-time; when one is enabled, incompatible modules will be automatically disabled.
+         *
+         * @param flags {@link ExclusiveFlag} flags for the exclusive mask
```
pupnewfster, line 267: On a new line in the javadocs: `@since 10.2.3`

src/api/java/mekanism/api/gear/ModuleData.java
mewacaser: I just got rid of it, most people know sizes I think if it becomes a problem which it most likely won't

src/main/java/mekanism/common/content/gear/ModuleConfigItem.java
```diff
@@ -52,10 +53,8 @@ public void set(@Nonnull TYPE val, @Nullable Runnable callback) {
             boolean checkModeState;
             if (name.equals(Module.ENABLED_KEY) && val == Boolean.TRUE) {
                 // disable other exclusive modules
-                if (module.getData().isExclusive()) {
-                    if (m.getData().isExclusive() && m.getData() != module.getData()) {
-                        m.setDisabledForce(callback != null);
-                    }
+                if (m.getData().isExclusive(module.getData().getExclusiveFlags()) && m.getData() != module.getData()) {
```
pupnewfster, line 56: I know it wasn't that way before but I feel like we might as well flip the order of this if statement and check `m.getData() != module.getData()` first before checking if they are exclusive of each other.

pupnewfster (commented):
Looks pretty good, haven't tested it yet but one thing I am wondering is if you have any ideas how we could maybe make it clearer to users what types of things it is exclusive between rather than just saying "exclusive". Doesn't really matter in terms of I am not considering it a blocker for this PR getting merged, just wondering if maybe you have any good ideas how to represent it,

src/main/java/mekanism/common/registries/MekanismModules.java
mewacaser: I mean we can just split Interact block to shearable/not and let addons deal with it. It would kinda be on them if they made the farming module do something to a shearable block if we set its exclusivity to say it doesn't. Also they could potentially override its mask to say it does.

Somewhat related, what do you think about trying to deal with dynamic exclusivity. For example the teleport module could technically just be INTERACT_BLOCK if the requires block config is active. I think it's doable, we'd just have to check exclusivity with most of the config changes and they'd be defining a mask supplier instead of a constant mask.

mewacaser:
I added a helper to get conflicting modules. I don't think it's exactly what you meant but I also added it to the details in inventory, but didn't make it "pretty" yet as I'm not sure you'll want it there.

pupnewfster:
Hmm at a glance from the code I think that potentially could work, though realistically we would likely want the conflicting module list to be a lazily calculated map/cache. Also if we are showing it in the tooltip like that as part of the cleanup of making it use lang keys and such, we may want to make it to `ItemBlockTooltip` in that having one button to show supported items and one to show conflicts. Alternatively we could put the list of conflicts in JEI when viewing usages of the item (using jei's info tab)

src/main/java/mekanism/common/registries/MekanismModules.java
pupnewfster: I feel like dynamic exclusivity isn't worth the headache of taking into account then disabling other conflicting modules when other sub settings are changed. And I think just make the shear one for now also have INTERACT_BLOCK given some mod may add a block that can be both sheared and tilled, hoed, etc

mewacaser:
It's got a lazy cache now, but I'm still not sure what to do with showing the conflicts though. I don't know about adding another button to show them as everything else just has description/details and it does fit in the details category. I could see them in JEI info but I'm not set on it either as vanilla enchants don't even show conflicts there.

pupnewfster:
Up to you if you clean it up and give it lang keys or just remove the stuff you did for adding it to the tooltip and put a TODO so that we can figure it out later.

src/main/java/mekanism/common/registries/MekanismModules.java
```diff
@@ -54,19 +55,18 @@ private MekanismModules() {
           ModuleAttackAmplificationUnit::new, () -> MekanismItems.MODULE_ATTACK_AMPLIFICATION.asItem(), builder -> builder.maxStackSize(4).rarity(Rarity.UNCOMMON)
                 .rendersHUD());
     public static final ModuleRegistryObject<ModuleFarmingUnit> FARMING_UNIT = MODULES.register("farming_unit", ModuleFarmingUnit::new,
-          () -> MekanismItems.MODULE_FARMING.asItem(), builder -> builder.maxStackSize(4).rarity(Rarity.UNCOMMON).exclusive());
-    //TODO: Eventually we may want to come up with a better exclusive method given realistically the shearing unit and farming unit don't need to be
-    // exclusive of each other, but they both should be exclusive in regards to the teleportation unit
+          () -> MekanismItems.MODULE_FARMING.asItem(), builder -> builder.maxStackSize(4).rarity(Rarity.UNCOMMON).exclusive(ExclusiveFlag.INTERACT_BLOCK));
+    // While you can shear blocks, in general they don't overlap with anything that supports the farming unit right click.
```
pupnewfster, line 59: This comment should go away now that we are actually marking it as exclusive for blocks as well.

src/main/java/mekanism/common/item/ItemModule.java
```diff
@@ -46,9 +49,17 @@ public Rarity getRarity(@Nonnull ItemStack stack) {
     @Override
     public void appendHoverText(@Nonnull ItemStack stack, Level world, @Nonnull List<Component> tooltip, @Nonnull TooltipFlag flag) {
         if (MekKeyHandler.isKeyPressed(MekanismKeyHandler.detailsKey)) {
+            tooltip.add(new TextComponent("Supported by:"));
```
pupnewfster, line 52: Converting this and conflicting to translation keys should be fine, especially if you add/rename `MekanismLang.UPGRADE_DISPLAY` to `MekanismLang.GENERIC_LIST` and then move it to the generic section and use that for the elements being added to each sub category.

src/main/java/mekanism/common/item/ItemModule.java
```diff
@@ -46,9 +49,17 @@ public Rarity getRarity(@Nonnull ItemStack stack) {
...
             for (Item item : MekanismAPI.getModuleHelper().getSupported(getModuleData())) {
                 tooltip.add(item.getName(new ItemStack(item)));
             }
+            Set<ModuleData<?>> conflicting = MekanismAPI.getModuleHelper().getConflicting(getModuleData());
```
pupnewfster, line 56: Might as well put `MekanismAPI.getModuleHelper()` in a local variable given there are now multiple calls to it in this method.

mewacaser:
if I end up making another commit I'll rebase, otherwise just let me know if you want me to do that or if you just manually resolve the cache stuff

src/main/java/mekanism/common/item/ItemModule.java
```diff
@@ -49,15 +50,16 @@ public Rarity getRarity(@Nonnull ItemStack stack) {
...
             }
-            Set<ModuleData<?>> conflicting = MekanismAPI.getModuleHelper().getConflicting(getModuleData());
-            if(!conflicting.isEmpty()){
-                tooltip.add(new TextComponent("Conflicting:"));
+            Set<ModuleData<?>> conflicting = moduleHelper.getConflicting(getModuleData());
+            if (!conflicting.isEmpty()) {
+                tooltip.add(MekanismLang.MODULE_CONFLICTING.translateColored(EnumColor.RED));
                 for (ModuleData<?> module : conflicting) {
-                    tooltip.add(module.getTextComponent());
+                    tooltip.add(MekanismLang.GENERIC_LIST.translate(module.getTextComponent()));
```
pupnewfster, line 62: Explicit `getTextComponent` call shouldn't be needed anymore

mewacaser: lol, even thought of that when I was renaming the GENERIC_LIST just forgot when putting it in

pupnewfster:
Also rebasing would be of use so then I can just resolve instead of have to rerun the datagen myself after merging to fix what values the cache has.

mewacaser:
So I rebased, but it was having trouble with runData that I didn't take the time to really address. Something about a InvocationTargetException, not thinking it was something I did, but it was saying something about the block registry.

pupnewfster:
Ah I see why it is due to one of the deps running into an issue with a change forge made recently that made an accidental breaking change.

PR #7517, opened 2022-06-14
Add another skin

agnor99:
Adding another skin

I don't like the location of the custom RobitSkin class, but I'm unsure, where to put it, as there was no really good package for it


src/main/java/mekanism/common/entity/AllayRobitSkin.java
```diff
@@ -0,0 +1,28 @@
+package mekanism.common.entity;
```
pupnewfster, line 1: Add a package in entity call skin or skins or something like that

src/main/java/mekanism/common/entity/AllayRobitSkin.java
```diff
@@ -0,0 +1,28 @@
...
+import net.minecraft.world.entity.player.Player;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class AllayRobitSkin extends RobitSkin {
+
+    public AllayRobitSkin() {
+        super(new ResourceLocation[]{
+                Mekanism.rl( "allay"), Mekanism.rl("allay2")}
+        );
```
pupnewfster, line 15: Just use the varargs no need for explicit array creation.

src/main/java/mekanism/common/entity/AllayRobitSkin.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Nullable
+    @Override
+    public ResourceLocation getCustomModel() {
+        return Mekanism.rl("item/robit_allay");
```
pupnewfster, line 21: `private static final ResourceLocation MODEL =` (This method gets called a fair bit iirc)

src/main/java/mekanism/common/network/to_server/PacketRobit.java
```diff
@@ -24,7 +24,8 @@ public class PacketRobit implements IMekanismPacket {
 
     private static final Map<String, List<IRobitSkinProvider>> EASTER_EGGS = Map.of(
             "sara", getPrideSkins(RobitPrideSkinData.TRANS, RobitPrideSkinData.LESBIAN),
-            "agnor", getPrideSkins(RobitPrideSkinData.GAY)
+            "agnor", getPrideSkins(RobitPrideSkinData.GAY),
+            "allay", List.<IRobitSkinProvider>of(MekanismRobitSkins.ALLAY)
```
pupnewfster, line 28: I think this is unnecessary. Given at a certain point having name just go to the skin itself makes things really not be an *easter egg*

src/main/java/mekanism/common/registries/MekanismRobitSkins.java
```diff
@@ -27,4 +27,5 @@ private MekanismRobitSkins() {
         }
         return Collections.unmodifiableMap(internal);
     });
+    public static final RobitSkinRegistryObject<RobitSkin> ALLAY = ROBIT_SKINS.register("allay", AllayRobitSkin::new);
```
pupnewfster, line 30: This can just go after the base line for readability. Also it looks like you are missing the mekanism lang provider entries for this?

src/main/java/mekanism/common/entity/AllayRobitSkin.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Override
+    public boolean isUnlocked(@NotNull Player player) {
+        return false;
+    }
```
pupnewfster, line 27: Unneeded this is just what super is, unless we want to try require finding an allay before players can use this

src/main/java/mekanism/common/registries/MekanismRobitSkins.java
pupnewfster: Maybe something along the lines of "Allay Costume" ?

src/main/java/mekanism/common/entity/AllayRobitSkin.java
agnor99: that would be true, false is not unlocked (so only available by name)

src/datagen/main/java/mekanism/client/lang/MekanismLangProvider.java
```diff
@@ -417,6 +417,7 @@ private void add(MekanismDamageSource damageSource, String value, String valueEs
 
     private void addRobitSkins() {
         add(MekanismRobitSkins.BASE, "Default");
+        add(MekanismRobitSkins.ALLAY, "Allay");
```
pupnewfster, line 420: Imo it might be better as "Allay Costume"

src/main/java/mekanism/common/entity/AllayRobitSkin.java
agnor99: just once for each skin and once per frame for each rendered item, but will change that

src/main/java/mekanism/common/entity/AllayRobitSkin.java
pupnewfster: Ah hmm that is true I need to think on this more, given setting the name to allay any only making it change the skin if it is on base feels weird. Part of me would rather there is actually some unlock condition  and setting by name doesn't adjust the skin

agnor99: Entitiy Kills are tracked, so killing an allay should be a possible unlock condition.
writing that didn't feel great tbh

pupnewfster: Part of me initially was thinking player has advancement `husbandry/allay_deliver_item_to_player` but to check that needs a server player and I don't think we are guaranteed that in context everywhere we call it (though if we were then we could easily changed during the breaking change window to having the param be a server player... but I don't believe we can) Though I wonder if the context is accurate enough to if server player check and otherwise assume true on the client side? That might work honestly as I don't think it tends to be called with the player on the client side.

agnor99: I couldn't find clientside validation

agnor99: Wait there is

agnor99: Should a chat message appear on completion of the advancement, that this skin is now available?

pupnewfster: Na, I don't think that is needed. If you really want you could put a `//TODO: Do we eventually want to make a system for announcing unlocks, maybe using toast notifications` but I don't think there is any real need to *do* so  currently

PR #7552, opened 2022-07-14
Fix energy dupe bug while charging stacks of induction cells

GotoFinal:

Currently you can stack induction cells, but charging any stack of mekanism items will charge all items in stack for the cost of one, allowing to dupe huge amounts of energy in very short time - especially when combined with much faster wireless charging mods like flux networks.

Also makes induction cells display sum of data to represent real state of what is inside the slot.

### DIFF
--- a/src/main/java/mekanism/common/capabilities/energy/item/ItemStackEnergyHandler.java
+++ b/src/main/java/mekanism/common/capabilities/energy/item/ItemStackEnergyHandler.java
@@ -5,9 +5,11 @@ import java.util.function.Consumer;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
+import mekanism.api.Action;
 import mekanism.api.NBTConstants;
 import mekanism.api.energy.IEnergyContainer;
 import mekanism.api.energy.IMekanismStrictEnergyHandler;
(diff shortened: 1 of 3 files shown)


pupnewfster:
I see, this is a bug introduced by https://github.com/mekanism/Mekanism/commit/d9ab250bb38e5a4ddd4c5df49ce0624c9b2ea843#diff-db2aefecda17b8b1b6de1c47f2c60ae3427710660e3652f8087c7b020c34a7d9 because of the parent structure. This PR isn't quite right though as induction cells are not supposed to expose an energy cap at all. Given we haven't released a 1.19 build yet and this is a dupe bug though I will go ahead and release one final 1.18 build (even though it was basically EOL) that addresses this issue with the same fix as in https://github.com/mekanism/Mekanism/commit/6615138b3d6ed7bda1bcdfa0de9272ccbb45bdaa before actually completely EOL'ing 1.18 development.

GotoFinal:
@pupnewfster ah yes, makes sense, was wondering why its even exposed, but did not had enough time to understand how i can hide it in some nice way, so went with this approach, as its also pretty generic for any stackable items if there is more.

pupnewfster:
Yeah that is why I went ahead and changed it how I did in the linked commit (as I believe SPS ports might also have a similar issue)

pupnewfster:
10.2.5 is now live with a fix

PR #7584, opened 2022-08-18
Added ItemDecorator for Mekasuit helmet / bodyarmor

agnor99:
This PR renders an additional bar on the mekasuit helmet and bodyarmor showing the fluid/gas amount if the required module is present in the color of the fluid/chemical.
image

This also updates forge and minecraft dependencies and version ranges to a supported version of this feature

### DIFF
--- a/src/main/java/mekanism/client/ClientRegistration.java
+++ b/src/main/java/mekanism/client/ClientRegistration.java
@@ -116,7 +116,7 @@ import mekanism.client.render.entity.RenderFlame;
 import mekanism.client.render.entity.RenderRobit;
 import mekanism.client.render.hud.MekaSuitEnergyLevel;
 import mekanism.client.render.hud.MekaSuitHUD;
-import mekanism.client.render.item.ChemicalFluidBarDecorator;
+import mekanism.client.render.item.MekaSuitBarDecorator;
 import mekanism.client.render.item.block.RenderChemicalDissolutionChamberItem;
 import mekanism.client.render.item.block.RenderEnergyCubeItem;
 import mekanism.client.render.item.block.RenderFluidTankItem;
@@ -159,14 +159,12 @@ import mekanism.client.render.transmitter.RenderUniversalCable;
 import mekanism.client.sound.SoundHandler;
 import mekanism.common.Mekanism;
 import mekanism.common.base.HolidayManager;
-import mekanism.common.capabilities.Capabilities;
 import mekanism.common.content.gear.shared.ModuleColorModulationUnit;
 import mekanism.common.integration.MekanismHooks;
 import mekanism.common.item.ItemConfigurationCard;
 import mekanism.common.item.ItemCraftingFormula;
 import mekanism.common.item.ItemPortableQIODashboard;
 import mekanism.common.item.block.ItemBlockCardboardBox;
-import mekanism.common.item.gear.ItemMekaSuitArmor;
 import mekanism.common.lib.Color;
 import mekanism.common.lib.FieldReflectionHelper;
 import mekanism.common.lib.radiation.RadiationManager;
@@ -208,7 +206,6 @@ import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.level.block.entity.BlockEntity;
 import net.minecraftforge.api.distmarker.Dist;
-import net.minecraftforge.client.IItemDecorator;
 import net.minecraftforge.client.event.EntityRenderersEvent;
 import net.minecraftforge.client.event.ModelEvent.BakingCompleted;
 import net.minecraftforge.client.event.ModelEvent.RegisterAdditional;
@@ -584,12 +581,8 @@ public class ClientRegistration {
 
     @SubscribeEvent
     public static void registerItemDecorations(RegisterItemDecorationsEvent event) {
-        IItemDecorator mekaSuitDecorator = new ChemicalFluidBarDecorator(true,
-                itemStack -> itemStack.getItem() instanceof ItemMekaSuitArmor armor && //only show bar for parts that have the module installed
-                        (armor.hasModule(itemStack, MekanismModules.NUTRITIONAL_INJECTION_UNIT) || armor.hasModule(itemStack, MekanismModules.JETPACK_UNIT)),
-                Capabilities.GAS_HANDLER);
-        event.register(MekanismItems.MEKASUIT_HELMET, mekaSuitDecorator);
-        event.register(MekanismItems.MEKASUIT_BODYARMOR, mekaSuitDecorator);
+        event.register(MekanismItems.MEKASUIT_HELMET, MekaSuitBarDecorator.INSTANCE);
+        event.register(MekanismItems.MEKASUIT_BODYARMOR, MekaSuitBarDecorator.INSTANCE);
     }
 
     @SubscribeEvent
--- a/src/main/java/mekanism/client/ClientRegistration.java
+++ b/src/main/java/mekanism/client/ClientRegistration.java
@@ -116,6 +116,7 @@ import mekanism.client.render.entity.RenderFlame;
 import mekanism.client.render.entity.RenderRobit;
 import mekanism.client.render.hud.MekaSuitEnergyLevel;
 import mekanism.client.render.hud.MekaSuitHUD;
+import mekanism.client.render.item.ChemicalFluidBarDecorator;
 import mekanism.client.render.item.block.RenderChemicalDissolutionChamberItem;
 import mekanism.client.render.item.block.RenderEnergyCubeItem;
 import mekanism.client.render.item.block.RenderFluidTankItem;
@@ -158,12 +159,14 @@ import mekanism.client.render.transmitter.RenderUniversalCable;
 import mekanism.client.sound.SoundHandler;
 import mekanism.common.Mekanism;
 import mekanism.common.base.HolidayManager;
+import mekanism.common.capabilities.Capabilities;
 import mekanism.common.content.gear.shared.ModuleColorModulationUnit;
 import mekanism.common.integration.MekanismHooks;
 import mekanism.common.item.ItemConfigurationCard;
 import mekanism.common.item.ItemCraftingFormula;
 import mekanism.common.item.ItemPortableQIODashboard;
 import mekanism.common.item.block.ItemBlockCardboardBox;
+import mekanism.common.item.gear.ItemMekaSuitArmor;
 import mekanism.common.lib.Color;
 import mekanism.common.lib.FieldReflectionHelper;
 import mekanism.common.lib.radiation.RadiationManager;
@@ -212,6 +215,7 @@ import net.minecraftforge.client.event.ModelEvent.RegisterGeometryLoaders;
 import net.minecraftforge.client.event.RegisterClientReloadListenersEvent;
 import net.minecraftforge.client.event.RegisterColorHandlersEvent;
 import net.minecraftforge.client.event.RegisterGuiOverlaysEvent;
+import net.minecraftforge.client.event.RegisterItemDecorationsEvent;
 import net.minecraftforge.client.event.RegisterKeyMappingsEvent;
 import net.minecraftforge.client.event.RegisterParticleProvidersEvent;
 import net.minecraftforge.client.gui.overlay.VanillaGuiOverlay;
@@ -575,6 +579,16 @@ public class ClientRegistration {
                 ClientRegistrationUtil.registerItemColorHandler(event, (stack, index) -> index == 1 ? tint : -1, entry.getValue());
             }
         }
+    }
+
+    @SubscribeEvent
+    public static void registerItemDecorations(RegisterItemDecorationsEvent event) {
+        var mekaSuitDecorator = new ChemicalFluidBarDecorator(true,
+                itemStack -> itemStack.getItem() instanceof ItemMekaSuitArmor armor && //only show bar for parts that have the module installed
+                        (armor.hasModule(itemStack, MekanismModules.NUTRITIONAL_INJECTION_UNIT) || armor.hasModule(itemStack, MekanismModules.JETPACK_UNIT)),
+                Capabilities.GAS_HANDLER);
+        event.register(MekanismItems.MEKASUIT_HELMET, mekaSuitDecorator);
+        event.register(MekanismItems.MEKASUIT_BODYARMOR, mekaSuitDecorator);
     }
 
     @SubscribeEvent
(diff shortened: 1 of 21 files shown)


src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,116 @@
...
+        return color >> 16 & 0xFF;
+    }
+
+    private int getGreen(int color) {
+        return color >> 8 & 0xFF;
+    }
+
+    private int getBlue(int color) {
+        return color & 0xFF;
+    }
```
pupnewfster, line 115: Use helpers from `MekanismRenderer` instead

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,116 @@
...
+        return true;
+    }
+
+    private void renderBar(BufferBuilder bufferBuilder, int stackXPos, int yPos, ChemicalStack<?> stack, long capacity) {
+        int color = stack.getChemicalColorRepresentation();
+        renderBar(bufferBuilder, stackXPos, yPos, (float)((double) stack.getAmount() / capacity), color);
+    }
+
+    private void renderBar(BufferBuilder bufferBuilder, int stackXPos, int yPos, FluidStack stack, long capacity) {
+        int color = IClientFluidTypeExtensions.of(stack.getFluid()).getTintColor(stack);
```
pupnewfster, line 91: Just split `FluidUtil#getRGBDurabilityForDisplay` into two methods one that accepts an `ItemStack` like it currently does (and grabs the fluid stack), and the other that takes the fluid stack directly. That way we can keep our special case handling of lava and stuff in case we need it in the future.

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,116 @@
...
+                    renderBar(bufferBuilder, xOffset, yOffset, fluidHandler.getFluidInTank(tank), fluidHandler.getTankCapacity(tank));
+                }
+            }
+        }
+        return true;
+    }
+
+    private void renderBar(BufferBuilder bufferBuilder, int stackXPos, int yPos, ChemicalStack<?> stack, long capacity) {
+        int color = stack.getChemicalColorRepresentation();
+        renderBar(bufferBuilder, stackXPos, yPos, (float)((double) stack.getAmount() / capacity), color);
```
pupnewfster, line 87: `StorageUtils#getRatio(stack.getAmount(), capacity)`

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private void renderBar(BufferBuilder bufferBuilder, int stackXPos, int yPos, FluidStack stack, long capacity) {
+        int color = IClientFluidTypeExtensions.of(stack.getFluid()).getTintColor(stack);
+        renderBar(bufferBuilder, stackXPos, yPos, (float)((double) stack.getAmount() / capacity), color);
```
pupnewfster, line 92: `StorageUtils#getRatio(stack.getAmount(), capacity)`

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private void renderBar(BufferBuilder bufferBuilder, int stackXPos, int yPos, float width, int color) {
```
pupnewfster, line 95: Make width accept a double rather than a float especially given the `convertWidth` method already accepts a double

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -87,4 +87,6 @@ public net.minecraft.world.level.block.state.BlockBehaviour$Properties f_60886_
 public net.minecraft.world.level.chunk.ChunkGenerator f_223020_ #featuresPerStep
 public net.minecraft.world.level.levelgen.placement.PlacedFeature m_226368_(Lnet/minecraft/world/level/levelgen/placement/PlacementContext;Lnet/minecraft/util/RandomSource;Lnet/minecraft/core/BlockPos;)Z # placeWithContext
 
-public net.minecraft.world.phys.AABB m_82325_(Lnet/minecraft/world/phys/AABB;Lnet/minecraft/world/phys/Vec3;[DLnet/minecraft/core/Direction;DDD)Lnet/minecraft/core/Direction; # getDirection
\ No newline at end of file
+public net.minecraft.world.phys.AABB m_82325_(Lnet/minecraft/world/phys/AABB;Lnet/minecraft/world/phys/Vec3;[DLnet/minecraft/core/Direction;DDD)Lnet/minecraft/core/Direction; # getDirection
+
+public net.minecraft.client.renderer.entity.ItemRenderer m_115152_(Lcom/mojang/blaze3d/vertex/BufferBuilder;IIIIIIII)V # fillRect
```
pupnewfster, line 92: We sort our ATs alphabetically but honestly I feel like with how simple a method it is we may want to just copy it to `GuiUtils`, or better yet (unless it is because of the set gl states) is there a reason this can't just use `GuiUtils#fill` and pass it a `new PoseStack()` as the current pose?

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    public ChemicalFluidBarDecorator(boolean showFluid, Predicate<ItemStack> visibleFor, Capability<? extends IChemicalHandler<?,?>>... chemicalCaps) {
+        this.showFluid = showFluid;
+        this.chemicalCaps = chemicalCaps;
+        this.visibleFor = visibleFor;
+    }
+
+    @Override
+    public boolean render(Font font, ItemStack stack, int xOffset, int yOffset, float blitOffset) {
+        if (!visibleFor.test(stack))
```
pupnewfster, line 43: Missing brackets

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,116 @@
...
+            Optional<? extends IChemicalHandler<?, ?>> capabilityInstance = stack.getCapability(chemicalCap).resolve();
+            if (capabilityInstance.isPresent()) {
+                IChemicalHandler<?, ?> chemicalHandler = capabilityInstance.get();
+                if (chemicalHandler.getTanks() == 0) {
+                    continue;
+                }
+                int tank = 0;
+                if (chemicalHandler.getTanks() > 1 && Minecraft.getInstance().level != null) {
+                    //Cycle through multiple tanks every second, to save some space if multiple tanks are present
+                    tank = (int) (Minecraft.getInstance().level.getGameTime() / 20) % chemicalHandler.getTanks();
```
pupnewfster, line 61: This feels kind of weird, though I can understand at some point we would run out of space... though maybe it is worth it. Either way I think it would be worth just having a helper method like:
```java
private static int getDisplayTank(int tanks) {
    if (tanks == 0) {
        return -1;
    } else if (tanks > 1 && level != null) {
        //Cycle
        return (int) (level.getGameTime() / 20) % tanks;
    }
    return 0;
}
```
And then effectively change the code for this and fluids to be:
```java
int tank = getTank(...getTanks());
if (tank != 1) {
    renderBar....
    yOffset--;
}
```

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private void renderBar(BufferBuilder bufferBuilder, int stackXPos, int yPos, FluidStack stack, long capacity) {
+        int color = IClientFluidTypeExtensions.of(stack.getFluid()).getTintColor(stack);
+        renderBar(bufferBuilder, stackXPos, yPos, (float)((double) stack.getAmount() / capacity), color);
+    }
+
+    private void renderBar(BufferBuilder bufferBuilder, int stackXPos, int yPos, float width, int color) {
+        ItemRenderer itemRenderer = Minecraft.getInstance().getItemRenderer();
+        itemRenderer.fillRect(bufferBuilder, stackXPos + 2, yPos, 13, 1, 0, 0, 0, 255);
+        itemRenderer.fillRect(bufferBuilder, stackXPos + 2, yPos, convertWidth(width), 1, getRed(color), getGreen(color), getBlue(color), 255);
```
pupnewfster, line 98: I am guessing this is done like vanilla did it but is there a reason to not have the two draws be:
```java
int barWidth = convertWidth(width);
fill(..., stackXPos + 2, yPos, barWidth, 1, color);
int emptyWidth = 13 - barWidth;
if (emptyWidth > 0) {
    fill(..., stackXPos + 2 + barWidth, yPos, 13 - barWidth, 1, black);
}
```

And if we do end up using `GuiUtils#fill`, could also just add in validation to that which makes it NO-OP if width or height is zero so that some of the above can just be inlined for drawing the black part

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,116 @@
...
+    private final Capability<? extends IChemicalHandler<?,?>>[] chemicalCaps;
+    private final boolean showFluid;
+    private final Predicate<ItemStack> visibleFor;
+
+    /**
+     * @param showFluid if the fluidcapability should be checked for display, display above chemicalCaps if both are present
+     * @param visibleFor checks if bars should be rendered for the given itemstack
+     * @param chemicalCaps the capabilities to be displayed in order, starting from the bottom
+     */
+    public ChemicalFluidBarDecorator(boolean showFluid, Predicate<ItemStack> visibleFor, Capability<? extends IChemicalHandler<?,?>>... chemicalCaps) {
```
pupnewfster, line 35: Missing space after the `,` in the `IChemicalHandler` generic. I also believe this constructor might need a `@SafeVarargs` but I am not positive about that as I haven't pulled it locally yet to look at

src/main/java/mekanism/client/ClientRegistration.java
```diff
@@ -577,6 +581,16 @@ public static void registerItemColorHandlers(RegisterColorHandlersEvent.Item eve
         }
     }
 
+    @SubscribeEvent
+    public static void registerItemDecorations(RegisterItemDecorationsEvent event) {
+        var mekaSuitDecorator = new ChemicalFluidBarDecorator(true,
```
pupnewfster, line 586: Don't use `var` either use `ChemicalFluidBarDecorator` or `IItemDecorator`

src/main/java/mekanism/client/ClientRegistration.java
```diff
@@ -577,6 +581,16 @@ public static void registerItemColorHandlers(RegisterColorHandlersEvent.Item eve
...
+                itemStack -> itemStack.getItem() instanceof ItemMekaSuitArmor armor && //only show bar for parts that have the module installed
+                        (armor.hasModule(itemStack, MekanismModules.NUTRITIONAL_INJECTION_UNIT) || armor.hasModule(itemStack, MekanismModules.JETPACK_UNIT)),
```
pupnewfster, line 588: Rather than `hasModule` we may want to consider doing `isModuleEnabled`, though I need to think about what makes more sense (for now you can probably leave this as hasModule)

pupnewfster (changes requested):

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
agnor99: I've seen them but deemed them "unusuable", because they return a float divided by 255. I would have to multiply that value back and it's not worth it imo

src/main/java/mekanism/client/ClientRegistration.java
agnor99: it's the check that the fluid/gas handler use so I just used them for now

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+        RenderSystem.disableDepthTest();
+        yOffset = yOffset + 12;
+        for (Capability<? extends IChemicalHandler<?,?>> chemicalCap : chemicalCaps) {
+            Optional<? extends IChemicalHandler<?, ?>> capabilityInstance = stack.getCapability(chemicalCap).resolve();
+            if (capabilityInstance.isPresent()) {
+                IChemicalHandler<?, ?> chemicalHandler = capabilityInstance.get();
+                if (chemicalHandler.getTanks() == 0) {
+                    continue;
+                }
```
pupnewfster, line 54: This isn't needed anymore because it is handled by the `tanks != -1` check below

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,107 @@
...
+                    yOffset--;
+                }
+            }
+        }
+
+        if (showFluid) {
+            Optional<IFluidHandlerItem> capabilityInstance = FluidUtil.getFluidHandler(stack).resolve();
+            if (capabilityInstance.isPresent()) {
+                IFluidHandlerItem fluidHandler = capabilityInstance.get();
+                if (fluidHandler.getTanks() != 0) {
```
pupnewfster, line 67: This isn't needed anymore because it is handled by the `tanks != -1` check below

src/main/java/mekanism/client/render/item/ChemicalFluidBarDecorator.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private void renderBarChemical( int stackXPos, int yPos, ChemicalStack<?> stack, long capacity) {
+        int color = stack.getChemicalColorRepresentation();
+        renderBar(stackXPos, yPos, StorageUtils.getRatio(stack.getAmount(), capacity), color);
+    }
+
+    private void renderBarFluid(int stackXPos, int yPos, FluidStack stack, long capacity) {
+        int color = FluidUtils.getRGBDurabilityForDisplay(stack).orElse(0xFFFFFFFF);
+        renderBar(stackXPos, yPos, StorageUtils.getRatio(stack.getAmount(), capacity), color);
```
pupnewfster, line 85: Given this is the same for renderBarChemical and this method, it might be worth simplifying the method they both call some to have the signature:
```java
private void renderBar(int stackXPos, int yPos, long amount, long capacity, int color)
```
That way they don't need to duplicate the calls to `getRatio`

PR #7628, opened 2022-10-02
Added 'Cosmetic Armor Reworked' integration

gisellevonbingen:

I want to see my skin, even while meka-suit working.
Cosmetic Armor Reworked mod did what I wanted.

Except for just one thing that worries me.
This pull request is replace itemstack for render hand on 1st person view.
Match with same as seen in 3rd person view.

CosmeticArmor

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -404,6 +404,8 @@ dependencies {
 
     implementation fg.deobf("com.blamejared.crafttweaker:CraftTweaker-forge-${minecraft_version}:${crafttweaker_version}")
     implementation fg.deobf("com.blamejared.jeitweaker:JEITweaker-${minecraft_version}:${jeitweaker_version}")
+    
+    implementation fg.deobf("curse.maven:cosmetic-armor-reworked-237307:${cosmetic_armror_id}")
 
     //Mods we have dependencies on but don't bother loading into the dev environment
     compileOnly fg.deobf("curse.maven:projecte-api-226410:${projecte_api_id}")
--- a/gradle.properties
+++ b/gradle.properties
@@ -41,6 +41,7 @@ oc2_id=3739469
 projecte_api_id=3722832
 top_version=1.18-5.1.0-8
 wildfire_gender_mod_id=3706808
+cosmetic_armror_id=3738144
 
 #Mod dependency min version ranges
 crafttweaker_version_range=[9.1.139,)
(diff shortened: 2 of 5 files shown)


pupnewfster:
While I am not fully against this idea, I do somewhat feel like this might be better in the cosmetic armor mod itself so that every single mod that renders armor via that event doesn't have to add specific support for it? So then in cosmetic armor:
- adjust stack in a listener with priority high or highest
- other mods then render in normal
- return stack to normal in a listener with priority low or lowest

Which I believe is semi similar to how it does things (all be it with two different events) for getting things to show/hide appropriately when rendering the player.

gisellevonbingen:
Is it meaning adjust chest equipment item stack temporary during posting RenderArmEvent?
um.. I don't know that is right way for render.

I guess 'Add registration about render hand' is better instead of adjust itemstack.
Curios mod is very well implement this way.

Like this:  Simply Jetpacks 2 mod's snippet
https://github.com/Tomson124/SimplyJetpacks2/blob/1.18/src/main/java/stormedpanda/simplyjetpacks/integration/CuriosIntegration.java#L33
```java
CuriosRendererRegistry.register(RegistryHandler.JETPACK_CREATIVE.get(), () -> new JetpackRenderer(getJetpackTexture("creative")));
```
JetpackRenderer is not called automatically when curios slot set as hide.

So... If Cosmetic Armor Reworked added registration for render hand,
Is it ok to change like below?

Add return statement on head of event listener at RenderTickHandler:
```java
--▲--
@SubscribeEvent
public void renderArm(RenderArmEvent event) {
    AbstractClientPlayer player = event.getPlayer();
    ItemStack chestStack = player.getItemBySlot(EquipmentSlot.CHEST);
        
+   if (Mekanism.hooks.CosmeticArmorLoaded) {
+       return;
+   }
--▼--
```

pupnewfster:
I mean something like this: <https://github.com/zlainsama/CosmeticArmorReworked/blob/master/src/main/java/lain/mods/cos/impl/client/PlayerRenderHandler.java#L76-L103> but for the `RenderArmEvent` which I believe is newer than their code for the `RenderHandEvent`. And then I believe Mekanism (and any other mod that changes rendering to make the armor render via the `RenderArmEvent` will just work out of the box with cosmetic armor reworked.

gisellevonbingen:
Ah, i understand now.
Thank for exactly explain.

I will be back, if any question or problem
Thanks!

PR #7710, opened 2023-01-20
Optimize all .png files

Jatie1:
All .png files losslessly optimized with [File Optimizer](https://sourceforge.net/projects/nikkhokkho/files/FileOptimizer/).

All images should be functionally identical due to the lossless compression programs used.


pupnewfster:
Thank you for your contribution but we already optimize the majority of our png's using a set of arguments that we know to not change the quality: https://github.com/mekanism/Mekanism/blob/1.19.x/build.gradle#L601-L629,  which we run after adding new assets

PR #7748, opened 2023-02-28
Rewrite how transporters transfer items along their calculated paths

pupnewfster:
Draft PR for now, description will be updated as things are implemented and figured out better, but general idea of this will be to:
- Adds Game Tests to validate transfer behavior of Logistical Transporters and ensure it remains consistent as things are rewritten
- Slightly improves memory churn of path caching by making cached paths immutable and using the cached list directly in the destination logic rather than copying it
- Make item movement along paths happen at a network level instead of having each transporter have to keep track of the stacks that are currently in them and forward them onto the next transporter in the path. This should greatly improve server client networking, as well as hopefully fix some bugs and provide a general improvement to performance by having path invalidation and recalculation happen as needed rather than eagerly.


Gamercave:
has this been merged into a specific 1.19.2 download we are currently facing a consistent crash issue on a server and I am hoping their may be a version where this is a quick solution

PR #7751, opened 2023-03-04
Ensure Mekanism modules are added in a thread-safe manner

andriihorpenko:

Ensures Mekanism modules are added in a thread-safe manner.

## Background

After Mekanism itself is loaded, modules are loading in parallel. In some occasions, modules access `Mekanism#loadedModules` concurrently in bad timings, which leads to an inevitable crash as plain `ArrayList` is not thread-safe. In my case, this triggers `IndexOutOfBoundException`.

## Inspiration

Approach used in this PR is identical to the one used in 1.19.x. Newer versions of Mekanism introduced a synchronized `addModule` method, which solves the given issue.

https://github.com/mekanism/Mekanism/blob/ee5707f96155c0306f99b768212cb893d43b60e8/src/main/java/mekanism/common/Mekanism.java#L253-L255

### DIFF
--- a/src/additions/java/mekanism/additions/common/MekanismAdditions.java
+++ b/src/additions/java/mekanism/additions/common/MekanismAdditions.java
@@ -51,7 +51,7 @@ public class MekanismAdditions implements IModModule {
     public static VoiceServerManager voiceManager;
 
     public MekanismAdditions() {
-        Mekanism.modulesLoaded.add(instance = this);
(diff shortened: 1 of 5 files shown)


pupnewfster:
I will merge this just so that it is fixed on the branch, but I wouldn't necessarily count on a release fixing this any time soon (moving this week and also we aren't currently developing or really releasing updates for 1.18 anymore)

PR #7770, opened 2023-04-11
[1.12] Optimize CableUtils#emit()

DaMatrix:
*Not sure if you're still accepting PRs for older MC versions, if not just close this :)*

Previously, `CableUtils#emit(IEnergyWrapper)` would always create a new `HashSet` containing a single `EnergyAcceptorTarget` (to be passed to `EmitUtils#sendToAcceptors()`). Iterating over that isn't particularly fast, and creating it results in allocating a total of 4 objects totalling ~288 bytes (assuming my math is correct, could be slightly off).

This PR replaces it with a single-element `Set` created by `Collections#singleton(Object)` - this makes the implementation of `Set#forEach()` much simpler (and easily inlineable!), and reduces the GC overhead to a single allocated object @ 32 bytes.

Based on about an hour of profiling before and after, this reduced the overall ms/tick on my server by a respectable ~6%.

(also fixed up some dependency references, as a few of the referenced versions and/or repositories are no longer available)


a-droid-boop:
Mek devs have stopped dev support for older versions, but this is still a welcome addition regardless

PR #7783, opened 2023-04-30
Fix Nutritional Paste/Canteen giving too much food saturation

Lykrast:
With Appleskin to show food saturation, I have noticed that the canteen in mek 10.3.8 fills saturation meter wayyyy more than it sounds like it should.

After searching the code I realized the paste's intended saturation ratio is multiplied by the food restored and then given to the `eat` function, but that function expects a ratio and multiplies it by the food amount (this gets done in `FoodData.eat(int, float)`). So nutritional paste is wayyyy more nutritrious than it should.

This simple change should fix that.

*(Please squash the commits, I thought making stuff on github directly would be easier but it just made one commit per file)*

### DIFF
--- a/src/main/java/mekanism/common/content/gear/mekasuit/ModuleNutritionalInjectionUnit.java
+++ b/src/main/java/mekanism/common/content/gear/mekasuit/ModuleNutritionalInjectionUnit.java
@@ -41,7 +41,7 @@ public class ModuleNutritionalInjectionUnit implements ICustomModule<ModuleNutri
                 module.useEnergy(player, usage.multiply(toFeed));
                 FluidUtil.getFluidHandler(container).ifPresent(handler ->
                       handler.drain(MekanismFluids.NUTRITIONAL_PASTE.getFluidStack(toFeed * MekanismConfig.general.nutritionalPasteMBPerFood.get()), FluidAction.EXECUTE));
-                player.getFoodData().eat(needed, needed * MekanismConfig.general.nutritionalPasteSaturation.get());
+                player.getFoodData().eat(needed, MekanismConfig.general.nutritionalPasteSaturation.get());
             }
         }
     }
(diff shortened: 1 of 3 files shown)


pupnewfster:
Fairly certain the multiplying by amount actually consumed is intended. Given the way it works is if you then were eating it constantly (so only using base amount) it would provide that saturation level so overall it is an equivalent amount of saturation it provides when you end up filling lots of hunger shanks instead of being basically worthless as a food source

Lykrast:
No I'm pretty sure you're *not* supposed to multiply it before giving it to `player.getFoodData().eat`.

Going back from the call hierarchy in the code (forge 1.19.2):

* `Item.finishUsingItem` calls `LivingEntity.eat` when it's something edible.
* `Player.eat` (which overrides `LivingEntity.eat`) calls `FoodData.eat(Item, ItemStack, LivingEntity)` with the item.
* `FoodData.eat(Item, ItemStack, LivingEntity)` does `this.eat(foodproperties.getNutrition(), foodproperties.getSaturationModifier());`: no multiplying food by saturation done here.

Then looking at `FoodData.eat(int nutrition, float saturation)` (with renamed arguments)
```java
this.foodLevel = Math.min(nutrition + this.foodLevel, 20);
this.saturationLevel = Math.min(this.saturationLevel + (float)nutrition * saturation * 2.0F, (float)this.foodLevel);
```
Which is where the multiplication is done, so it shouldn't be done before the `eat` call.

And then looking at vanilla foods (in `Foods`) we see that saturation is 0.8 for all cooked meats, just like default paste saturation. So eating 6 food units should restore as much saturation as cooked mutton (9.6), and 8 food units as much as a steak (12.8). But because paste multiplies it before the `eat` call, eating 6 food units of paste gives 57.6 units of saturation, instantly filling the saturation bar, and 8 food units 102.4, wayyyyy more than expected.

pupnewfster:
Thanks and yeah I looked back all the way to 1.16 and that is how vanilla handled it back then as well so I am not quite sure why I was under the impression that it didn't and that we had to multiply manually.

PR #7812, opened 2023-07-04
Add Radial Menu for Blasting Unit

GaviTSRA:
- Add a radial menu to the blasting unit for the meka tool, similar to the excavation and vein mining units
Without this option the blasting unit feels useless as you have to constantly change the setting, leading to it not being used.

I am very new to forge modding, any comments on the code are welcome, even though its nearly all the same as for the excavation module.
I know the icons are not good and will try to update them in a few hours, i am not really good with something like that though. If anyone would be able to create good icons for this menu it would mean a lot!


pupnewfster (approved):
Looks great, main comment is that for consistency with our other radials the icons probably shouldn't have been colored but I am going to go ahead and merge this regardless as you said you are open to help for the textures and cyanide is up for creating textures for this so these textures should be fine for now as temporary ones.

PR #7818, opened 2023-07-12
Rework how multiblock caches are tracked and persisted

pupnewfster:
- Expands the data corruption temp file backup we have for saved data to be used for all our saved data implementations instead of just the QIO item lookup
- Removed the old fallback loading/checks relating to the old UUID format as it was either transitioned over in 1.16 or 1.18 (I forget which) and at this point doesn't serve much purpose
- Reworks how multiblock caches are managed and persisted to be based on a per manager world save data rather than saving multiple bits of copies to individual tiles. Some additional notes about the implementation and what this means:
  - Any existing multiblock contents will be voided/lost, as this change is being made during a breaking change window of changing MC versions this isn't a big deal but is worth mentioning and documenting.
  - If the control block (be it a reactor controller or the randomly chosen block) gets destroyed the contents of the multiblock will be consistently saved instead of the previous case of voided in the majority of situations.
  - When multiblocks reform there is a bit of disk churn as they get new ids and thus the save data becomes dirty and has to save again. This is a side effect of the route taken to avoid tracking what positions are for each cache in a way that doesn't lead to potential dupe issues or multiple now separated multiblocks referencing the same cache. Previously this was a potential issue as well (as noted in the now removed `MultiblockManager#pullInventory` method) though was only limited to portions that were potentially unloaded or inside of things like cardboard boxes. One solution to this would be to track all the positions in the multiblock and then if any are missing when reforming use a new id, but the extra overhead for memory and validating each position does not seem particularly worth it.
  - A lot of the extra cycles relating to trying to keep the `CacheWrapper` up to date with all the different multiblock tiles and having them all have their caches pointing at the same same instance, as well as tracking their current positions is no longer necessary as we remove caches when merging them so if something has a stale id that no longer points at a cache we can just ignore it instead of needing to make sure we clear every reference to the id.

### DIFF
--- a/src/main/java/mekanism/common/lib/MekanismSavedData.java
+++ b/src/main/java/mekanism/common/lib/MekanismSavedData.java
@@ -41,13 +41,10 @@ public abstract class MekanismSavedData extends SavedData {
      * Note: This should only be called from the server side
      */
     public static <DATA extends MekanismSavedData> DATA createSavedData(DimensionDataStorage dataStorage, Supplier<DATA> createFunction, String name) {
-        //TODO: Should we always prepend the modid to the name like we are doing for the multiblock managers?
-        // In general our names are probably unique enough but do we want to ensure they should be truly unique?
-        // This is especially prevalent for things like the radiation_manager
         return dataStorage.computeIfAbsent(tag -> {
             DATA handler = createFunction.get();
             handler.load(tag);
             return handler;
-        }, createFunction, name);
+        }, createFunction, Mekanism.MODID + "_" + name);
     }
 }
--- a/src/main/java/mekanism/common/lib/MekanismSavedData.java
+++ b/src/main/java/mekanism/common/lib/MekanismSavedData.java
@@ -1,0 +1,53 @@
+package mekanism.common.lib;
+
+import java.io.File;
+import java.util.function.Supplier;
+import mekanism.common.Mekanism;
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.world.level.saveddata.SavedData;
+import net.minecraft.world.level.storage.DimensionDataStorage;
+import net.minecraftforge.server.ServerLifecycleHooks;
+import org.jetbrains.annotations.NotNull;
+
+public abstract class MekanismSavedData extends SavedData {
+
+    public abstract void load(@NotNull CompoundTag nbt);
+
+    @Override
+    public void save(@NotNull File file) {
+        if (isDirty()) {
+            //This is loosely based on Refined Storage's RSSavedData's system of saving first to a temp file
+            // to reduce the odds of corruption if the user's computer crashes while the file is being written
+            File tempFile = file.toPath().getParent().resolve(file.getName() + ".tmp").toFile();
+            super.save(tempFile);
+            if (file.exists() && !file.delete()) {
+                Mekanism.logger.error("Failed to delete " + file.getName());
+            }
+            if (!tempFile.renameTo(file)) {
+                Mekanism.logger.error("Failed to rename " + tempFile.getName());
+            }
+        }
+    }
+
+    /**
+     * Note: This should only be called from the server side
+     */
+    public static <DATA extends MekanismSavedData> DATA createSavedData(Supplier<DATA> createFunction, String name) {
+        DimensionDataStorage dataStorage = ServerLifecycleHooks.getCurrentServer().overworld().getDataStorage();
+        return createSavedData(dataStorage, createFunction, name);
+    }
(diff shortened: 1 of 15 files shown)


src/main/java/mekanism/common/lib/multiblock/FormationProtocol.java
```diff
@@ -127,14 +131,21 @@ public FormationResult doUpdate() {
...
+            boolean trackCache = cache == null;
+            if (trackCache) {
                 cache = manager.createCache();
             }
 
             cache.apply(structureFound);
             structureFound.inventoryID = idToUse;
             structureFound.onCreated(world);
+            if (trackCache) {
+                //If it is a new fresh cache we need to make sure to then sync the multiblock back to the cache
```
thiakil, line 143: I don't quite understand why this wouldn't need to happen with a reused/merged cache?

src/main/java/mekanism/common/tile/prefab/TileEntityMultiblock.java
```diff
@@ -232,9 +218,6 @@ public void setRemoved() {
         super.setRemoved();
         if (!isRemote()) {
             structure.invalidate(level);
-            if (cachedID != null) {
```
thiakil, line 235: is setRemoved handled elsewhere?

pupnewfster: It doesn't need to be because of things getting new ids that if they point to a cache that ends up not existing at some point then we can just treat them as not having a cache.

src/main/java/mekanism/common/lib/multiblock/FormationProtocol.java
pupnewfster: The issue is that a fresh cache has no slots or tanks etc listed for how many slots it has, this basically forces that syncing to happen whereas otherwise it is already up to date from the cache actually having stuff to apply a few lines above in `cache.apply(structureFound)`. I just left that line for new caches as well in case there are things that are set for defaults in regards to say injection rate that has to be synced from the cache to the multiblock even if the rest of the cache hasn't been initialized yet.

src/main/java/mekanism/common/tile/prefab/TileEntityMultiblock.java
thiakil: Perhaps this is my ignorance, when is this method called? (I've not had a chance to open this in an ide)

pupnewfster: `setRemoved` is called when the tile is removed (broken) or the chunk it is in unloads. As the MultiblockManager no longer keeps track of a id -> cache + positions map and only keeps track of an id -> cache map there is no need to be removing the position from the manager (which is what the invalidate method was doing)

thiakil: OK, my assumption was correct there.

So it doesn't need to mark the multiblock as invalid? (because part of it was removed/unloaded)

pupnewfster: That is what the `structure.invalidate(level)` line is doing. The `MultiblockManager#invalidate` was purely for invalidating it in regards to the cache.

thiakil: Cool, thanks for explaining!

src/main/java/mekanism/common/lib/MekanismSavedData.java
```diff
@@ -0,0 +1,53 @@
...
+    public static <DATA extends MekanismSavedData> DATA createSavedData(Supplier<DATA> createFunction, String name) {
+        DimensionDataStorage dataStorage = ServerLifecycleHooks.getCurrentServer().overworld().getDataStorage();
+        return createSavedData(dataStorage, createFunction, name);
+    }
+
+    /**
+     * Note: This should only be called from the server side
+     */
+    public static <DATA extends MekanismSavedData> DATA createSavedData(DimensionDataStorage dataStorage, Supplier<DATA> createFunction, String name) {
+        //TODO: Should we always prepend the modid to the name like we are doing for the multiblock managers?
```
thiakil: yes, we probably should

thiakil:
While it won't have a huge benefit, can we make `mekanism.common.lib.multiblock.MultiblockCache#save` not call `DataHandlerUtils.writeContainers` if it doesn't have any containers?

thiakil (approved):
LGTM

PR #7821, opened 2023-07-15
Change language file in Mekanism: Generator

39Choko:
Small update fr_fr.json in Mekanism: Generator


AlexHgC:
That's not how to translate Mekanism, you need to use Crowdin for that. Some of your change are wrong, like "Prod max d'Eau", doesn't need a E, and "Refroidis par l'eau", "Refroidis par eau" is the right thing too, like "Moteur refroidi par eau"

PR #7824, opened 2023-07-20
Combine biofuel recipes for various leaves, saplings and seeds

TurnrDev:
Make use of tags to ensure compatibility with leaves, saplings and seeds from mods and future updates, and to combine the recipes into one, making JEI a look cleaner.


TurnrDev:
I've proposed this for 1.18.x as that's the version I use, but there's no reason this can't be ported to other versions

pupnewfster:
Firstly we are no longer developing for 1.18 so if you would like to continue this it should be targeted at the 1.20 branch. Secondly manually editing generated json files will just cause them to get reverted the next time the data is generated, so should be done by somehow tweaking the logic behind https://github.com/mekanism/Mekanism/blob/8554cc4d6089346fdbb648743f1561105e952d0c/src/datagen/main/java/mekanism/common/recipe/impl/CrusherRecipeProvider.java#L421

TurnrDev:
I should have known by the fact they said "generated" - so does that mean 1.18.2 Mekanism won't be getting any more updates? Gutting but understandable! 

I'll adjust my PR to suit when I get time to familiarise myself with the mod's code

TurnrDev:
Just commenting to say that this PR is *not* abandoned, I've just been spending my spare time (weekends) learning mod development so I can understand what changes I'm making when I make it 😄

TurnrDev:
I think it's safe to say I probably won't get around to this, too lazy

PR #7825, opened 2023-07-20
Add forge:bioethanol to forge:biofuel tag

TurnrDev:
I've seen some mods use forge:biofuel and some use forge:bioethanol, since bioethanol is more specific than biofuel, I've included that tag in the biofuel tag.


pupnewfster:
We are no longer developing for 1.18 so please target the 1.20 branch, and also tag adjustments should be done via datagen and not manual file modification.

TurnrDev:
Just commenting to say that this PR is not abandoned, I've just been spending my spare time (weekends) learning mod development so I can understand what changes I'm making when I make it 😄

TurnrDev:
I think it's safe to say I probably won't get around to this, too lazy

PR #7850, opened 2023-08-20
Update MekanismRecipeType.java

igentuman:
Addon devs can't register without reflection


pupnewfster:
This is an internal class that we just use as a partial helper to make it easier for us to instantiate and register new `RecipeType`s. First off:
- Addons should not be registering using it as it adds things to mekanism's modid and not whatever mod you have...
- You can just register things directly to the RecipeType registry...

PR #7857, opened 2023-09-03
Computer methods help

thiakil:

- Add `help()` and `help(methodName)` methods to all bound method holders
    - Exposes params, return types, descriptions, and enum values
- Adds documentation for methods to the annotations
- Adds datagens for CSV and JSON formats for the help data
- Adds capability to use `Either<A,B>` for computer method returns (easier to document)
- Adds `TileEntityGenerator.getProductionRate` and moves mismatching methods with same purpose to match (computer names matched)
- Moves the factories to use a Builder for `MethodData` to clean up default values

Not all methods have been documented yet


annotation-processor/src/main/java/mekanism/visitors/AnnotationHelper.java
```diff
@@ -36,12 +37,36 @@ public AnnotationHelper(Elements elementUtils, AnnotationMirror annotationMirror
      */
     public Object getLiteral(String key, Object defaultValue) {
         ExecutableElement element = nameToElement.get(key);
+        if (element == null) {
```
pupnewfster, line 40: Why does this and `getEnumConstantName` null check the element whereas other methods seem to null check the value returned by the value map?

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+import net.minecraft.util.ExtraCodecs;
+import net.minecraftforge.common.data.ExistingFileHelper;
+import org.apache.commons.lang3.tuple.Pair;
+import org.jetbrains.annotations.NotNull;
+
+@MethodsReturnNonnullByDefault
+@ParametersAreNotNullByDefault
+public class ComputerHelpProvider implements DataProvider {
+    private static final String[] METHOD_CSV_HEADERS = new String[]{"Class", "Method Name", "Params", "Returns", "Restriction", "Requires Public Security", "Description"};
+    private static final String[] ENUM_CSV_HEADERS = new String[]{"Type Name", "Values"};
```
pupnewfster: These two lines don't need the explicit `new String[]` declaration.

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+    private static String csvReturnsValue(Returns returns) {
+        if (returns == Returns.NOTHING) {
+            return "";
+        }
+        return returns.type();
+    }
+
+    @Override
+    public String getName() {
+        return "ComputerHelp:" + modid;
```
pupnewfster: We should put a space after the `:` so that the name of the provider is for example `ComputerHelp: mekanism`

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    //implementation isn't quite "stable" like json, but input is always sorted
```
pupnewfster: What does it mean by not quite stable? Is the order sometimes different from one run to the next? If so that is a problem.

src/main/java/mekanism/common/integration/computer/MethodData.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public static class Builder<T> {
+
+        private final String methodName;
+        private MethodRestriction restriction = MethodRestriction.NONE;
+        private String[] requiredMods = NO_STRINGS;
+        private boolean threadSafe = false;
+        private String[] argumentNames = NO_STRINGS;
+        private Class<?>[] argClasses = NO_CLASSES;
+        private Class<?> returnType = void.class;
```
pupnewfster, line 44: In general in places I have been using `Void.TYPE` rather than `void.class`, so it is probably good to be consistent here with that.

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
```diff
@@ -0,0 +1,170 @@
...
+        public static final Returns NOTHING = new Returns("Nothing", void.class, NO_CLASSES, null);
+        public static final Codec<Returns> CODEC = RecordCodecBuilder.create(instance->instance.group(
+              Codec.STRING.fieldOf("type").forGetter(Returns::type),
+              CLASS_TO_STRING_CODEC.fieldOf("javaType").forGetter(Returns::javaType),
+              CLASS_TO_STRING_CODEC.listOf().optionalFieldOf("javaExtra", Collections.emptyList()).<Class<?>[]>xmap(cl -> cl.toArray(new Class[0]), Arrays::asList).forGetter(Returns::javaExtra)/*,
+              Codec.STRING.listOf().optionalFieldOf("values", null).forGetter(Returns::values)*/
+        ).apply(instance, Returns::new));
+
+        public static Returns from(MethodData<?> data) {
+            if (data.returnType() != void.class) {
```
pupnewfster, line 151: In general in places I have been using `Void.TYPE` rather than `void.class`, so it is probably good to be consistent here with that.

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
```diff
@@ -0,0 +1,170 @@
...
+            return new Param(paramName, getHumanType(argClass), argClass, getEnumConstantNames(argClass));
+        }
+    }
+
+    public record Returns(String type, Class<?> javaType, Class<?>[] javaExtra, @Nullable List<String> values){
+        public Returns(String type, Class<?> javaType, Class<?>[] javaExtra) {
+            this(type, javaType, javaExtra, null);
+        }
+
+        public static final Returns NOTHING = new Returns("Nothing", void.class, NO_CLASSES, null);
```
pupnewfster, line 142: In general in places I have been using `Void.TYPE` rather than `void.class`, so it is probably good to be consistent here with that.

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
```diff
@@ -0,0 +1,170 @@
...
+        return from(data.method());
+    }
+
+    public static MethodHelpData from(MethodData<?> data) {
+        List<Param> params = new ArrayList<>();
+        for (int i = 0; i < data.argumentNames().length; i++) {
+            params.add(Param.from(data.argClasses()[i], data.argumentNames()[i]));
+        }
+
+        return new MethodHelpData(data.name(), params.isEmpty() ? null : params, Returns.from(data), data.methodDescription(), data.restriction(), data.requiresPublicSecurity());
```
pupnewfster: If the goal of this is so that if the parameters are empty then it should always use null instead of say an empty list we might as well move the check to the record's compact constructor:
```java
public MethodHelpData {
    if (params != null && params.isEmpty()) {
        params = null;
    }
}
```

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
```diff
@@ -0,0 +1,170 @@
...
+              Codec.STRING.fieldOf("type").forGetter(Returns::type),
+              CLASS_TO_STRING_CODEC.fieldOf("javaType").forGetter(Returns::javaType),
+              CLASS_TO_STRING_CODEC.listOf().optionalFieldOf("javaExtra", Collections.emptyList()).<Class<?>[]>xmap(cl -> cl.toArray(new Class[0]), Arrays::asList).forGetter(Returns::javaExtra)/*,
+              Codec.STRING.listOf().optionalFieldOf("values", null).forGetter(Returns::values)*/
+        ).apply(instance, Returns::new));
+
+        public static Returns from(MethodData<?> data) {
+            if (data.returnType() != void.class) {
+                List<String> enumConstantNames = getEnumConstantNames(data.returnType());
+                if (enumConstantNames == null && data.returnExtra().length > 0) {
```
pupnewfster: Correct me if I am wrong but I believe all removing this if statement would do would be to make `i` be initialized to `zero`? (Ignoring the fact the check is in the opposite direction in the for loop condition as in the if statement) At which point is it worth this extra check/line?

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
```diff
@@ -0,0 +1,170 @@
...
+                    for (int i = 0; i < data.returnExtra().length && enumConstantNames == null; i++) {
+                        enumConstantNames = getEnumConstantNames(data.returnExtra()[0]);
```
pupnewfster: I am guessing this should be using `i` as the index retrieved and not always just use `0`?

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
```diff
@@ -0,0 +1,170 @@
...
+            }
+            return humanType;
+        }
+        return clazz.getSimpleName();
+    }
+
+    @SuppressWarnings("unchecked")
+    @Nullable
+    public static List<String> getEnumConstantNames(Class<?> argClass) {
+        if (!Enum.class.isAssignableFrom(argClass)) {
```
pupnewfster: Is there a reason this checks the enum class like this as opposed to using `argClass.isEnum()`? Is it because of anonymous classes and not wanting to deal with any potential cases of `getDeclaringClass`?

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    @NotNull
+    private static String getHumanType(Class<?> clazz) {
+        return getHumanType(clazz, NO_CLASSES);
+    }
+
+    @NotNull
+    private static String getHumanType(Class<?> clazz, Class<?>[] extraTypes) {
+        if (clazz == UUID.class || clazz == ResourceLocation.class || clazz == Item.class || Enum.class.isAssignableFrom(clazz)) {
```
pupnewfster: Is there a reason this checks the enum class like this as opposed to using `clazz.isEnum()`? Is it because of anonymous classes and not wanting to deal with any potential cases of `getDeclaringClass`?

Also given we have items here do we also want the primitive Fluid and Chemical?

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+            }
+        });
+    }
+
+    @NotNull
+    private static Map<Class<?>, List<String>> getEnumValues(Map<Class<?>, List<MethodHelpData>> helpData) {
+        Map<Class<?>, List<String>> enumToValues = new TreeMap<>(Comparator.comparing(Class::getSimpleName));
+        helpData.forEach((unused, methods)->{
+            for (MethodHelpData method : methods) {
+                if (Enum.class.isAssignableFrom(method.returns().javaType())) {
```
pupnewfster: Is there a reason this checks the enum class like this as opposed to using `method.returns().javaType().isEnum()`? Is it because of anonymous classes and not wanting to deal with any potential cases of `getDeclaringClass`?

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+                    Class<?> jType = method.returns().javaType();
+                    enumToValues.put(jType, method.returns().values());
+                }
+                for (Class<?> extraClass : method.returns().javaExtra()) {
+                    if (Enum.class.isAssignableFrom(extraClass)) {
```
pupnewfster: Is there a reason this checks the enum class like this as opposed to using `extraClass.isEnum()`? Is it because of anonymous classes and not wanting to deal with any potential cases of `getDeclaringClass`?

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
```diff
@@ -0,0 +1,170 @@
...
+            return "String ("+clazz.getSimpleName()+")";
+        }
+        if (Frequency.class.isAssignableFrom(clazz) || clazz == Coord4D.class || Vec3i.class.isAssignableFrom(clazz) || clazz == FluidStack.class || clazz == ItemStack.class || clazz == BlockState.class || ChemicalStack.class.isAssignableFrom(clazz) || IFilter.class.isAssignableFrom(clazz)) {
+            return "Table ("+clazz.getSimpleName()+")";
+        }
+        if (clazz == int.class || clazz == long.class || clazz == float.class || clazz == double.class || clazz == FloatingLong.class || Number.class.isAssignableFrom(clazz)) {
```
pupnewfster, line 67: In general in places I have been using `Integer.TYPE`, `Long.TYPE` etc rather than `primitive.class`, so it is probably good to be consistent here with that. Also is there a reason we are not checking bytes or shorts?

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
```diff
@@ -0,0 +1,170 @@
...
+            if (ClassUtils.isPrimitiveWrapper(clazz)) {
```
pupnewfster, line 68: Just to verify/leave a note so one of us remembers to check... Is `ClassUtils` (or more specifically apache lang3) present on the server side? Or is it a dependency that is just present on the client? I feel like I remember that library or some similarly named one only being on one side.

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
```diff
@@ -0,0 +1,170 @@
...
+            String humanType = "List";
+            if (extraTypes.length > 0) {
+                humanType += " (" + getHumanType(extraTypes[0]) + ")";
+            }
+            return humanType;
+        }
+        if (clazz == Convertable.class || clazz == Either.class) {
+            String humanType = "Varies";
+            if (extraTypes.length > 0) {
+                humanType = Arrays.stream(extraTypes).map(MethodHelpData::getHumanType).collect(Collectors.joining(" or "));
```
pupnewfster: We can just return the result of this line rather than overwriting the variable, and then just inline the `Varies` return to if `extraTypes.length` is zero given the value the string is initialized to is unused if there are any `extraTypes` (unless the `=` is meant to be `+=`).

src/main/java/mekanism/common/integration/computer/annotation/SyntheticComputerMethod.java
```diff
@@ -39,4 +39,9 @@
      * Whether the synthetic setter is thread-safe or needs to be queued to run on the main thread.
      */
     boolean threadSafeSetter() default false;
-}
\ No newline at end of file
+
+    /**
+     * Method description for documentation.
+     */
+    String methodDescription() default "";
```
pupnewfster: Is having a singular field for method description appropriate here or should it be split into a description for the getter and setter?

src/main/java/mekanism/common/integration/computer/annotation/ComputerMethod.java
```diff
@@ -29,4 +29,19 @@
...
+     */
+    boolean requiresPublicSecurity() default false;
+
+    /**
+     * Method description for documentation.
+     */
+    String methodDescription() default "";
+
+    /**
+     * For {@link mekanism.common.integration.computer.Convertable} return types, specifies the possible values for documentation
```
pupnewfster, line 44: Do we also want to link to `Either`? Or is this strictly only used for `Convertable`?

src/main/java/mekanism/common/integration/computer/opencomputers2/MekanismDevice.java
```diff
@@ -161,6 +161,11 @@ private RPCParameter[] buildOCParams() {
         public Optional<RPCMethod> findOverload(RPCInvocation invocation) {
             return Optional.ofNullable(invocation.getParameters().size() == methodData.argumentNames().length ? this : null);
         }
+
+        @Override
+        public Optional<String> getDescription() {
```
pupnewfster, line 166: Do we want to also track descriptions for parameters given OC2 supports that? Personally I am not sure how useful it would be (especially as OC2 isn't updated to 1.20), but wanted to bring it up.

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+    private static final String[] METHOD_CSV_HEADERS = new String[]{"Class", "Method Name", "Params", "Returns", "Restriction", "Requires Public Security", "Description"};
+    private static final String[] ENUM_CSV_HEADERS = new String[]{"Type Name", "Values"};
+    private final PackOutput.PathProvider pathProvider;
+    private final ExistingFileHelper existingFileHelper;
+    private final String modid;
+
+    public ComputerHelpProvider(PackOutput output, ExistingFileHelper existingFileHelper, String modid) {
+        this.modid = modid;
+        this.existingFileHelper = existingFileHelper;
+        this.pathProvider = output.createPathProvider(PackOutput.Target.DATA_PACK, "computerHelp");
```
pupnewfster: Can we make the target folder path be `computer_help` instead of `computerHelp`? It would make it so it better matches the naming scheme for most asset and data things MC has. Just so that it is technically possible for them to be ResourceLocations by not being capitalized.

src/datagen/main/java/mekanism/common/MekanismDataGenerator.java
```diff
@@ -63,6 +64,7 @@ public static void gatherData(GatherDataEvent event) {
         gen.addProvider(event.includeServer(), new MekanismCrTExampleProvider(output, existingFileHelper));
         //Data generator to help with persisting data when porting across MC versions when optional deps aren't updated yet
         gen.addProvider(true, new PersistingDisabledProvidersProvider(output, recipeProvider.getDisabledCompats()));
+        gen.addProvider(event.includeServer(), new ComputerHelpProvider(output, existingFileHelper, Mekanism.MODID));
```
pupnewfster: Please move this to above the comment about persisting data. The `PersistingDisabledProvidersProvider` is very intentionally meant to be the very last provider we add.

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+import org.jetbrains.annotations.NotNull;
+
+@MethodsReturnNonnullByDefault
+@ParametersAreNotNullByDefault
+public class ComputerHelpProvider implements DataProvider {
+    private static final String[] METHOD_CSV_HEADERS = new String[]{"Class", "Method Name", "Params", "Returns", "Restriction", "Requires Public Security", "Description"};
+    private static final String[] ENUM_CSV_HEADERS = new String[]{"Type Name", "Values"};
+    private final PackOutput.PathProvider pathProvider;
+    private final ExistingFileHelper existingFileHelper;
```
pupnewfster: It doesn't seem like this `ExistingFileHelper` is used at all, is there a reason that is planned where having it available here would be useful? If not we may as well just not pass it and then pass it later once we have a use case.

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+
+    public ComputerHelpProvider(PackOutput output, ExistingFileHelper existingFileHelper, String modid) {
+        this.modid = modid;
+        this.existingFileHelper = existingFileHelper;
+        this.pathProvider = output.createPathProvider(PackOutput.Target.DATA_PACK, "computerHelp");
+    }
+
+    @Override
+    public CompletableFuture<?> run(CachedOutput pOutput) {
+        FactoryRegistry.load();
```
pupnewfster, line 84: Please just add a one line comment that says something like:
```
//Manually initialize the computer method factory registry
```

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+    private final String modid;
+
+    public ComputerHelpProvider(PackOutput output, ExistingFileHelper existingFileHelper, String modid) {
+        this.modid = modid;
+        this.existingFileHelper = existingFileHelper;
+        this.pathProvider = output.createPathProvider(PackOutput.Target.DATA_PACK, "computerHelp");
+    }
+
+    @Override
+    public CompletableFuture<?> run(CachedOutput pOutput) {
```
pupnewfster: Please rename this variable to `output` (as well as other places in this file that have it as `pOutput`). The reason the `p` was generated from super is that parchment prepends `p` to parameters to ensure they don't conflict name wise with any local or class level fields.

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+        FactoryRegistry.load();
+        Map<Class<?>, List<MethodHelpData>> helpData = FactoryRegistry.getHelpData();
+        Map<Class<?>, List<String>> enumValues = getEnumValues(helpData);
+        return CompletableFuture.allOf(
+              makeMethodsJson(pOutput, helpData),
+              makeMethodsCsv(pOutput, helpData),
+              makeEnumsCsv(pOutput, enumValues),
+              makeEnumsJson(pOutput, enumValues)
```
pupnewfster: While it doesn't really matter we probably should make the jsons and csvs in the same order. As in either have it be: `makeMethodsJson, makeMethodsCsv, makeEnumsJson, makeEnumsCsv` or `makeMethodsCsv, makeMethodsJson, makeEnumsCsv, makeEnumsJson`

Also random comment/question: Why is there no enum json generated but there is a csv one generated?

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+        );
+    }
+
+    @NotNull
+    private CompletableFuture<?> makeMethodsJson(CachedOutput pOutput, Map<Class<?>, List<MethodHelpData>> helpData) {
```
pupnewfster: I think we should replace this helper (and `makeEnumsJson`) with a more general one that we call instead:
```java
@NotNull
private <T> CompletableFuture<?> makeJson(CachedOutput output, String path, Codec<T> codec, T data) {
    JsonElement jsonElement = codec.encodeStart(JsonOps.INSTANCE, data).getOrThrow(false, e -> {
        throw new RuntimeException(e);
    });
    return DataProvider.saveStable(output, jsonElement, this.pathProvider.json(new ResourceLocation(this.modid, path)));
}
```

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+        JsonElement jsonElement = ENUMS_CODEC.encodeStart(JsonOps.INSTANCE, enumValues).getOrThrow(false, e -> {
+            throw new RuntimeException(e);
+        });
+        return DataProvider.saveStable(pOutput, jsonElement, this.pathProvider.json(new ResourceLocation(this.modid, "enums")));
+    }
+
+    @NotNull
+    private CompletableFuture<?> makeMethodsCsv(CachedOutput pOutput, Map<Class<?>, List<MethodHelpData>> helpData) {
+        return saveCSV(pOutput, this.pathProvider.file(new ResourceLocation(this.modid, "methods"), "csv"), METHOD_CSV_HEADERS, output -> {
+            List<Pair<String, List<MethodHelpData>>> friendlyList = helpData.entrySet().stream().map(entry -> Pair.of(getFriendlyName(entry.getKey()), entry.getValue())).sorted(Entry.comparingByKey()).toList();
```
pupnewfster: Given we only use things that exist on entry we might as well just use `Map.entry` instead of `Pair.of`, unless we maybe should not even do the `.toList()` after sorting the stream and instead just apply the things to the stream instead of looping the list.

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,274 @@
...
+        return "ComputerHelp:" + modid;
+    }
+
+    //implementation isn't quite "stable" like json, but input is always sorted
+    @SuppressWarnings("UnstableApiUsage")
+    static CompletableFuture<?> saveCSV(CachedOutput pOutput, Path pPath, String[] headers, IOConsumer<CsvOutput> rowGenerator) {
+        return CompletableFuture.runAsync(() -> {
+            try {
+                ByteArrayOutputStream bytearrayoutputstream = new ByteArrayOutputStream();
+                HashingOutputStream hashingoutputstream = new HashingOutputStream(Hashing.sha1(), bytearrayoutputstream);
```
pupnewfster: This should probably be using try with resources similar to how we do it in `BaseCrTExampleProvider`

annotation-processor/src/main/java/mekanism/visitors/AnnotationHelper.java
thiakil: This one uses the `ExecutableElement`, but the others don't, so they can just fall through to a null map key having no value

src/datagen/main/java/mekanism/common/MekanismDataGenerator.java
thiakil: that should be in the comment about it :P

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
thiakil: I think I just didn't run the datagen. It definitely does

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
thiakil: it's used like this as there are `IOException`s involved. Can use `Map.entry`, I didn't know about it when this was written

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
thiakil: In terms of the MC json stuff enforces sorting of keys, but the CSV can't do that without copying and significantly rewriting the class used for CSV abstraction

src/main/java/mekanism/common/integration/computer/annotation/ComputerMethod.java
thiakil: Only for convertable, as it will always be `Convertable<?>`, whereas `Either` has its type params defined

src/datagen/main/java/mekanism/common/MekanismDataGenerator.java
thiakil: hmm it's still showing the BoP files as being removed... was the datagen run on the main branch after updating deps?

src/main/java/mekanism/common/integration/computer/opencomputers2/MekanismDevice.java
thiakil: even writing javadoc et al I find they get redundant or duplicated. So for now I'd say not

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
thiakil: I didn't know that was a thing

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
thiakil: personally I prefer the explicit `int.class` way - it makes it clear that it's referring to the primitive and not the boxed type

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
thiakil: in practice, its not a problem due to the sorting mentioned

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
thiakil: neat, TIL

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
thiakil: > Also given we have items here do we also want the primitive Fluid and Chemical?

When they become acceptable types

src/main/java/mekanism/common/integration/computer/annotation/SyntheticComputerMethod.java
thiakil: split for now, though it may get re-joined using a placeholder for `get` or `set`

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
thiakil: done, but only BAOS needs closing, the HOS just calls the BAOS param's close method

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
thiakil: found it in the server jar 👍

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
pupnewfster: Yeah sounds reasonable to me, though maybe we want to expand on the comment about the input always being sorted that: the row generator should be responsible for ensuring the input is sorted and consistent from one run to the next

src/main/java/mekanism/common/integration/computer/MethodHelpData.java
pupnewfster: Fair enough

src/datagen/main/java/mekanism/common/MekanismDataGenerator.java
pupnewfster: I believe it has been ran on the main branch since deps were updated. Which files specifically is it causing to be removed?

src/datagen/main/java/mekanism/common/MekanismDataGenerator.java
```diff
@@ -88,4 +103,26 @@ public static void bootstrapConfigs(String modid) {
...
+     */
+    @SuppressWarnings({"UnstableApiUsage", "deprecation"})
+    public static CompletableFuture<?> save(CachedOutput cache, IOConsumer<OutputStream> osConsumer, Path path) {
+        return CompletableFuture.runAsync(() -> {
+            try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+                 HashingOutputStream hashingOutputStream = new HashingOutputStream(Hashing.sha1(), outputStream)) {
+                osConsumer.accept(hashingOutputStream);
+                cache.writeIfNeeded(path, outputStream.toByteArray(), hashingOutputStream.hash());
+            } catch (IOException ioexception) {
+                LOGGER.error("Failed to save file to {}", path, ioexception);
```
pupnewfster: This can just reference the same logger as the others use: `DataProvider.LOGGER`

src/main/java/mekanism/common/tile/multiblock/TileEntityInductionPort.java
```diff
@@ -76,14 +76,14 @@ public int getRedstoneLevel() {
     }
 
     //Methods relating to IComputerTile
-    @ComputerMethod
+    @ComputerMethod(methodDescription = "true -> output, false -> input. TODO change this to enum")
```
pupnewfster: We may want a proper "TODO" that is a comment here as well so that if someone looks at intellij's todo tab it is listed instead of having to actually do a find in files for "TODO"

src/main/java/mekanism/common/tile/multiblock/TileEntitySPSPort.java
```diff
@@ -96,14 +96,14 @@ public int getRedstoneLevel() {
     }
 
     //Methods relating to IComputerTile
-    @ComputerMethod
+    @ComputerMethod(methodDescription = "true -> output, false -> input. TODO change this to enum")
```
pupnewfster: We may want a proper "TODO" that is a comment here as well so that if someone looks at intellij's todo tab it is listed instead of having to actually do a find in files for "TODO"

src/main/java/mekanism/common/tile/multiblock/TileEntitySPSPort.java
```diff
@@ -96,14 +96,14 @@ public int getRedstoneLevel() {
...
     boolean getMode() {
         return getActive();
     }
 
-    @ComputerMethod
+    @ComputerMethod(methodDescription = "true -> output, false -> input. TODO change this to enum")
```
pupnewfster: We may want a proper "TODO" that is a comment here as well so that if someone looks at intellij's todo tab it is listed instead of having to actually do a find in files for "TODO"

src/main/java/mekanism/common/integration/computer/ComputerMethodFactory.java
```diff
@@ -80,6 +81,10 @@ void bindTo(@Nullable T subject, BoundMethodHolder holder) {
         }
     }
 
+    public List<MethodHelpData> getHelpData() {
+        return this.methods.stream().sorted(METHODDATA_COMPARATOR).map(MethodHelpData::from).toList();
```
pupnewfster: Given how this method is used in `FactoryRegistry#getHelpData` I believe this is intended to be using `collect(Collectors.toList())` rather than `toList()` as it seems like the desired outcome is for the returned list to be mutable.

src/main/java/mekanism/common/integration/computer/TableType.java
```diff
@@ -0,0 +1,73 @@
...
+import com.mojang.serialization.codecs.RecordCodecBuilder;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import mekanism.common.util.MekCodecs;
+
+/**
+ * Defines the format of a "table" (Map) format of a Mekanism data structure
+ */
+public record TableType(String description, String humanName, Map<String, FieldType> fields, Class<?> extendz) {
```
pupnewfster: Instead of using `extendz` to make it so that it doesn't clash with the keyword `extends`, how do you feel about using `extended`?

src/main/java/mekanism/common/integration/computer/TableType.java
```diff
@@ -0,0 +1,73 @@
...
+        public Builder addField(String name, Class<?> javaType, String description, Class<?>... javaExtra) {
+            if (javaExtra == null) {
+                javaExtra = ComputerMethodFactory.NO_CLASSES;
+            }
+            this.fields.put(name, new FieldType(description, javaType, MethodHelpData.getHumanType(javaType, javaExtra), javaExtra));
+            return this;
+        }
+
+        public TableType build(Map<Class<?>, TableType> destination) {
+            TableType tableType = new TableType(description, humanName, fields, extendz);
```
pupnewfster: Do we want to ensure a copy on `fields` here? Given otherwise technically if we ended up re-using a builder and add a field after having built it once then it will also be added to the already built variant. Otherwise we can just add docs to the methods that after `build` is called none of the other methods should be used.

annotation-processor/src/main/java/mekanism/builder/ComputerHandlerBuilder.java
```diff
@@ -558,49 +596,77 @@ private CodeBlock callTargetMethod(String annotatedName, boolean isStatic, List<
...
-        if (modsRequired.isEmpty()) {
-            registerMethodBuilder.add("NO_STRINGS, ");
-        } else {
-            registerMethodBuilder.add("$L, ", annotationValues.getLiteral("requiredMods", "NO_STRINGS"));
+        if (!modsRequired.isEmpty()) {
+            registerMethodBuilder.add(".requiredMods($L)", annotationValues.getLiteral("requiredMods", "NO_STRINGS"));
         }
         //threadsafe
-        registerMethodBuilder.add("$L, ", threadSafeLiteral);
+        if (!(threadSafeLiteral instanceof Boolean) || ((Boolean) threadSafeLiteral)) {
```
pupnewfster: What are the cases this may not be a boolean?

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
```diff
@@ -0,0 +1,309 @@
...
+
+@MethodsReturnNonnullByDefault
+@ParametersAreNotNullByDefault
+public class ComputerHelpProvider implements DataProvider {
+    private static final String[] METHOD_CSV_HEADERS = {"Class", "Method Name", "Params", "Returns", "Restriction", "Requires Public Security", "Description"};
+    private static final String[] ENUM_CSV_HEADERS = {"Type Name", "Values"};
+    private final PackOutput.PathProvider pathProvider;
+    private final String modid;
+
+    public ComputerHelpProvider(PackOutput output, String modid) {
```
pupnewfster, line 76: Does this include mekanism generators' data and that is why this provider is never called with mekanism generator's modid?

src/generators/java/mekanism/generators/common/tile/TileEntityGenerator.java
```diff
@@ -83,4 +83,7 @@ protected ISyncableData syncableMaxOutput() {
     public BasicEnergyContainer getEnergyContainer() {
         return energyContainer;
     }
-}
\ No newline at end of file
+
+    @ComputerMethod(methodDescription = "Get the amount of energy produced by this generator in the last tick.")
+    abstract FloatingLong getProductionRate();
```
pupnewfster, line 88: As CodeQL is warning please add `@Override` annotations to the implementers of this.

src/main/java/mekanism/common/integration/computer/MethodData.java
```diff
@@ -0,0 +1,95 @@
...
+            return this;
+        }
+
+        public Builder<T> threadSafe() {
+            this.threadSafe = true;
+            return this;
+        }
+
+        public Builder<T> arguments(String[] argumentNames, Class<?>[] argClasses) {
+            this.argumentNames = argumentNames;
```
pupnewfster, line 78: Do we want extra validation either here in the builder or maybe in the compact constructor of `MethodData` to validate that we have the same number of argument names as argument classes?

src/main/java/mekanism/common/integration/computer/BoundMethodHolder.java
```diff
@@ -23,6 +37,11 @@ public abstract class BoundMethodHolder {
 
     protected Lazy<String[]> methodNames = Lazy.of(()->this.methods.keys().toArray(new String[0]));
 
+    public BoundMethodHolder() {
```
pupnewfster: This can just be `protected` as it is an abstract class so everything has to implement it anyways if it wants to be called.

src/main/java/mekanism/common/integration/computer/BoundMethodHolder.java
```diff
@@ -81,4 +100,30 @@ public int hashCode() {
             return result;
         }
     }
+
+    public static Object generateHelp(ListMultimap<String, BoundMethodData<?>> methods, BaseComputerHelper helper) {
```
pupnewfster, line 104: This should be marked as `@Nullable`

src/main/java/mekanism/common/integration/computer/BoundMethodHolder.java
```diff
@@ -81,4 +100,30 @@ public int hashCode() {
...
+            return null;
+        }
+        Map<String, MethodHelpData> helpItems = new HashMap<>();
+        methods.values().stream().sorted(METHODDATA_COMPARATOR).forEach(md->
+              helpItems.put(md.name()+"("+String.join(", ", md.argumentNames())+")", MethodHelpData.from(md))
+        );
+        return helper.convert(helpItems, helper::convert, helper::convert);
+    }
+
+    public static Object generateHelpSpecific(ListMultimap<String, BoundMethodData<?>> methods, BaseComputerHelper helper) throws ComputerException {
```
pupnewfster, line 115: This should be marked as `@Nullable`

pupnewfster (changes requested):

src/datagen/main/java/mekanism/common/MekanismDataGenerator.java
thiakil: it doesn't always happen, but when it does, its these ones
image

src/datagen/main/java/mekanism/common/integration/computer/ComputerHelpProvider.java
thiakil: yes. Since it's 99% for website documentation where they wouldn't be separated I didn't bother splitting them

pupnewfster: No idea how large a file it is so it may not matter, but do we potentially want to be excluding it from the built jar then?

thiakil: I think it's good to have in there somewhere as a definitive reference, at least the markdown/yaml file. CSV was just to add to the google sheet

It should compress down to very little, as the strings would share zip dictionary entries with the class files as they would contain all the strings

pupnewfster: Sounds reasonable enough to me

pupnewfster (approved):

Issue #7855, opened 2023-08-31
dupe with mekanism personal chest and fast fingers


Xplodin:
### Issue description

a user in my discord let me know of a bug that deals with personal chest and dropping and opening them at the same time

### Steps to reproduce

https://www.imgur.com/RotgkTW
^
it basically right click, 10ms, drop item

It works on all versions of mekanism from what I heard

### Minecraft version

1.19.2 (Latest)

### Forge version

1.20.1 - 1.12.2

### Mekanism version

10.3.9 (Latest)

### Other relevant versions

apparently it works from latest - 1.12.2

### If a (crash)log is relevant for this issue, link it here: (It's almost always relevant)

sorry, no logs are required this time around.

pupnewfster:
Does this happen with just Mekanism or do you have any performance mods or inventory modifications mods like one that changes how item transfer happens. As far as I know we have fixed all the remaining issues related to this bug unless some other mod is breaking it. (Though I will try look into it some regardless)

PR #7867, opened 2023-09-17
Move Personal Storage Item Inventory to World Saved Data

thiakil:

- Instead of being on the itemstack, the inv is stored in world saved data, one file per owner uuid
- Stack only has the Security/owner info and an inv UUID
- Refactors `ISustainedInventory` a little to avoid accidental misuse and slightly redundant varargs
- Removed some no longer needed fallbacks from loot generation (9d53c3a8ee80ab8a5542e0b4798b1569fbfa4ad3)
- Added ability for Inventory Attribute to specify a custom loot generation
- Old saved item inventories will be converted when they're opened or placed
- Item GUIs will now ensure that the hand that opened it still contains the item (#7855 though the dupe is resolved by the move to saved data)

## Breaking(-ish) changes
- Clone/pick block will no longer contain a copy of the inventory (so WAILA/TOP doesn't cause an explosion in the saved data size)


src/api/java/mekanism/api/NBTConstants.java
```diff
@@ -159,6 +159,7 @@ private NBTConstants() {
     public static final String OWNER_UUID = "owner";
     public static final String PARTIAL_WASTE = "partialWaste";
     public static final String PATH_TYPE = "pathType";
+    public static final String PERSONAL_STORAGE_ID = "personalStorageId";
```
pupnewfster, line 163: We should add a since javadoc on this field

src/main/java/mekanism/common/block/attribute/Attributes.java
```diff
@@ -44,9 +48,29 @@ private AttributeSecurity() {
...
+    public static class AttributeInventory<T extends ConditionUserBuilder<T> & FunctionUserBuilder<T>> implements Attribute {
+
+        private final BiFunction<T, CopyNbtFunction.Builder, Boolean> customLootBuilder;
+
+        /**
+         * Create an Inventory attribute with custom loot function handling
+         * @param customLootBuilder consumes the Builders and returns `hasContents` for use in {@link mekanism.common.loot.table.BaseBlockLootTables#dropSelfWithContents(java.util.List)}
+         */
+        @SuppressWarnings("JavadocReference")
+        public AttributeInventory(BiFunction<T, CopyNbtFunction.Builder, @NotNull Boolean> customLootBuilder) {
```
pupnewfster: This parameter and the field should probably both have nullable annotations

src/main/java/mekanism/common/block/attribute/Attributes.java
```diff
@@ -44,9 +48,29 @@ private AttributeSecurity() {
...
 
         private AttributeInventory() {
+            this(null);
+        }
+
+        public boolean hasCustomLoot() {
+            return this.customLootBuilder != null;
+        }
+
+        public BiFunction<T, CopyNbtFunction.Builder, Boolean> getCustomLootBuilder() {
```
pupnewfster: Is there a use case for this returning the `BiFunction` rather than just being a boolean method that accepts the params and passes them to the `BiFunction`?

src/main/java/mekanism/common/item/loot/PersonalStorageContentsLootFunction.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+        return itemStack;
+    }
+
+    @Override
+    public Set<LootContextParam<?>> getReferencedContextParams() {
+        return REFERENCED_PARAMS;
+    }
+
+    public static class Serialiser implements Serializer<PersonalStorageContentsLootFunction> {
```
pupnewfster: Given the rest of the codebase is american english this should be `Serializer` rather than `Serialiser`. Or if the reason this is named the way it is so that the implementation doesn't need to be a FQN then maybe `PersonalStorageSerializer`

src/main/java/mekanism/common/registries/MekanismContainerTypes.java
```diff
@@ -165,7 +165,7 @@ private MekanismContainerTypes() {
     public static final ContainerTypeRegistryObject<MekanismTileContainer<TileEntityEnergyCube>> ENERGY_CUBE = CONTAINER_TYPES.custom("energy_cube", TileEntityEnergyCube.class).armorSideBar(180, 41, 0).build();
     public static final ContainerTypeRegistryObject<MekanismTileContainer<TileEntityInductionCasing>> INDUCTION_MATRIX = CONTAINER_TYPES.custom("induction_matrix", TileEntityInductionCasing.class).armorSideBar(-20, 41, 0).build();
     public static final ContainerTypeRegistryObject<MekanismTileContainer<TileEntityBoilerCasing>> THERMOELECTRIC_BOILER = CONTAINER_TYPES.custom("thermoelectric_boiler", TileEntityBoilerCasing.class).offset(20, 0).build();
-    public static final ContainerTypeRegistryObject<PersonalStorageItemContainer> PERSONAL_STORAGE_ITEM = CONTAINER_TYPES.register("personal_storage_item", ItemBlockPersonalStorage.class, PersonalStorageItemContainer::new);
+    public static final ContainerTypeRegistryObject<PersonalStorageItemContainer> PERSONAL_STORAGE_ITEM = CONTAINER_TYPES.register("personal_storage_item", () -> MekanismItemContainerType.item(ItemBlockPersonalStorage.class, PersonalStorageItemContainer::new));
```
pupnewfster, line 168: What is the reason behind changing this line? As I believe the previous variant is just a helper for what this is now doing?

src/main/java/mekanism/common/lib/inventory/personalstorage/PersonalStorageData.java
```diff
@@ -0,0 +1,80 @@
...
+     *          {
+     *              "uuid": UUID,
+     *              "inv": PersonalStorageItemInventory
+     *          }
+     *     ]
+     * }
+     */
+    @Override
+    public void load(@NotNull CompoundTag nbt) {
+        ListTag entries = nbt.getList("data", Tag.TAG_COMPOUND);
```
pupnewfster: `NBTConstants.DATA`

src/main/java/mekanism/common/lib/inventory/personalstorage/PersonalStorageData.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public CompoundTag save(CompoundTag pCompoundTag) {
+        ListTag entries = new ListTag();
+        inventoriesById.forEach((uuid, inv) -> {
+            CompoundTag nbtEntry = new CompoundTag();
+            nbtEntry.putUUID("uuid", uuid);
+            nbtEntry.put("inv", inv.serializeNBT());
+            entries.add(nbtEntry);
+        });
+        pCompoundTag.put("data", entries);
```
pupnewfster: `NBTConstants.DATA`

src/main/java/mekanism/common/lib/inventory/personalstorage/PersonalStorageData.java
```diff
@@ -0,0 +1,80 @@
...
+        for (int i = 0; i < entries.size(); i++) {
+            CompoundTag entry = entries.getCompound(i);
+            PersonalStorageItemInventory inv = createInventory();
+            inv.deserializeNBT(entry.getList("inv", Tag.TAG_COMPOUND));
+            inventoriesById.put(entry.getUUID("uuid"), inv);
+        }
+    }
+
+    @Override
+    public CompoundTag save(CompoundTag pCompoundTag) {
```
pupnewfster: Remove the `p` from the parameter name

src/main/java/mekanism/common/lib/inventory/personalstorage/PersonalStorageData.java
```diff
@@ -0,0 +1,80 @@
...
+     * }
+     */
+    @Override
+    public void load(@NotNull CompoundTag nbt) {
+        ListTag entries = nbt.getList("data", Tag.TAG_COMPOUND);
+        for (int i = 0; i < entries.size(); i++) {
+            CompoundTag entry = entries.getCompound(i);
+            PersonalStorageItemInventory inv = createInventory();
+            inv.deserializeNBT(entry.getList("inv", Tag.TAG_COMPOUND));
+            inventoriesById.put(entry.getUUID("uuid"), inv);
```
pupnewfster: These should use `NBTConstants`, whether it is using `ITEMS` and `INVENTORY_ID` or whether we add two more for this. (Likewise the saving then should use the same constants as are used here)

src/main/java/mekanism/common/item/loot/PersonalStorageContentsLootFunction.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Loot function which copies the Personal Storage inventory to the saved data and adds an inv id to the stack
+ */
+@MethodsReturnNonnullByDefault
+@ParametersAreNotNullByDefault
+public class PersonalStorageContentsLootFunction implements LootItemFunction {
+
+    private static final PersonalStorageContentsLootFunction INSTANCE = new PersonalStorageContentsLootFunction();
+    private static final Set<LootContextParam<?>> REFERENCED_PARAMS = ImmutableSet.of(LootContextParams.BLOCK_ENTITY);
```
pupnewfster: Rather than doing `ImmutableSet.of` we may as well use the builtin java variant `Set.of`

src/main/java/mekanism/common/item/loot/PersonalStorageContentsLootFunction.java
```diff
@@ -0,0 +1,86 @@
...
+    public LootItemFunctionType getType() {
+        return MekanismLootFunctions.PERSONAL_STORAGE_LOOT_FUNC.get();
+    }
+
+    @Override
+    public ItemStack apply(ItemStack itemStack, LootContext lootContext) {
+        BlockEntity blockEntity = lootContext.getParam(LootContextParams.BLOCK_ENTITY);
+        if (blockEntity instanceof TileEntityPersonalStorage personalStorage) {
+            List<IInventorySlot> tileSlots = personalStorage.getInventorySlots(null);
+            //only save if its empty
```
pupnewfster: I believe this comment is meant to say `only save if it's not empty`

src/main/java/mekanism/common/item/block/ItemBlockPersonalStorage.java
```diff
@@ -18,13 +17,11 @@
...
 import net.minecraft.world.item.context.UseOnContext;
 import net.minecraft.world.level.Level;
-import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraftforge.common.util.FakePlayer;
 import org.jetbrains.annotations.NotNull;
 
-public class ItemBlockPersonalStorage<BLOCK extends Block & IHasDescription & IPersonalStorage> extends ItemBlockTooltip<BLOCK> implements IItemSustainedInventory,
-      IGuiItem {
+public class ItemBlockPersonalStorage<BLOCK extends BlockPersonalStorage<?,?>> extends ItemBlockTooltip<BLOCK> implements IGuiItem {
```
pupnewfster: Why does this no longer implement `IItemSustainedInventory`? One feature this now doesn't support is `InventoryUtils#dropItemContents`. Which yes potentially should be overridden so we only drop contents when the item is destroyed if it is public, but previously I had them drop them regardless just so that player's have more of a chance at their items not all getting destroyed (which I still think makes the most sense).

Additionally we probably want to override `onDestroyed(ItemEntity, DamageSource` so that when the stack is destroyed then we can clean up the WSD and make it not keep track of the contents for the stack anymore.

src/main/java/mekanism/common/lib/inventory/personalstorage/PersonalStorageManager.java
```diff
@@ -0,0 +1,95 @@
...
+     * @return the existing or new inventory
+     */
+    public static PersonalStorageItemInventory getInventoryFor(ItemStack stack) {
+        UUID owner = SecurityUtils.get().getOwnerUUID(stack);
+        if (owner == null) {
+            throw new IllegalStateException("Stack has no owner!");
+        }
+        UUID invId = getInventoryId(stack);
+        PersonalStorageItemInventory storageItemInventory = forOwner(owner).getOrAddInventory(invId);
+        ListTag legacyData = ItemDataUtils.getList(stack, NBTConstants.ITEMS);
```
pupnewfster, line 52: Do we want a TODO to remove this at some point? Such as `//TODO - 1.21: Remove legacy loading`?

pupnewfster (changes requested):

src/main/java/mekanism/common/registries/MekanismContainerTypes.java
thiakil: `IMekanismItemContainerFactory` vs `IMekanismSidedItemContainerFactory` - I copied what the QIO dashboard does

src/main/java/mekanism/common/item/block/ItemBlockPersonalStorage.java
thiakil: Well it not longer satisfies the general contract of that interface as the itemstack itself doesn't contain the inventory. Key among it being that the Loot functions try to copy NBT data.

This kind of destroying thing came up on Discord when this was being discussed, I think it was in Neoforged, that if someone loses their personal chest / backpack, it's quite useful for an admin to be able to restore it. The rest of the world save size will always dwarf the lost space of one chest, even if it's full.

src/main/java/mekanism/common/registries/MekanismContainerTypes.java
pupnewfster: Ahh I misread which class the call was against.

src/main/java/mekanism/common/item/block/ItemBlockPersonalStorage.java
pupnewfster: I still need to do a bit of testing relating to the commit I just made, but what are your thoughts on that as a midway solution?

pupnewfster (approved):

PR #7903, opened 2023-10-21
Workaround Axe Enchantments on Paxel by Checking an Axe

### LINKED ISSUE #7860: Cannot use axe enchants on paxel
### Issue description

I am unable to enchant paxels with axe-compatible enchantments using any method - vanilla or modded - outside of creative mode. It seems that there was a similar issue in the past ([4122](https://github.com/mekanism/Mekanism/issues/4122)) with a commit ([f5dfe3f](https://github.com/tomelfring/Mekanism/commit/f5dfe3f19235669ee4c66058645d5a66e4321d33)) merged into 1.10 but I cannot find any similar documentation for Minecraft 1.19.2/Mekanism 10.3.8.477.

Any assistance would be appreciated - unfortunately I cannot update to a later version of Mekanism due to the changed Forge version requirement. Is there a way to patch this with a datapack or somesuch?

### Steps to reproduce

_No response_

### Minecraft version

1.19.2 (Latest)

### Forge version

43.2.3

### Mekanism version

10.3.8

### Other relevant versions

_No response_

### If a (crash)log is relevant for this issue, link it here: (It's almost always relevant)

_No response_

ChampionAsh5357:
Closes #7860 

Allows the paxel to have axe enchantments by checking whether the enchantment can be applied to an axe. I opted for this solution compared extending `AxeItem` since that class assumes the tag is `MINEABLE_WITH_AXE` while paxels have their own tags. This would mean that all of the logic associated with the tag would need to be replaced and swapped out. In favor of avoiding redundancy, it is simpler to just check whether the item is an axe item.

Modded enchantments should check whether the action can perform `AXE_DIG`, so it should remain compatible.

As this is a workaround, this should work until the mod loader API switches out the enchantment logic for checking tool actions instead.


pupnewfster:
The one issue with this solution is I believe sharpness is allowed on axes but only via enchanted books and not via the enchantment table. Or at least there was something like that when I looked into it some. Now I don’t know how much that actually matters but definitely needs some more thought. Cc: @thiakil

thiakil:
Can we use `enchantment.category.canEnchant(AXE_item)` instead? (preferably with the paxel's normal axe version)
I'm not a fan of making an itemstack here

Re sharpness: on a quick glance look like the methods an anvil uses will pass through here by default so should be fine

pupnewfster:
My point is it works for anvil here but it adds enchantments as possible to the enchantment table that are normally anvil only.

thiakil:
how so?

ChampionAsh5357:
Ah, I see what you're talking about. `Enchantment#canEnchant` is what's used to check whether the item can be enchanted with that instance at all while `Enchantment#canApplyAtEnchantingTable` checks only if the logic can be applied at an enchanting table. `#canEnchant` calls `#canApplyAtEnchantingTable`.

So, this would probably need to be addressed in NeoForge as a way to disambiguate the two. Otherwise, the only solution I could think of to maintain the logic is to mixin in a check somehow, which wouldn't be very mod-compatible since you revolve. Another solution would just be to say paxels are cool like that and can have the enchantments at the enchanting table.

This is assuming we are still using this method. The safest way would probably just be to extend `AxeItem` and just override all the tag checks with your own, which is a bit of a logic rewrite, but there's not much to do there.

I'll wait for further opinions before going forward here.

thiakil:
Which category are axe enchants? (or do they specifically check for axe?)

ChampionAsh5357:
There is none, it's essentially a hardcoded check in `Enchantment#canEnchant` within the `DamageEnchantment`. If anything, they fall under the `DIGGER` category by default.

thiakil:
Hmm I see. Ideally a forge fix for being hardcoded would be the way.

Though if you wanna see how it looks with paxels extending axe I'd take a look

ChampionAsh5357:
@thiakil here you are. It simply just extends `AxeItem` and changes any indirect checks on the tag to direct ones.

src/tools/java/mekanism/tools/common/item/ItemMekanismPaxel.java
```diff
@@ -189,4 +191,15 @@ public void addToBuilder(ImmutableMultimap.Builder<Attribute, AttributeModifier>
         builder.put(Attributes.ATTACK_DAMAGE, new AttributeModifier(BASE_ATTACK_DAMAGE_UUID, "Tool modifier", getAttackDamage(), Operation.ADDITION));
         builder.put(Attributes.ATTACK_SPEED, new AttributeModifier(BASE_ATTACK_SPEED_UUID, "Tool modifier", material.getPaxelAtkSpeed(), Operation.ADDITION));
     }
+
+    @Override
+    public boolean isCorrectToolForDrops(BlockState state) {
+        // Some modded items may not check the stack sensitive version
```
thiakil: given it's the same code, only one of these should exist (calling the other if absolutely required)

src/tools/java/mekanism/tools/common/item/ItemMekanismPaxel.java
```diff
@@ -37,26 +37,28 @@
...
 import org.jetbrains.annotations.Nullable;
 
-public class ItemMekanismPaxel extends DiggerItem implements IHasRepairType, IAttributeRefresher {
+public class ItemMekanismPaxel extends AxeItem implements IHasRepairType, IAttributeRefresher {
 
     private static final ToolAction PAXEL_DIG = ToolAction.get("paxel_dig");
 
     private final IPaxelMaterial material;
     private final AttributeCache attributeCache;
+    private static final TagKey<Block> MINEABLE_WITH_PAXEL = ToolsTags.Blocks.MINEABLE_WITH_PAXEL;
```
pupnewfster: There isn't much reason to inline this feel like this as opposed to referencing it the one or two places it is used so that the discoverability of actual usages can be done in one step instead of two.

src/tools/java/mekanism/tools/common/item/ItemMekanismPaxel.java
ChampionAsh5357: Okay, I'll remove the Forge added method since it delegates to this one.

src/tools/java/mekanism/tools/common/item/ItemMekanismPaxel.java
ChampionAsh5357: Fair enough.

pupnewfster:
If we are doing it by making it extend the axe we may want to make:
- canPerformAction call super instead of checking the default actions
- Make usOn do something like:
```java
InteractionResult axeResult = super.useOn(context);
if (result != InteractionResult.PASS) {
    return result;
}
```
And then remove the `useAsAxe` method. That way we won't have to worry about if mojang adds more features to the axe at some point as it will be handled by the super call.

Also in regards to overriding one vs both `isCorrectToolForDrops` I don't think we have a choice as in `DiggerItem` both are overridden to reference `blocks` rather than passthrough to the less specific variant. Though an alternate solution would be to add an AT in Mekanism Tools' AT file that definalizes and makes blocks `protected` and then update the tag key that is checked to being the paxel tag.

ChampionAsh5357:
Ok, I'll get around to it after the 5th since I won't have a computer that can run Java until then.

ChampionAsh5357:
Okay, I've addressed the comments and added back in the second method for `#isCorrectToolForDrops` because of pup's comment. I could go for the definalization method if wanted, however.

src/tools/java/mekanism/tools/common/item/ItemMekanismPaxel.java
```diff
@@ -96,28 +96,33 @@ public InteractionResult useOn(UseOnContext context) {
         BlockPos blockpos = context.getClickedPos();
         Player player = context.getPlayer();
         BlockState blockstate = world.getBlockState(blockpos);
```
pupnewfster: This should be below attempting to use it as an axe as we unfortunately can't make use of it and super will calculate it anyway, but then at least if we can use it as an axe we only have to look up the block state once instead of twice.

pupnewfster (commented):
One more minor nitpick but other than that it looks pretty good (assuming when I test it things work as expected)

src/tools/java/mekanism/tools/common/item/ItemMekanismPaxel.java
```diff
@@ -189,4 +168,15 @@ public void addToBuilder(ImmutableMultimap.Builder<Attribute, AttributeModifier>
         builder.put(Attributes.ATTACK_DAMAGE, new AttributeModifier(BASE_ATTACK_DAMAGE_UUID, "Tool modifier", getAttackDamage(), Operation.ADDITION));
         builder.put(Attributes.ATTACK_SPEED, new AttributeModifier(BASE_ATTACK_SPEED_UUID, "Tool modifier", material.getPaxelAtkSpeed(), Operation.ADDITION));
     }
+
+    // Need to override both method as DiggerItem performs two different behaviors
+    @Override
+    public boolean isCorrectToolForDrops(BlockState state) {
```
thiakil, line 174: one of these should just call the other, this solves both cases and removes the duplicate

PR #7921, opened 2023-11-11
Add Trusted Frequencies

Bommels05:
Adds Trusted Frequencies by using `SecurityMode` instead of a `publicFreq` boolean. I decided to add an owner id to `FrequencyIdentity` so the server can check if the player is trusted by the owner and then use the UUID of the owner to execute the actions. I also added a method to `SecurityUtils` to check that.
I changed the buttons to now have icons instead of text so there is enough space for all three.


TureBentzin:
That's a feature I always looked for. Way better for teams building bases together.

pupnewfster:
Thanks for getting this started! I went ahead and merged it as part of https://github.com/mekanism/Mekanism/commit/153b8ab57df5c04dd664a66641d7ac2af324a361. I updated this to 1.20.4 and also fixed a handful of bugs related to verification of if a player has permissions to say delete a frequency, and ensure trusted frequencies get unset in cases where the owner of the item can't access it anymore.

PR #7968, opened 2023-12-26
Refactor salt disk feature to call vanilla placement method.

KdotJPG:

The surface salt deposit placements currently use a copy of the vanilla code. This causes them not to pick up changes made by other mods that try to break up the monotony of (or otherwise change) the shapes of the vanilla disks:

image

By reworking the disk feature in Mekanism to call the vanilla code, the changes are now picked up:

image

Without the mods in question present, Mekanism behaves as it did before:

image

### DIFF
--- a/src/main/java/mekanism/common/world/ResizableDiskConfig.java
+++ b/src/main/java/mekanism/common/world/ResizableDiskConfig.java
@@ -2,30 +2,30 @@ package mekanism.common.world;
 
 import com.mojang.serialization.Codec;
 import com.mojang.serialization.codecs.RecordCodecBuilder;
-import java.util.List;
 import java.util.function.IntSupplier;
 import mekanism.common.config.MekanismConfig;
 import mekanism.common.config.WorldConfig.SaltConfig;
 import net.minecraft.util.valueproviders.IntProvider;
 import net.minecraft.world.level.block.Blocks;
-import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.levelgen.blockpredicates.BlockPredicate;
 import net.minecraft.world.level.levelgen.feature.configurations.FeatureConfiguration;
+import net.minecraft.world.level.levelgen.feature.stateproviders.RuleBasedBlockStateProvider;
 
 public class ResizableDiskConfig implements FeatureConfiguration {
 
     public static final Codec<ResizableDiskConfig> CODEC = RecordCodecBuilder.create(builder -> builder.group(
-          BlockState.CODEC.fieldOf("state").forGetter(config -> config.state)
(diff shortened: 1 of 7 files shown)


src/main/java/mekanism/common/world/ResizableDiskConfig.java
```diff
@@ -2,30 +2,30 @@
...
     public static final Codec<ResizableDiskConfig> CODEC = RecordCodecBuilder.create(builder -> builder.group(
-          BlockState.CODEC.fieldOf("state").forGetter(config -> config.state)
-    ).apply(builder, state -> new ResizableDiskConfig(state, MekanismConfig.world.salt)));
+            RuleBasedBlockStateProvider.CODEC.fieldOf("state_provider").forGetter(config -> config.stateProvider)
+    ).apply(builder, stateProvider -> new ResizableDiskConfig(stateProvider, MekanismConfig.world.salt)));
 
-    public final BlockState state;
+    public final RuleBasedBlockStateProvider stateProvider;
     public final IntProvider radius;
     public final IntSupplier halfHeight;
```
KdotJPG: This file changed to match the vanilla config as closely as possible. The only remaining difference is the `IntSupplier` since I am not aware of any clean approach towards making `configured_feature/salt.json` pull in the Mekanism config value for `halfHeight`.

src/main/java/mekanism/common/world/ResizableDiskReplaceFeature.java
```diff
@@ -19,38 +14,21 @@ public ResizableDiskReplaceFeature(Codec<ResizableDiskConfig> codec) {
...
-                                    break;
-                                }
-                            }
-                        }
-                    }
-                }
-            }
-        }
-        return placed;
+        ResizableDiskConfig config = context.config();
```
KdotJPG: If my understanding of escape analysis is correct, the following instantiations should be stack-allocated if the function is called often enough to be optimized by the JVM.

pupnewfster:
I will try to look at this when I get a chance, out of curiosity though so when I do get to testing it, what mod are you testing it with?

KdotJPG:
Simply Improved Terrain, 1.20(.1) alpha
https://www.curseforge.com/minecraft/mc-mods/simply-improved-terrain

All The Mods 9 (with Mekanism already present, and the above mod added)

pupnewfster:
Love this, it is great having less duplicate code as it makes porting easier. Did a couple minor cleaning related things to this PR so that more of the stuff can be overridden by changing the configured feature rather than just changing the config, and also fixed salt spawning outside of water as it seems MC moved it from being hardcoded like our copy had it to being a placement modifier

PR #7997, opened 2024-02-11
Add a data map for meka suit damage absorption

Matyrobbrt:
This allows both pack devs and mod devs to configure a damage type to only be partially (or fully) absorbed by a full meka suit.

The path of the data map is `data/mekanism/data_maps/damage_type/mekasuit_absorption.json` and maps a damage type (or tag) to a float [0, 1]:
```json
{
  "values": {
     "minecraft:trident": 0.12
  }
}
```


src/main/java/mekanism/common/config/GearConfig.java
```diff
@@ -399,10 +394,6 @@ public class GearConfig extends BaseMekanismConfig {
         mekaSuitUnspecifiedDamageRatio = CachedFloatValue.wrap(this, builder.comment("Percent of damage taken from other non explicitly supported damage types that don't bypass armor when the MekaSuit has enough power and a full suit is equipped.",
                     "Note: Support for specific damage types can be added by adding an entry for the damage type in the damageReductionRatio config.")
```
pupnewfster: Please update this comment

src/main/java/mekanism/common/config/GearConfig.java
```diff
@@ -151,7 +147,6 @@ public class GearConfig extends BaseMekanismConfig {
     public final CachedIntValue mekaSuitBootsArmor;
     public final CachedFloatValue mekaSuitToughness;
     public final CachedFloatValue mekaSuitKnockbackResistance;
-    public final CachedRL2FloatMapConfigValue mekaSuitDamageRatios;
```
pupnewfster, line 154: You can delete the CachedRL2FloatMapConfigValue class as well while you are at it

src/main/java/mekanism/common/datamaps/MekaSuitAbsorption.java
```diff
@@ -0,0 +1,8 @@
+package mekanism.common.datamaps;
+
+import com.mojang.serialization.Codec;
+
+public record MekaSuitAbsorption(float absorption) {
```
pupnewfster: We probably want to expose this to the API so addons can easily add and set their own additional defaults

src/main/java/mekanism/common/item/gear/ItemMekaSuitArmor.java
```diff
@@ -477,17 +479,18 @@ private static float getDamageAbsorbed(Player player, DamageSource source, float
...
+                                .map(registry -> registry.wrapAsHolder(source.type()).getData(MekanismDataMapTypes.MEKA_SUIT_ABSORPTION))
                                 .orElse(null)
                           );
-                    if (damageTypeName != null) {//Note: This should not be null unless something went wrong and the damage source is for a damage type that is not registered
-                        absorbRatio = MekanismConfig.gear.mekaSuitDamageRatios.get().get(damageTypeName);
+                    if (damageTypeAbsorb != null) {
+                        absorbRatio = damageTypeAbsorb.absorption();
                     }
                     if (absorbRatio == null) {
                         absorbRatio = MekanismConfig.gear.mekaSuitUnspecifiedDamageRatio.getAsFloat();
```
pupnewfster: This all can actually just become
```java
absorbRatio = source.typeHolder().unwrapKey()
     // Reference holders can query data map values
      .map(type -> source.typeHolder())
      //Note: In theory the above path should always be done as vanilla only makes damage sources with reference holders
      // but just in case have the fallback to look up the name from the registry
      .or(() -> player.level().registryAccess().registry(Registries.DAMAGE_TYPE).map(registry -> registry.wrapAsHolder(source.type())))
      .map(holder -> holder.getData(MekanismDataMapTypes.MEKA_SUIT_ABSORPTION))
      .map(MekaSuitAbsorption::absorption)
      .orElseGet(MekanismConfig.gear.mekaSuitUnspecifiedDamageRatio::get);
```

src/datagen/main/java/mekanism/common/MekanismDataMapsProvider.java
```diff
@@ -32,5 +36,8 @@ protected void gather() {
         builder(NeoForgeDataMaps.FURNACE_FUELS)
               .add(MekanismBlocks.CHARCOAL_BLOCK.getId(), new FurnaceFuel(16_000), false)
         ;
+
+        final var absorption = builder(MekanismDataMapTypes.MEKA_SUIT_ABSORPTION);
+        ItemMekaSuitArmor.BASE_ALWAYS_SUPPORTED.forEach(key -> absorption.add(key, new MekaSuitAbsorption(ItemMekaSuitArmor.getBaseDamageRatio(key)), false));
```
pupnewfster: Firstly we don't use `var` here, second I think it is reasonable to just let everything default to the unspecified of 1 except for sonic boom, so then we only need to add a value to the map for sonic boom at `0.75`. We then can remove the `ItemMekaSuitArmor#getBaseDamageRatio` method, and we can even inline `ItemMekaSuitArmor#BASE_ALWAYS_SUPPORTED` into our tag provider

src/main/java/mekanism/common/registries/MekanismDataMapTypes.java
```diff
@@ -0,0 +1,14 @@
...
+import mekanism.common.datamaps.MekaSuitAbsorption;
+import mekanism.common.registration.impl.DataMapTypeRegister;
+import net.minecraft.core.registries.Registries;
+import net.minecraft.world.damagesource.DamageType;
+import net.neoforged.neoforge.registries.datamaps.DataMapType;
+
+public class MekanismDataMapTypes {
+    public static final DataMapTypeRegister REGISTER = new DataMapTypeRegister(Mekanism.MODID);
+
+    public static final DataMapType<DamageType, MekaSuitAbsorption> MEKA_SUIT_ABSORPTION = REGISTER.registerSimple("mekasuit_absorption", Registries.DAMAGE_TYPE, MekaSuitAbsorption.CODEC);
```
pupnewfster: We probably want to expose this `DataMapType` to the API, unless there is some other way for addons and the like to potentially get it.

pupnewfster (changes requested):

src/main/java/mekanism/common/registries/MekanismDataMapTypes.java
Matyrobbrt: `RegistryManager#getDataMap` can be used for that, though it's a bit annoying.

src/api/java/mekanism/api/datamaps/MekaSuitAbsorption.java
```diff
@@ -0,0 +1,23 @@
...
+import com.mojang.serialization.Codec;
+import mekanism.api.MekanismAPI;
+import net.minecraft.resources.ResourceKey;
+import net.minecraft.resources.ResourceLocation;
+
+/**
+ * A {@link net.minecraft.core.registries.Registries#DAMAGE_TYPE damage type} data map that allows
+ * changing how much damage of a given type the meka suit should absorb.
+ *
+ * @param absorption how much damage will be absorbed
```
pupnewfster, line 12: Add a `@since 10.5.0`

pupnewfster (changes requested):

pupnewfster (approved):

PR #7999, opened 2024-02-12
Add missing lang conversion for "Stabilisation"

ChristianLW:
The Gyroscopic Stabilisation Unit has an incorrect name in UK and Australian English ("z" instead of "s"), as the word "Stabilization" isn't included in the autoconversion. This PR simply adds that missing entry.


thiakil:
Good catch, thanks!

PR #8003, opened 2024-02-22
Update to 1.20.4 CC:T

SquidDev:
This updates the CC:T to use the 1.20.4 version. There's a small change in the API which need to be accounted for:

 - `IArguments.escapes` mutates in place rather than returning a new value.

Also:
 - Use `PeripheralCapability.get()` rather than creating the capability ourselves.
 - Move "is mod loaded" checks to the constructor of `MekanismHooks`. These fields were being read too early, which meant the capabilities weren't being registered. We could move the initialisation, but it seemed better just to make sure this can never happen.


pupnewfster:
Thanks!

PR #8021, opened 2024-03-09
Qio recipe adjustments

thiakil:
image
image
image
image
image
image
image
image
image
image


pupnewfster:
That's a lot of ender pearls (which often are super annoying to get as a player), though I suppose it makes sense given it has no cables. Slightly unsure if it semi borders on being too early for it not requiring power (though I think that is just something we will have to experiment with, and it seems like it might be far enough back anyway).

This does mean that in some ways a digital miner is more expensive, which is not a bad thing. The one recipe that uses a teleportation core that we may want to change though is for the security desk. as making that require four ender pearls seems like overkill.

pupnewfster:
I adjusted a few of the recipes and fixed the merge conflict

image
image
image
image

thiakil:
Lgtm

krzj:
I have 1.20.1

PR #8046, opened 2024-04-03
Added "isBurning" ComputerMethod to FusionReactorMultiblockData

NullPointerCollection:

Added "isBurning" ComputerMethod to FusionReactorMultiblockData to check if a fusion reaction is occurring.

Fixed typo in OneInputCachedRecipe.

### DIFF
--- a/src/api/java/mekanism/api/functions/ConstantPredicates.java
+++ b/src/api/java/mekanism/api/functions/ConstantPredicates.java
@@ -4,7 +4,11 @@ import java.util.function.BiPredicate;
 import java.util.function.BooleanSupplier;
 import java.util.function.Predicate;
 import mekanism.api.AutomationType;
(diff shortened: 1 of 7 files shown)


src/datagen/generated/mekanism/data/mekanism/computer_help/jekyll.md
```diff
@@ -4579,5 +4584,5 @@ methods:
     returns:
       javaType: boolean
       type: boolean
-version: 10.5.13
+version: 0.0.0
```
pupnewfster: This seems wrong

NullPointerCollection: edit: looks like the version can't be parsed while having anything but 3 numbers (current version being 10.5.14.1), which is why datagen is defaulting to 0.0.0

PR #8051, opened 2024-04-06
Nonticking structural

pupnewfster:
- Makes Structural glass and reactor glass no longer be Ticking BEs
- Improves handling of structural blocks when they have multiple structures (still shouldn't be any cases this is a thing for yet)
- Improve handling so that placing structural blocks next to formed multiblocks is better at not causing them to reform

### DIFF
--- a/src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
+++ b/src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
@@ -53,21 +53,24 @@ public abstract class TileEntityStructuralMultiblock extends TileEntityMekanism 
         if (!removing) {
             //Don't try to remove it from the tile when the tile is the thing being removed
             if (structures.remove(structure.getManager(), structure)) {
+                boolean hasFormed = false;
+                boolean canAccess = false;
                 for (Structure struct : structures.values()) {
                     MultiblockData multiblock = getMultiblockData(struct);
                     if (multiblock != null && multiblock.isFormed()) {
-                        updateFormedMultiblock(true, multiblock.allowsStructuralGuiAccess(this));
-                        return;
+                        hasFormed = true;
+                        canAccess |= multiblock.allowsStructuralGuiAccess(this);
                     }
                 }
-                updateFormedMultiblock(false, false);
+                updateFormedMultiblock(hasFormed, canAccess);
             }
         }
     }
 
     @Override
     public void multiblockFormed(MultiblockData multiblock) {
-        updateFormedMultiblock(true, multiblock.allowsStructuralGuiAccess(this));
+        //Note: We pass the existing value of canAccessGui, as then we will validate when interacting, and only allow interacting with a specific sub one
+        updateFormedMultiblock(true, canAccessGui || multiblock.allowsStructuralGuiAccess(this));
     }
 
     private void updateFormedMultiblock(boolean hasFormed, boolean canAccess) {
--- a/src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
+++ b/src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
@@ -52,17 +52,22 @@ public abstract class TileEntityStructuralMultiblock extends TileEntityMekanism 
     public void removeStructure(Structure structure) {
         if (!removing) {
             //Don't try to remove it from the tile when the tile is the thing being removed
-            if (structures.remove(structure.getManager(), structure)) {
-                boolean hasFormed = false;
-                boolean canAccess = false;
-                for (Structure struct : structures.values()) {
-                    MultiblockData multiblock = getMultiblockData(struct);
-                    if (multiblock != null && multiblock.isFormed()) {
-                        hasFormed = true;
-                        canAccess |= multiblock.allowsStructuralGuiAccess(this);
+            if (hasFormedMultiblock || canAccessGui) {
+                MultiblockManager<?> manager = structure.getManager();
+                if (structures.get(manager) == structure) {
+                    boolean hasFormed = false;
+                    boolean canAccess = false;
+                    for (Structure struct : structures.values()) {
+                        if (struct != structure) {
+                            MultiblockData multiblock = getMultiblockData(struct);
+                            if (multiblock != null && multiblock.isFormed()) {
+                                hasFormed = true;
+                                canAccess |= multiblock.allowsStructuralGuiAccess(this);
+                            }
+                        }
                     }
+                    updateFormedMultiblock(hasFormed, canAccess);
                 }
-                updateFormedMultiblock(hasFormed, canAccess);
             }
         }
     }
@@ -109,7 +114,7 @@ public abstract class TileEntityStructuralMultiblock extends TileEntityMekanism 
     @Nullable
     private MultiblockData getMultiblockData(Structure structure) {
         //Like the getMultiblockData(MultiblockManager) method except can assume the structure is indeed in our structures map,
-        // so we can slightly short circuit lookup
+        // so we can slightly short circuit lookup, and return null if there is no data
         MultiblockData data = structure.getMultiblockData();
         if (data != null && data.isFormed()) {
             return data;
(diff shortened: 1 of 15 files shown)


src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
```diff
@@ -60,88 +98,61 @@ public boolean hasStructure(Structure structure) {
...
-                        return master.onActivate(player, hand, stack);
+        if (!structuralGuiAccessAllowed()) {
+            //If we don't have any structures that allow gui access, just short circuit and pass
+            return InteractionResult.PASS;
+        }
+        InteractionResult result = InteractionResult.PASS;
+        for (Structure structure : structures.values()) {
+            //If we already have an interaction that has been handled with one of our multiblocks just pass
+            // in 99% of cases we will not have a second iteration of the loop
+            if (result == InteractionResult.PASS) {
```
thiakil: This would be better as an early return/break, mostly just because it would be easier to read

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
```diff
@@ -169,18 +180,23 @@ public void onNeighborChange(Block block, BlockPos neighborPos) {
...
+        InteractionResult interactionResult = InteractionResult.PASS;
+        for (Structure structure : structures.values()) {
+            if (interactionResult == InteractionResult.PASS && structure.getController() != null) {
+                MultiblockData multiblock = getMultiblockData(structure);
+                if (multiblock == null || !multiblock.isFormed()) {
+                    FormationResult result = structure.runUpdate(this);
                     if (!result.isFormed() && result.getResultText() != null) {
                         player.sendSystemMessage(result.getResultText());
-                        return InteractionResult.sidedSuccess(isRemote());
+                        interactionResult = InteractionResult.sidedSuccess(isRemote());
```
thiakil: I don't see a reason to assign it to a local?

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
```diff
@@ -14,25 +12,27 @@
...
 
-    private final Map<MultiblockManager<?>, Structure> structures = new HashMap<>();
+    //Note: We never expect this to actually be filled, but we set the default to six in case it somehow is part of a structure in every direction
+    // given our expected is so low we use an array map
+    private final Map<MultiblockManager<?>, Structure> structures = new Reference2ObjectArrayMap<>(6);
     private final Structure invalidStructure = Structure.INVALID;
     private final MultiblockData defaultMultiblock = new MultiblockData(this);
-
-    private String clientActiveMultiblock = null;
+    private boolean removing;
```
thiakil, line 37: should have javadoc explaining, and maybe a better name

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
```diff
@@ -14,25 +12,27 @@
...
 import net.minecraft.world.level.block.state.BlockState;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 public abstract class TileEntityStructuralMultiblock extends TileEntityMekanism implements IStructuralMultiblock, IConfigurable {
 
-    private final Map<MultiblockManager<?>, Structure> structures = new HashMap<>();
+    //Note: We never expect this to actually be filled, but we set the default to six in case it somehow is part of a structure in every direction
+    // given our expected is so low we use an array map
+    private final Map<MultiblockManager<?>, Structure> structures = new Reference2ObjectArrayMap<>(6);
```
thiakil: Won't this break down if the block is part of two of the same type? e.g. 2 boilers
managers are singletons as far as I can see

the capacity should be smaller though, I'd go with 2 at most

pupnewfster: Does three make more sense given the fact it contains `null` and an actual formed structure?

pupnewfster: Also the way that multiblock formation currently works is if there are two frames that are remotely close to touching of the same type, then they get combined into a single multiblock and won't form/ So it can't be part of two of the same type. (granted there aren't cases we can't have it be part of two multiblocks currently regardless)

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
```diff
@@ -48,6 +48,44 @@ public void setStructure(MultiblockManager<?> manager, Structure structure) {
         structures.put(manager, structure);
     }
 
+    @Override
+    public void removeStructure(Structure structure) {
```
thiakil: shouldnt this remove the map entry?

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
thiakil: why is there a null entry? even so, that's only 2 cases. if there's ever a third it will resize

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
pupnewfster: I am not sure, my initial implementation before https://github.com/mekanism/Mekanism/pull/8051/commits/707069f3b1ffb988bfb54f88e36299a1f792ca09 had it removing it, but then when I thought about it more, realistically there are some cases where a multiblock forms and unforms, where I am not sure the structure becomes invalid. Like when revalidation is occurring (I think), so I wanted it to not have to remove the structure if it was still valid, and then recalculate it, compared to just be able to re-use it and only remove if necessary. Maybe this method would be better named `multiblockUnformed`?

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
pupnewfster: You would have to ask aidan. I believe it is for when structural multiblocks are connected together but are not part of a multiblock type yet. I semi tried to remove it as part of this PR and it was causing some weird edge cases where multiblock contents were getting voided...

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
pupnewfster: Similar reason to the one for gui interaction, but I will apply that comment to both places where it just does an early return.

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
Speiger: @pupnewfster please don't use ArrayMaps from fastUtil.
they are simply arraylist implementations that do a simply indexOf search for finding the key.
They are horrible in performance and access times while the iteration is faster technically, it is not that much faster then a linkedMap which is extremely similar to the datastructure like arrayMaps are.

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
```diff
@@ -14,25 +12,30 @@
...
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.state.BlockState;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 public abstract class TileEntityStructuralMultiblock extends TileEntityMekanism implements IStructuralMultiblock, IConfigurable {
 
-    private final Map<MultiblockManager<?>, Structure> structures = new HashMap<>();
+    //Note: We never expect this to actually be filled, but we set the default to six in case it somehow is part of a structure in every direction
```
thiakil: nitpick: comment still refers to 6

thiakil (approved):

src/main/java/mekanism/common/tile/prefab/TileEntityStructuralMultiblock.java
pupnewfster: First off, we don't really care about the iteration order, but secondly for tiny maps array maps do have better performance than linked maps.

Issue #6617, opened 2020-10-13
Gas tube / Radioactive waste distribution


MTitzer:
*Please use the search functionality before reporting an issue. Also take a look at the closed issues!*

#### Issue description:
Distribution of either nuclear waste (spent or not) unevenly distributes into mass amount of radioactive waste barrels. Output from either a fission reactor or PRC outputting spent nuclear waste distributes in large increments (1000 mbs from PRC or the set burn rate from fission reactor). This means that in a setup of approximately 1900 barrels, under 10 barrels holds most of the waste.

This prevents any sort of large scale build with the goal of quickly decaying nuclear materials as barrels seem to be randomly selected and filled with the output of the source machine at the output rate (fission at 32 burn rate or PRC at 1000 mb output)

#### Steps to reproduce:
1. Create a large radioactive waste dumping ground (mine is within 2 chunks - chunkloaded with teleporter w/ anchor)
2. Connect all barrels and run tube line 5 or so chunks away
3. Output waste from either fission reactor (@ 32 burn rate) or PRC creating polonium pellets)

#### Version (make sure you are on the latest version before reporting):
**Forge:** 32.0.108
**Mekanism:** 10.0.9.432
**Other relevant version:**
Running Valhelsia modpack 1.16.1 - version 3.0.16

### Screenshots of issue/build:

[https://imgur.com/a/xmBm2L7](https://imgur.com/a/xmBm2L7)

MTitzer:
After a few test cases, it seems that the issue is present when the barrels connected are greater than 1000 or so. I removed a tube and cut the network from 1900 barrels to 951 and it began to distribute waste evenly. Once the tube was replaced, then the issue came back.

pupnewfster:
Ya not quite sure when I will implement the fix as it is relatively complicated logic wise, but it has to do with that when it is in "send remainder" mode it just tries to send as much as possible of the remainder. The issue is when it is all remainder it has zero as the initial amount to try and send evenly distributed. I think I can at least slightly improve it by trying and making it send one more than the value if there is a remainder until the remainder is gone or we are onto the final pass where it tries to send all remaining to the various destinations.

Shados:
[...]

PR #8062, opened 2024-04-16
Improve remainder distribution logic

pupnewfster:

Improve remainder distribution to try and split the remainder as evenly as possible between the various destinations before falling back to sending to the first one it will fit in (#6617)

I also added some shortcuts to avoid attempting to send things and exit early if values hit zero.

### DIFF
--- a/src/main/java/mekanism/common/lib/distribution/SplitInfo.java
+++ b/src/main/java/mekanism/common/lib/distribution/SplitInfo.java
@@ -2,23 +2,63 @@ package mekanism.common.lib.distribution;
 
 public abstract class SplitInfo<TYPE extends Number & Comparable<TYPE>> {
 
+    /**
+     * Number of targets to split the contents among.
+     */
     protected int toSplitAmong;
+    /**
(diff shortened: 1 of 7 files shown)


src/main/java/mekanism/common/lib/distribution/SplitInfo.java
```diff
@@ -15,5 +16,9 @@ protected SplitInfo(int totalTargets) {
 
     public abstract TYPE getRemainderAmount();
```
thiakil, line 46: can we get some javadocs on these methods? I have zero idea how it all works without trying to trace it in an IDE

thiakil:
If we're being strict with tests, I'd expect to see one that verifies the _amount_ sent to each endpoint (i.e. that the remainder was indeed split evenly)
Unless I've missed something, it's only testing that the full amount was sent 'somewhere'

pupnewfster:
I will add verification about how much it sends. The main reason I didn't is that one of the destinations can only accept a part of it, so figured if all of it got accepted that technically means it got distributed as expected. But you are right, it doesn't hurt to be more explicit.

thiakil (approved):

PR #8067, opened 2024-04-30
[1.19.2 Bug Fix] Fixes the rendering of Structural Glasses

Sam2451:
A small fix to skip the rendering of Structural Glasses adiacent to Basic Multiblock Blocks (including the traslucent Laser Focus Matrix), making it seamless.


thiakil:
we no longer support 1.19, nor am i sure this is the correct solution

PR #8079, opened 2024-05-17
Fix two recipe viewer issues

Bommels05:
Fixes one issue in the Emi integration and another one in the Jei integration:

1. Fixes wrong tag names in Jei tooltip for chemicals. Before it would iterate over all tags and check if the size of the tag was right and then check if the tag contained all elements of itself instead of all elements of the searched tag.
2. Fixes tank widgets in Emi with tag ingredients that have an amount higher than 1 not being filled completely by using the amount of the ingredient instead of the stacks.

Both problems can be observed when looking at recipes with water vapor in the Chemical Injection Chamber.


pupnewfster:
Thanks!

PR #8123, opened 2024-06-18
Add AE2 facade integration

Bommels05:
Adds an integration with AE2 facades. The facades can now be placed on all transmitters and transmitters can connect trough facades just like in AE. They can be removed by shift right clicking with the configurator in wrench mode. 

Just the blocks of the facades are saved in the NBT data of the transmitter and because it wasn't really possible to completely move rendering with AE into separate classes there are some places were `Object` is used instead of the actual type so everything still works without it.

I know performance of the transmitters is pretty critical so I hope performance is still good, especially when no facades are used.
I also removed the `datagenMainCompileOnly` dependency because it looked like it is also covered by the implementation one, but if you set it to `compileOnly` you will probably have to add a `datagenMainRuntimeOnly` back.

### DIFF
--- a/src/main/java/mekanism/client/render/obj/TransmitterBakedModel.java
+++ b/src/main/java/mekanism/client/render/obj/TransmitterBakedModel.java
@@ -1,5 +1,9 @@
 package mekanism.client.render.obj;
 
+import appeng.block.networking.CableBusBlock;
+import appeng.client.render.cablebus.CableBusRenderState;
+import appeng.client.render.cablebus.FacadeBuilder;
+import appeng.client.render.cablebus.FacadeRenderState;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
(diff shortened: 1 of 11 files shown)


src/main/java/mekanism/client/render/obj/TransmitterBakedModel.java
```diff
@@ -1,5 +1,9 @@
 package mekanism.client.render.obj;
 
+import appeng.block.networking.CableBusBlock;
+import appeng.client.render.cablebus.CableBusRenderState;
+import appeng.client.render.cablebus.FacadeBuilder;
+import appeng.client.render.cablebus.FacadeRenderState;
```
Technici4n, line 6: Only the `appeng.api` package is part of the public API. Everything else can change at any time.

Bommels05: Yeah... I originally wanted to add this in a separate mod via mixins, but I thought it would be cool to have this native in Mekanism. If using the internals won't be accepted I could maybe change the PR to not use AE2, but then the facades would probably just be like a full block and not for individual sides.

thiakil:
while this is a neat idea, I would rather have them natively in mekanism rather than depend on another mod. (especially with the non-api usage)

I'll be adding it natively soon-ish, but working on another item first

Bommels05:
> while this is a neat idea, I would rather have them natively in mekanism rather than depend on another mod. (especially with the non-api usage)
> 
> I'll be adding it natively soon-ish, but working on another item first

Ok. I'll try to get it working without AE2 myself, but my rendering skills aren't really good so if I don't get it to work I will just close this PR and wait for you to implement it.

PR #8125, opened 2024-06-21
ModDevGradle

shartte:
You know what this is :-)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,9 +9,8 @@ import mekanism.OptimizePng
 import mekanism.OutputChangelog
 import net.darkhax.curseforgegradle.TaskPublishCurseForge
 import net.darkhax.curseforgegradle.UploadArtifact
-import org.gradle.api.tasks.testing.Test
-import java.util.function.Consumer
 
+import java.util.function.Consumer
 
 plugins {
     id('org.ajoberstar.grgit.service')//Version declared in buildSrc
@@ -21,8 +20,7 @@ plugins {
     id('eclipse')
     id('idea')
     id('maven-publish')
-    //https://projects.neoforged.net/neoforged/moddevgradle
-    id('net.neoforged.moddev') version('2.0.75')
+    id('net.neoforged.moddev') version('2.0.75')//https://projects.neoforged.net/neoforged/moddevgradle
 }
 
 tasks.named('wrapper', Wrapper) {
@@ -171,10 +169,6 @@ static void extendConfigurations(Configuration base, Configuration... configurat
     }
 }
 
-Provider<SourceSet> getExtraSourceSet(String base, String name) {
-    return project.sourceSets.named(base).flatMap(sourceSet -> getExtraSourceSet(sourceSet, name))
-}
-
 Provider<SourceSet> getExtraSourceSet(SourceSet base, String name) {
     return project.sourceSets.named(base.getTaskName(name, null))
 }
@@ -198,8 +192,8 @@ def setupTasks(SourceSet sourceSet) {
     tasks.named(sourceSet.processResourcesTaskName, ProcessResources) {
         setGroup('process resources')
         duplicatesStrategy = DuplicatesStrategy.FAIL
-        var versionProperties = ['file': ['jarVersion': mod_version], 'mc_version': minecraft_version_range, 'forge_version': forge_version_range, 'loader_version': loader_version_range,
-                                 'jei_version': jei_version_range]
+        var versionProperties = ['file'       : ['jarVersion': mod_version], 'mc_version': minecraft_version_range, 'forge_version': forge_version_range,
+                                 'loader_version': loader_version_range, 'jei_version': jei_version_range]
         //Mark the properties as inputs so that when they change things update
         inputs.properties(versionProperties)
         filesMatching('META-INF/neoforge.mods.toml') { expand(versionProperties) }
@@ -236,10 +230,6 @@ java {
     withSourcesJar()
 }
 
-final getDatagenSourceSet(codeSourceSet) {
-    return sourceSets.named("datagen" + codeSourceSet.name.capitalize())
-}
-
 neoForge {
     // This enables modding tasks and configurations
     enable {
@@ -249,25 +239,25 @@ neoForge {
 
     mods {
         mekanism {
-            sourceSet sourceSets.api
-            sourceSet sourceSets.main
+            sourceSet((SourceSet) sourceSets.api)
+            sourceSet((SourceSet) sourceSets.main)
         }
         // contains main mod + its datagen code
         mekanismData {
             modSourceSets = mods.mekanism.modSourceSets
-            sourceSet sourceSets.datagenMain
+            sourceSet((SourceSet) sourceSets.datagenMain)
         }
         mekanismtests {
-            sourceSet sourceSets.gameTest
+            sourceSet((SourceSet) sourceSets.gameTest)
         }
     }
     for (String name : secondaryModules) {
         final sourceSet = sourceSets.named(name)
         mods.maybeCreate(name).modSourceSets.add(sourceSet)
         // Create a secondary mod group for the modules main code + its datagen code
-        final datagenMod = mods.maybeCreate(name + "Data")
+        final datagenMod = mods.maybeCreate(name + 'Data')
         datagenMod.modSourceSets.add(sourceSet)
-        datagenMod.modSourceSets.add(getDatagenSourceSet(sourceSet))
+        datagenMod.modSourceSets.add(sourceSet.flatMap(s -> getExtraSourceSet(s, 'datagen')))
     }
 
     accessTransformers = [
@@ -298,11 +288,13 @@ neoForge {
             // Uncomment this to get verbose debug logging
             // logLevel = org.slf4j.event.Level.DEBUG
         }
-        //Note: To enable logging into the client account, set the enable_devlogin property to true in your gradle user home
-        // You can also run gradlew runClient -Penable_devlogin=true
+        //Note: To enable logging into the client account, set the mc_devlogin property to true in your gradle user home
+        // You can also run gradlew runClient -Pmc_devlogin=true
         client {
             client()
-            devLogin = Boolean.valueOf(findProperty("enable_devlogin") ?: "false")
+            if (findProperty('mc_devlogin') ?: false) {
+                devLogin = true
+            }
         }
         clientAlt {
             client()
@@ -314,27 +306,30 @@ neoForge {
             server()
         }
         gameTestServer {
-            type = "gameTestServer"
+            type = 'gameTestServer'
             sourceSet = sourceSets.gameTest
             loadedMods.add(mods.mekanismtests)
         }
         gameTestClient {
             client()
+            if (findProperty('mc_devlogin') ?: false) {
+                devLogin = true
+            }
             sourceSet = sourceSets.gameTest
             loadedMods.add(mods.mekanismtests)
-            programArguments.addAll("--quickPlaySingleplayer", "GameTestWorld")
+            programArguments.addAll('--quickPlaySingleplayer', 'GameTestWorld')
         }
         data {
             data()
             programArguments.addAll('--all', '--output', file('src/datagen/generated/').absolutePath,
-                                    '--mod', 'mekanism', '--existing', file('src/main/resources/').absolutePath)
+                    '--mod', 'mekanism', '--existing', file('src/main/resources/').absolutePath)
 
             sourceSet = sourceSets.datagenMain
             loadedMods = [mods.mekanismData]
 
             for (String name : secondaryModules) {
-                println("Adding $name")
-                loadedMods.add(mods.named(name + "Data"))
+                //println("Adding $name")
+                loadedMods.add(mods.named(name + 'Data'))
                 programArguments.addAll('--mod', 'mekanism' + name, '--existing', file("src/${name}/resources/").absolutePath)
             }
         }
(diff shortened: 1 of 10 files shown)


build.gradle
```diff
@@ -234,76 +235,99 @@ java {
...
+            mods.add(neoForge.mods.named("mekanism"))
+            for (String name : secondaryModules) {
+                mods.add(neoForge.mods.named(name))
+            }
 
-        dependencies {
-            runtime(configurations.datagenNonMod)
+            //if the selected toolchain is a JBR, enable DCEVM
+            if (javaToolchains.launcherFor(java.toolchain).map { it.metadata.vendor }.getOrElse('').contains('JetBrains')) {
+                // TODO MDG jvmArguments.add('-XX:+AllowEnhancedClassRedefinition')
```
thiakil: 🤔

Spinoscythe: this is possible now.
you just do `jvmArgument('-XX:+AllowEnhancedClassRedefinition')`

Spinoscythe (changes requested):

build.gradle
shartte: I have to revert this one, for some reason the IDE Sync for me used the JetBrains JVM (causing the JVM arg to be added), while running the actual game used another JVM causing it to fail :)

thiakil: uhh, i thought jvms were supposed to ignore unknown XX params :/

pupnewfster: I think that is only the case with `-XX:+IgnoreUnrecognizedVMOptions`

shartte: Yeah I added `-XX:+IgnoreUnrecognizedVMOptions` which makes that work with this weird Gradle toolchain too

gradle.properties
```diff
@@ -19,14 +18,16 @@ minecraft_version_range=[1.21.1]
...
 junit_version=5.11.0
 jqwik_version=1.9.0
 
-#NeoGradle Settings
-neogradle.subsystems.parchment.minecraftVersion=1.21.1
-neogradle.subsystems.parchment.mappingsVersion=2024.11.17
-neogradle.subsystems.conventions.sourcesets.enabled=false
+#Parchment Settings
+neoForge.parchment.minecraftVersion=1.21
+neoForge.parchment.mappingsVersion=2024.06.23
```
pupnewfster: This is downgrading the parchment version

gradle.properties
```diff
@@ -19,14 +18,16 @@ minecraft_version_range=[1.21.1]
 # options are: alpha, beta, release
 release_type=alpha
 
+# Enable or disable the use of dev logon for the client run
+enable_devlogin=false
```
pupnewfster: This shouldn't be here, we should instead have a comment saying to place this property in the user's gradle properties file.

build.gradle
```diff
@@ -422,6 +443,11 @@ dependencies {
     //datagenMainRuntimeOnly("curse.maven:projecte-226410:${projecte_id}")
 
     datagenNonMod("com.thiakil:yaml-ops:${yamlops_version}")
+
+    // Enable dev login in clientAlt run
+    if (enable_devlogin) {
+        runtimeOnly "net.covers1624:DevLogin:0.1.0.4"
+    }
```
pupnewfster: I believe this is no longer necessary?

settings.gradle
```diff
@@ -1,15 +1,13 @@
...
 pluginManagement {
-    repositories {
-        gradlePluginPortal()
-        maven {
-            name = "NeoForge"
-            url = "https://maven.neoforged.net/releases/"
-        }
+    plugins {
+        //https://projects.neoforged.net/neoforged/moddevgradle
+        id('net.neoforged.moddev') version('2.0.75')
```
pupnewfster: Why when porting projecte did you define this in the normal plugins block, but here you are doing it in the settings gradle file?

build.gradle
```diff
@@ -197,7 +198,7 @@ def setupTasks(SourceSet sourceSet) {
     tasks.named(sourceSet.processResourcesTaskName, ProcessResources) {
         setGroup('process resources')
         duplicatesStrategy = DuplicatesStrategy.FAIL
-        var versionProperties = ['version': mod_version, 'mc_version': minecraft_version_range, 'forge_version': forge_version_range, 'loader_version': loader_version_range,
+        var versionProperties = ['file': ['jarVersion': mod_version], 'mc_version': minecraft_version_range, 'forge_version': forge_version_range, 'loader_version': loader_version_range,
```
pupnewfster: Is there a reason for changing how this defines the version? I am not strongly opposed to it, I just want to understand it.

pupnewfster (changes requested):

pupnewfster:
Does the generated mc sources end up in build or do they end up elsewhere? If they end up in build we need to replace the codeql workflow's filtering that is done [here](https://github.com/mekanism/Mekanism/blob/1.21.x/.github/workflows/codeql.yml#L64-L65), and if not we can probably revert [this](https://github.com/mekanism/Mekanism/commit/aa2fcfdff2791a1df4871d575497eee80822a8ac#diff-12783128521e452af0cfac94b99b8d250413c516ec71fe6d97dbea666ff7ba27R51-R65) from that workflow.

We also need to update the gradle-publish workflow to either have the MDG outputs cached if they are in a different directory or if they are just part of the overall gradle cache then remove [this step](https://github.com/mekanism/Mekanism/blob/1.21.x/.github/workflows/gradle-publish.yml#L45-L51) from the workflow.

build.gradle
```diff
@@ -235,76 +236,99 @@ java {
...
+        version = forge_version
+        enabledSourceSets = sourceSets // All source sets use Minecraft code
+    }
 
-        //if the selected toolchain is a JBR, enable DCEVM
-        if (javaToolchains.launcherFor(java.toolchain).map { it.metadata.vendor }.getOrElse('').contains('JetBrains')) {
-            jvmArguments.add('-XX:+AllowEnhancedClassRedefinition')
+    mods {
+        mekanism {
+            modSourceSets = [sourceSets.api, sourceSets.main, sourceSets.datagenMain]
```
pupnewfster: If I understand this properly, this should not include datagenMain and instead should be as a separate mod similar to what you did for projecte. Given we don't want the datagen code applying for normal runs in case it causes things to break, or if we want to run the client before we finished porting the datagen stuff to the new MC version.

build.gradle
```diff
@@ -83,6 +82,7 @@ configurations {
     extendConfigurations(compileOnly, apiCompileOnly, testCompileOnly)
     extendConfigurations(runtimeOnly, apiRuntimeOnly)
     extendConfigurations(localRuntime, apiLocalRuntime)
+    gameTestRuntimeClasspath.extendsFrom runtimeClasspath
```
pupnewfster: What is this used for?

build.gradle
```diff
@@ -155,11 +154,13 @@ SourceSet setupExtraSourceSet(SourceSet baseSourceSet, String extra) {
...
     extraSourceSet.compileClasspath += project.sourceSets.api.output
     extraSourceSet.compileClasspath += project.sourceSets.main.output
+    extraSourceSet.runtimeClasspath += project.sourceSets.main.runtimeClasspath
     if (baseSourceSet != project.sourceSets.main) {
         //If the base sourceSet is main it already is the extra source set and has a reference to the base one from before this if statement
         extraSourceSet.compileClasspath += getExtraSourceSet(project.sourceSets.main, extra).get().output
         extraSourceSet.compileClasspath += baseSourceSet.output
-        extraSourceSet.runs.modIdentifier = name
+        extraSourceSet.runtimeClasspath += baseSourceSet.runtimeClasspath
+        neoForge.mods.maybeCreate(name).modSourceSets.add(extraSourceSet)
```
pupnewfster: I believe this will end up causing the datagen sourcesets to be with the corresponding module's mod, when it should only load for datagen instead of for normal runs as well.

build.gradle
```diff
@@ -341,14 +365,11 @@ repositories { RepositoryHandler handler ->
     exclusiveRepo(handler, 'https://maven.parchmentmc.org/', 'org.parchmentmc.data')
 }
 
-tasks.named('test') {
-    //Disable builtin test task as we use and build uses testJunit so there is no point in having it also attempt to run an empty test task
-    enabled = false
+tasks.named('test').configure {
+    useJUnitPlatform()
```
pupnewfster: Please add `org.gradle.api.tasks.testing.Test` as an import and then change this to:

```suggestion
tasks.named('test', Test) {
    useJUnitPlatform()
```

So that it can resolve the type of the task being configured for purposes of checking the method exists.

build.gradle
```diff
@@ -235,76 +236,99 @@ java {
...
         }
     }
+
+    unitTest {
+        enable()
+        testedMod = mods.mekanism
+    }
 }
 
+configurations.dataAdditionalRuntimeClasspath.extendsFrom configurations.datagenNonMod
```
pupnewfster, line 357: Why is this down here instead of the normal configurations block, and what is the purpose of this?

build.gradle
```diff
@@ -235,76 +236,99 @@ java {
...
+                println("Adding $name")
+                loadedMods.add(mods.named(name))
+                programArguments.addAll('--mod', 'mekanism' + name, '--existing', file("src/${name}/resources/").absolutePath)
+            }
         }
     }
+
+    unitTest {
+        enable()
+        testedMod = mods.mekanism
```
pupnewfster, line 349: Do we need to adjust the loadedMods list?

build.gradle
```diff
@@ -235,76 +236,99 @@ java {
...
+        }
+        server {
+            server()
+        }
+        gameTestServer {
+            type = "gameTestServer"
+            sourceSet = sourceSets.gameTest
+            loadedMods = neoForge.mods
+        }
+        gameTestClient {
```
pupnewfster, line 320: You can add the argument you mentioned to me to this:
`programArguments.addAll("--quickPlaySingleplayer", "GameTestWorld")`

pupnewfster (changes requested):

gradle.properties
shartte: The user gradle properties override this, this is just the default value.
(And the comment should indicate that, yes)

gradle.properties
shartte: Fixed

gradle.properties
shartte: I decided to still just remove it from gradle.properties and use findProperty instead

build.gradle
shartte: Yes, removed

settings.gradle
shartte: This was more preference than anything, I usually collect my plugin versions in pluginManagement. 
But: no real reason. So moved to project

build.gradle
shartte: Yes, doing it this way makes it so hot reload works properly with IntelliJ and/or Eclipse
The reason behind it is this: the IDE will copy over the resource file without expanding any template variables
`${file.jarVersion}` will continue to work in this scenario, and all the dependency versions will ultimately be ignored by FML if they have unresolved placeholders as their content.

In the MDK we use a different approach, which ends up being more complicated (essentially code-generating the neoforge.mods.toml into a second source folder, which the IDE can then pick up)

An alternative approach is what AE2 used to do: Have the version as `0.0.0` in neoforge.mods.toml and just replace that with the real version when building the jar, but replacing `${file.jarVersion}` with the real version statically at build time (like I do here)  seems cleaner.

shartte:
> Does the generated mc sources end up in build or do they end up elsewhere? If they end up in build we need to replace the codeql workflow's filtering that is done [here](https://github.com/mekanism/Mekanism/blob/1.21.x/.github/workflows/codeql.yml#L64-L65), and if not we can probably revert [this](https://github.com/mekanism/Mekanism/commit/aa2fcfdff2791a1df4871d575497eee80822a8ac#diff-12783128521e452af0cfac94b99b8d250413c516ec71fe6d97dbea666ff7ba27R51-R65) from that workflow.
> 
> We also need to update the gradle-publish workflow to either have the MDG outputs cached if they are in a different directory or if they are just part of the overall gradle cache then remove [this step](https://github.com/mekanism/Mekanism/blob/1.21.x/.github/workflows/gradle-publish.yml#L45-L51) from the workflow.

MDG does not expand the sources on disk ever, but they are included in the Jar artifact build/moddev/artifacts/*.jar

shartte:
> We also need to update the gradle-publish workflow to either have the MDG outputs cached if they are in a different directory or if they are just part of the overall gradle cache then remove [this step](https://github.com/mekanism/Mekanism/blob/1.21.x/.github/workflows/gradle-publish.yml#L45-L51) from the workflow.

Yes, we need/can remove that. NFRT caches are just in the gradle home, which should be cached by setup-gradle already. There's another enhancement that can be done after a recent change in NFRT, which is caching Minecraft assets separately (i.e. by using a fully static cache-key across MC versions, since assets have large reuse between versions)

build.gradle
shartte: jup it did

build.gradle
shartte: Apparently nothing

build.gradle
shartte: I just copied the client mods

build.gradle
shartte: It is only available after enabling modding (that's why its here), and the reason it exists is to add non-mod jars to the legacy classpath so FML picks them up

build.gradle
shartte: Reworked all mods

pupnewfster (approved):

PR #8126, opened 2024-06-22
Move to Curios continuation

Spinoscythe:
TheIllusiveC4 has finally released 1.20.6 stable version for Curios, and it has entering maintenance mode. Unfortunately, it's also been said that they're abandoning curios, and that anyone else can update curios themselves for 21.0.

[Curios continuation](https://github.com/SSKirillSS/Curios) seems to be the most popular alternative.
Hence, this change seems right. The mod id and the packages are the same for the sake of backwards compatibility.

This has been discussed in the NeoForge discord with CodexAdrian, TG, me and many more

The idea of making a similar API in Neo has been proposed. But until then, this is vital for having curios for 21.0 for now

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -328,7 +328,7 @@ repositories { RepositoryHandler handler ->
         filter.includeGroup('org.openzen.zencode')
     })
     exclusiveRepo(handler, 'https://maven.terraformersmc.com/', 'dev.emi')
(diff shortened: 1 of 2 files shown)


pupnewfster:
Thanks for the PR, but I already have made this change locally in addition to updating a bunch of other dependencies

PR #8174, opened 2024-07-28
Ingredient suggestions 1

thiakil:
- Circuit tiers using the infuser directly, saving the ingot for the alloy and a small efficiency bonus on the infusion stack
- Allow copper as an alternate to iron for the infused alloy. Kinda want this to be a replacement though

Open to suggestions/changes


src/datagen/main/java/mekanism/common/recipe/impl/ControlCircuitRecipeProvider.java
```diff
@@ -44,4 +50,12 @@ private void addCircuitUpgradeRecipe(RecipeOutput consumer, IItemProvider output
               .key(Pattern.ALLOY, alloyTag)
               .build(consumer, Mekanism.rl(basePath + name));
     }
+
+    private void addCircuitInfusionUpgrade(RecipeOutput consumer, IItemProvider output, TagKey<Item> circuitTag, TagKey<InfuseType> infusionType, int singleAlloyAmount, String basePath, String name) {
+        ItemStackChemicalToItemStackRecipeBuilder.metallurgicInfusing(
+              IngredientCreatorAccess.item().from(circuitTag),
+              IngredientCreatorAccess.infusionStack().from(infusionType, (int) (singleAlloyAmount * 1.5F)),
```
pupnewfster: I am not really a fan of this alternate recipe skipping a base input AND then on top of that lowering the amount of infuse type needed. I think we should instead have it use a bit more of the infuse type (say 3x or at least 2.5x as much), so that both recipes have their own merits and provide users an actual choice to make. Especially as for say higher tier circuits then this is removing the intermediary tier alloy infusing costs.

pupnewfster (changes requested):
I am fine with the copper recipe (and agree we should remove the iron to infused alloy one), but am kind of against the circuit recipe adjustments in their current state.

src/datagen/main/java/mekanism/common/recipe/impl/ControlCircuitRecipeProvider.java
thiakil: I'm fine to increase the infusion amount, but the current system is more tedious than anything. Banging 6 ingots into a circuit doesn't make much sense, it's just a grind

pupnewfster: Yeah, which is why I am fine with adding an alternative way. I just feel like the alternative way should be more infuse type intensive than using the ingots. So both systems have their own merits and gives players a choice to make. I think 3x as much would be a good place to have it at.

thiakil:
Using ingots: 
image

using 3x infusion:
image

src/datagen/main/java/mekanism/common/recipe/impl/OreProcessingRecipeProvider.java
```diff
@@ -64,6 +64,11 @@ public void addRecipes(RecipeOutput consumer, HolderLookup.Provider registries)
               IngredientCreatorAccess.infusionStack().from(MekanismAPITags.InfuseTypes.CARBON, 10),
               MekanismItems.ENRICHED_IRON.getItemStack()
         ).build(consumer, Mekanism.rl(basePath + "iron/enriched"));
+        ItemStackChemicalToItemStackRecipeBuilder.metallurgicInfusing(
+              IngredientCreatorAccess.item().from(MekanismTags.Items.PROCESSED_RESOURCES.get(ResourceType.DUST, PrimaryResource.IRON)),
```
pupnewfster, line 68: Do we want to add a second recipe like this, or just make the first recipe use an an or (compound) ingredient? I am fine with either. If you do decide to use an or ingredient you can use one of our helpers: `BaseRecipeProvider.createIngredient`, which you can see how we do it in how we add the recipes that create carbon.

I sort of think it makes more sense for it to just be a single recipe. But I am leaving that up to you.

src/datagen/main/java/mekanism/common/recipe/impl/ControlCircuitRecipeProvider.java
```diff
@@ -34,6 +35,11 @@ public void addRecipes(RecipeOutput consumer, HolderLookup.Provider registries)
         addCircuitUpgradeRecipe(consumer, MekanismItems.ADVANCED_CONTROL_CIRCUIT, MekanismTags.Items.CIRCUITS_BASIC, MekanismTags.Items.ALLOYS_INFUSED, basePath, "advanced");
         addCircuitUpgradeRecipe(consumer, MekanismItems.ELITE_CONTROL_CIRCUIT, MekanismTags.Items.CIRCUITS_ADVANCED, MekanismTags.Items.ALLOYS_REINFORCED, basePath, "elite");
         addCircuitUpgradeRecipe(consumer, MekanismItems.ULTIMATE_CONTROL_CIRCUIT, MekanismTags.Items.CIRCUITS_ELITE, MekanismTags.Items.ALLOYS_ATOMIC, basePath, "ultimate");
+
+        //infusion variants that save on the base ingots needed and half an alloy
```
pupnewfster: Is this comment about "half an alloy" correct? I feel like it is no longer correct.

pupnewfster (approved):
Overall looks quite good. One minor thought is do we want to adjust in `MekanismEmiDefaults` which circuit recipe we declare as defaults? Personally I think leaving them as is probably makes the most sense, as users can switch them over to these new recipes if they want, but I figured I would bring it up here.

src/datagen/main/java/mekanism/common/recipe/impl/OreProcessingRecipeProvider.java
thiakil: Maybe we can merge it later if we remember. Separate for now would promote visibility?

src/datagen/main/java/mekanism/common/recipe/impl/ControlCircuitRecipeProvider.java
thiakil: Whoops

pupnewfster (changes requested):
MekanismLangProvider, please update the translation for `MekanismAdvancements.INFUSED_ALLOY`

src/datagen/main/java/mekanism/common/recipe/impl/OreProcessingRecipeProvider.java
pupnewfster: Up to you, though I don't think the visibility is that bad given the fact there is basically only ever one iron ingot in modded playthroughs, and then it just cycles between the two inputs.

pupnewfster: I think I am going to merge them for now. Visibility doesn't seem that bad to me for the above stated reason

pupnewfster: Actually scratch that, I didn't think about the fact that whil ethere is only one iron ingot, there are often multiple iron dusts.

pupnewfster (approved):

PR #8196, opened 2024-08-09
Chemical stack reunification

thiakil:
- Merges chemical types to single Chemical type
- Slight change to `IChemicalHandler` to allow for direct implementation on blocks alongside fluid handling (method clashes)
- Complete removal of Boxed types
- Adds Infusion recipes to Oxidiser

### backcompat loading
- [x] Dynamic tank
- [x] Turbine
- [x] Fusion
- [x] Fission
- [x] SPS
- [x] Chemical Dissolution - Tile fine, items with both tanks present will lose both due to shared indexes
- [x] Entangoloporter (first non-empty chemical tank wins)

### other
- [x] crafttweaker example scripts generating extra lines
- [x] testing CT
- [x] testing JSON Things
- [x] ensure redundant caps got removed
- [x] Tubes not rendering contents until transmitter added/removed
- [x] gasses not loading after saving world in new format
- [x] Rename gas upgrade to chemical upgrade, and potentially rename in code as well?
- [x] Remove excess old chemical colors from gui_objects color map
- [x] Implement propper support for the new `ItemStackChemicalToItemStackRecipe#perTickUsage` method. Currently is only used by things that are generic to the `IC2I` recipe for viewers and mappers, but can't be set on a per recipe basis and giving different values than the default doesn't change the actual behavior of the machines. (This method allows us to further deduplicate some of our code, and is also something that has been requested for some of our recipes that currently are per tick based)


src/main/java/mekanism/common/content/tank/TankMultiblockData.java
```diff
@@ -121,10 +112,8 @@ public void writeUpdateTag(CompoundTag tag, HolderLookup.Provider provider) {
     private float getScale() {
         return switch (mergedTank.getCurrentType()) {
             case FLUID -> MekanismUtils.getScale(prevScale, getFluidTank());
-            case GAS -> MekanismUtils.getScale(prevScale, getGasTank());
-            case INFUSION -> MekanismUtils.getScale(prevScale, getInfusionTank());
-            case PIGMENT -> MekanismUtils.getScale(prevScale, getPigmentTank());
-            case SLURRY -> MekanismUtils.getScale(prevScale, getSlurryTank());
+            case CHEMICAL -> MekanismUtils.getScale(prevScale, getChemicalTank());
+            //todo shouldn't this use the lowest amount? - Thiakil
```
pupnewfster, line 116: Unsure? My initial thinking was that in general we display the larger number for chemicals in the GUI when the tank is empty. Though technically based on the implementation details of MekanismUtils#getScale, because we pass zero as the stored amount, the passed capacity is effectively ignored.

thiakil: > My initial thinking was that in general we display the larger number for chemicals in the GUI when the tank is empty

but why? its misleading - though I think the capacity should be the same or at least very similar, it causes confusion when they get _less_ than they bargained for

pupnewfster (approved):

PR #8209, opened 2024-08-21
Add support for chemicals as camos on framed blocks

XFactHD:

Add FramedBlocks integration to support chemicals as camos on framed blocks.

A few noteworthy points:
- I've left a TODO about light emission as this is currently not exposed by the chemical itself and instead only available at the `IChemicalConstant` level
- ~~I've left a TODO about `MapColor` as the chemicals of course don't have any, considering they are never placed in-world. Returning null will fall back to the framed block's map color. It might be interesting to store a mapping of chemical->map color in the integration which is automatically computed from the chemical's color representation~~ Addressed
- Adding and removing a chemical camo currently consumes and returns respectively the same 1000mB as the fluid camo does
- ~~Stepping/falling on, hitting and breaking a block with a chemical camo uses the stone sound type for now. Fluid camos have the same issue and I have not found a satisfactory solution for this yet~~ Addressed

### DIFF
--- a/src/api/java/mekanism/api/MekanismAPITags.java
+++ b/src/api/java/mekanism/api/MekanismAPITags.java
@@ -88,6 +88,8 @@ public class MekanismAPITags {
         public static final TagKey<Chemical> CLEAN = tag("clean");
         /**
          * Chemicals in this tag cannot be inserted into framed blocks
+         *
+         * @since 10.7.1
          */
         public static final TagKey<Chemical> FRAMEDBLOCKS_BLACKLISTED = tag("framedblocks_blacklisted");
 
--- a/src/api/java/mekanism/api/MekanismAPITags.java
+++ b/src/api/java/mekanism/api/MekanismAPITags.java
@@ -86,6 +86,10 @@ public class MekanismAPITags {
          * Represents all clean slurries.
          */
         public static final TagKey<Chemical> CLEAN = tag("clean");
+        /**
+         * Chemicals in this tag cannot be inserted into framed blocks
+         */
+        public static final TagKey<Chemical> FRAMEDBLOCKS_BLACKLISTED = tag("framedblocks_blacklisted");
 
         private static TagKey<Chemical> tag(String name) {
             return TagKey.create(MekanismAPI.CHEMICAL_REGISTRY_NAME, rl(name));
--- a/src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
+++ b/src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
@@ -1,0 +1,143 @@
+package mekanism.common.integration.framedblocks;
+
+import com.mojang.serialization.MapCodec;
+import mekanism.api.Action;
+import mekanism.api.MekanismAPI;
+import mekanism.api.chemical.Chemical;
+import mekanism.api.chemical.ChemicalStack;
+import mekanism.api.chemical.IChemicalHandler;
+import mekanism.api.text.TextComponentUtil;
+import mekanism.common.MekanismLang;
+import mekanism.common.capabilities.Capabilities;
+import net.minecraft.core.BlockPos;
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.network.RegistryFriendlyByteBuf;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.codec.StreamCodec;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.Level;
+import org.jetbrains.annotations.Nullable;
+import xfacthd.framedblocks.api.camo.CamoContainerFactory;
+import xfacthd.framedblocks.api.camo.TriggerRegistrar;
+import xfacthd.framedblocks.api.util.CamoMessageVerbosity;
+import xfacthd.framedblocks.api.util.ConfigView;
+
+final class ChemicalCamoContainerFactory extends CamoContainerFactory<ChemicalCamoContainer> {
+
+    private static final MapCodec<ChemicalCamoContainer> CODEC = Chemical.CODEC
+            .xmap(ChemicalCamoContainer::new, ChemicalCamoContainer::getChemical)
+            .fieldOf("chemical");
+    private static final StreamCodec<RegistryFriendlyByteBuf, ChemicalCamoContainer> STREAM_CODEC =
+            Chemical.STREAM_CODEC.map(ChemicalCamoContainer::new, ChemicalCamoContainer::getChemical);
+    private static final Component MSG_RADIOACTIVE = TextComponentUtil.translate(
+            MekanismLang.FRAMEDBLOCKS_CAMO_RADIOACTIVE.getTranslationKey()
+    );
+
+    @Override
+    protected void writeToNetwork(CompoundTag tag, ChemicalCamoContainer camo) {
+        Chemical chemical = camo.getChemical();
+        tag.putInt("chemical", MekanismAPI.CHEMICAL_REGISTRY.getId(chemical));
+    }
+
+    @Override
+    protected ChemicalCamoContainer readFromNetwork(CompoundTag tag) {
+        Chemical chemical = MekanismAPI.CHEMICAL_REGISTRY.byId(tag.getInt("chemical"));
+        return new ChemicalCamoContainer(chemical);
+    }
+
+    @Override
+    @Nullable
+    public ChemicalCamoContainer applyCamo(Level level, BlockPos pos, Player player, ItemStack stack) {
+        IChemicalHandler handler = Capabilities.CHEMICAL.getCapability(stack);
+        if (handler == null || handler.getChemicalTanks() <= 0) {
+             return null;
+        }
+
+        ChemicalStack chemical = handler.getChemicalInTank(0);
+        if (!isValidChemical(chemical.getChemical(), player)) {
+            return null;
+        }
+
+        if (!player.isCreative() && ConfigView.Server.INSTANCE.shouldConsumeCamoItem()) {
+            ChemicalStack extracted = handler.extractChemical(0, FramedBlocksIntegration.Constants.CHEMICAL_AMOUNT, Action.SIMULATE);
+            if (extracted.getAmount() != FramedBlocksIntegration.Constants.CHEMICAL_AMOUNT) {
+                return null;
+            }
+
+            if (!level.isClientSide()) {
+                handler.extractChemical(0, FramedBlocksIntegration.Constants.CHEMICAL_AMOUNT, Action.EXECUTE);
+            }
+        }
+
+        return new ChemicalCamoContainer(chemical.getChemical());
+    }
+
+    @Override
+    public boolean removeCamo(Level level, BlockPos pos, Player player, ItemStack stack, ChemicalCamoContainer camo) {
+        if (stack.isEmpty()) {
+            return false;
+        }
+
+        IChemicalHandler handler = Capabilities.CHEMICAL.getCapability(stack);
+        if (handler == null || handler.getChemicalTanks() <= 0) {
+            return false;
+        }
+
+        ChemicalStack chemical = camo.getChemical().getStack(FramedBlocksIntegration.Constants.CHEMICAL_AMOUNT);
+        if (handler.insertChemical(chemical, Action.SIMULATE).isEmpty()) {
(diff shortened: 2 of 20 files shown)


src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContent.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    @Override
+    public boolean isEmissive() {
+        return false;
+    }
+
+    @Override
+    public SoundType getSoundType() {
+        return SoundType.STONE;
```
pupnewfster: Up to you, but wouldn't this make more sense as potentially `SoundType.WET_SPONGE`? (Or similar to fluid blocks just `SoundType.EMPTY`

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContent.java
```diff
@@ -0,0 +1,189 @@
...
+
+    @Override
+    public boolean canEntityDestroy(BlockGetter level, BlockPos pos, Entity entity) {
+        return true;
+    }
+
+    @Override
+    @Nullable
+    public MapColor getMapColor(BlockGetter level, BlockPos pos) {
+        // TODO: Chemicals don't provide a map color
```
pupnewfster: If you cache the value in the camo content so it doesn't have to be recalculated every time this method is called, you could make it be:
```java
MapColor mapColor = FluidDeferredRegister.getClosestColor(chemical.getColorRepresentation());
```

src/main/java/mekanism/common/integration/framedblocks/FramedBlocksIntegration.java
```diff
@@ -0,0 +1,83 @@
...
+import xfacthd.framedblocks.api.camo.CamoContainerFactory;
+import xfacthd.framedblocks.api.util.FramedConstants;
+
+public final class FramedBlocksIntegration {
+
+    private static final DeferredRegister<CamoContainerFactory<?>> CAMO_FACTORIES = DeferredRegister.create(
+            FramedConstants.CAMO_CONTAINER_FACTORY_REGISTRY_KEY,
+            Mekanism.MODID
+    );
+    private static final DeferredRegister<ParticleType<?>> PARTICLE_TYPES = DeferredRegister.create(
```
pupnewfster: Just for consistency within our codebase switch this to using a `ParticleTypeDeferredRegister` as we already have that as a subtype of deferred register.

src/main/java/mekanism/common/integration/framedblocks/FramedBlocksIntegration.java
```diff
@@ -0,0 +1,83 @@
...
+            ClientEvents.init(modBus);
+        }
+    }
+
+    public static final class Constants {
+
+        /**
+         * Chemicals tagged with this tag cannot be inserted into a framed blocks
+         */
+        public static final TagKey<Chemical> CHEMICAL_BLACKLISTED = TagKey.create(
```
pupnewfster: We should probably add this in `MekanismLangProvider#addTags` or EMI will complain in dev that it doesn't have a translation.

Also do you think it makes sense to expose this constant in `MekanismAPITags`?

src/main/java/mekanism/common/integration/framedblocks/ChemicalSpriteParticle.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    protected float getV1() {
+        return sprite.getV((vo + 1.0F) / 4.0F);
+    }
+
+    @Override
+    @SuppressWarnings("deprecation")
+    public int getLightColor(float partialTick) {
+        int light = level.hasChunkAt(pos) ? LevelRenderer.getLightColor(level, pos) : 0;
```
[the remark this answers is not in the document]
pupnewfster: To avoid the deprecation warning you can switch this to `WorldUtils.isChunkLoaded(level, pos)`, unless something about that implementation is doing something different than this

src/main/java/mekanism/common/integration/framedblocks/ChemicalSpriteParticle.java
```diff
@@ -0,0 +1,79 @@
...
+        this.gravity = 1F;
+        this.quadSize /= 2F;
+        this.uo = random.nextFloat() * 3F;
+        this.vo = random.nextFloat() * 3F;
+        this.brightness = 0;
+
+        int tint = chemical.getTint();
+        this.rCol = .6F * (float)(tint >> 16 & 0xFF) / 255F;
+        this.gCol = .6F * (float)(tint >>  8 & 0xFF) / 255F;
+        this.bCol = .6F * (float)(tint       & 0xFF) / 255F;
```
pupnewfster: Can you switch the (float) (tint....) parts of these to using `MekanismRenderer#getRed` etc?

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
```diff
@@ -0,0 +1,143 @@
...
+    public ChemicalCamoContainer applyCamo(Level level, BlockPos pos, Player player, ItemStack stack) {
+        IChemicalHandler handler = Capabilities.CHEMICAL.getCapability(stack);
+        if (handler == null || handler.getChemicalTanks() <= 0) {
+             return null;
+        }
+
+        ChemicalStack chemical = handler.getChemicalInTank(0);
+        if (!isValidChemical(chemical.getChemical(), player)) {
+            return null;
+        }
```
pupnewfster: Realistically we might want this to check all chemical tanks, and just take the first chemical tank that has a valid chemical in it.

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
```diff
@@ -0,0 +1,143 @@
...
+
+        IChemicalHandler handler = Capabilities.CHEMICAL.getCapability(stack);
+        if (handler == null || handler.getChemicalTanks() <= 0) {
+            return false;
+        }
+
+        ChemicalStack chemical = camo.getChemical().getStack(FramedBlocksIntegration.Constants.CHEMICAL_AMOUNT);
+        if (handler.insertChemical(chemical, Action.SIMULATE).isEmpty()) {
+            if (!level.isClientSide() && !player.isCreative() && ConfigView.Server.INSTANCE.shouldConsumeCamoItem()) {
```
pupnewfster: Shouldn't we check if the camo item has to be consumed before trying to insert the chemical? Because if it doesn't need to be then we can theoretically just return true?

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public boolean validateCamo(ChemicalCamoContainer camo) {
+        return isValidChemical(camo.getChemical(), null);
+    }
+
+    private static boolean isValidChemical(Chemical chemical, @Nullable Player player) {
+        if (chemical.isEmptyType()) {
+            return false;
+        }
+        if (chemical.isRadioactive()) {
```
pupnewfster: We probably want to use `Chemical#hasAttributesWithValidation` instead (with a slightly different validation message). That way if we later add any attributes other than radiation that need validation or an addon does, we can respect that here.

src/main/java/mekanism/common/integration/framedblocks/ChemicalModel.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Override
+    public List<BakedQuad> getQuads(@Nullable BlockState state, @Nullable Direction side, RandomSource random) {
+        return getQuads(state, side, random, ModelData.EMPTY, RenderType.translucent());
+    }
+
+    @Override
+    public List<BakedQuad> getQuads(@Nullable BlockState state, @Nullable Direction side, RandomSource rand, ModelData extraData, RenderType layer) {
+        return (side == null || layer != renderType) ? List.of() : quads.get(side);
```
pupnewfster: This should return `Collections.emptyList()` instead of `List.of()` as the backing implementation is more efficient in a variety of cases due to it short circuiting various methods instead of just basically being an immutable view of an array list.

pupnewfster (changes requested):

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
XFactHD: Sounds reasonable for the validation stage. Would you also want it to continue searching if the chemical is valid but the extraction fails or should that bail out?

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContent.java
XFactHD: That's a clever idea, I really should have gone through the vanilla sound types again. I ended up using `WET_GRASS` because it sounds a bit less harsh and more "splashy".

src/main/java/mekanism/common/integration/framedblocks/FramedBlocksIntegration.java
XFactHD: Exposing it in the API sounds reasonable to me, moved it there.

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
XFactHD: This is something I debated when I implemented fluid camos in FramedBlocks itself. The idea was that only a fluid handler (or chemical handler in this case) that could accept the given fluid/chemical would be usable as a removal tool (i.e. tank is empty or contains the same thing), regardless of whether anything would actually be returned to the player.
I'll make it check whether the tank considers the chemical valid and the existing chemical is either empty or the same one and only do the simulated insertion when the chemical would actually be returned to the tank. Let me know whether that's a better solution.

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
XFactHD: That's fair, do you have an idea for a sensible, more generic message?

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
pupnewfster: I think you can do the simulate check then. And check if it fully fit for when it requires the resource, and just if any fit for when it doesn’t. Or maybe even just if it has that type stored? Not really sure.

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
pupnewfster: Chemicals that require special handling cannot be put in a framed block

Or something along those lines

src/api/java/mekanism/api/MekanismAPITags.java
```diff
@@ -86,6 +86,10 @@ private Chemicals() {
          * Represents all clean slurries.
          */
         public static final TagKey<Chemical> CLEAN = tag("clean");
+        /**
+         * Chemicals in this tag cannot be inserted into framed blocks
+         */
```
pupnewfster, line 93: This should have a since 10.7.1 tag and please also edit the api version in MekanismAPI to be 10.7.1

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
XFactHD: The latter is basically what I implemented. It has the benefit of allowing you to remove the camo with a full tank, which would fail if I tested it with a simulated 1mB insertion.

src/main/java/mekanism/common/integration/framedblocks/ChemicalCamoContainerFactory.java
pupnewfster: Probably keep searching, because if there is say something in an input tank and output and the input is first, it may fail when the output would succeed

XFactHD: That's a good point, yeah. No additional change needed then, that's already what the latest code does.

pupnewfster (approved):
Looks great, thanks! I will try to test it tomorrow, and assuming I don't run into any bugs, and that it doesn't accidentally introduce a hard dep on framed blocks I will go ahead and merge it afterwards.

PR #8211, opened 2024-08-22
[1.12] Fix buggy OpenComputers integration.

mckuhei:
Fix this bug:
image


thiakil:
Please note that this will not be released as we're no longer developing 1.12, but the code looks fine

mckuhei:
Ok, thanks!

PR #8219, opened 2024-08-28
Auto eject round robin

thiakil:
Gives machines with a Transporter config the ability to do round robin auto-eject like a logistical sorter

### Todo
- [ ] Rejig the GUI to have both Strict Input and RR be the new on/off buttons

### DIFF
--- a/src/main/java/mekanism/common/content/network/transmitter/LogisticalTransporterBase.java
+++ b/src/main/java/mekanism/common/content/network/transmitter/LogisticalTransporterBase.java
@@ -26,6 +26,8 @@ import mekanism.common.network.PacketUtils;
 import mekanism.common.network.to_client.transmitter.PacketTransporterBatch;
 import mekanism.common.network.to_client.transmitter.PacketTransporterSync;
 import mekanism.common.tier.TransporterTier;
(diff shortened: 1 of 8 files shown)


src/main/java/mekanism/common/content/network/transmitter/LogisticalTransporterBase.java
```diff
@@ -403,6 +405,14 @@ public <BE extends BlockEntity & IAdvancedTransportEjector> TransitResponse inse
         return insert(outputter, outputterPos, request, color, doEmit, min);
     }
 
+    public TransitResponse insertMaybeRR(@NotNull TileEntityConfigurableMachine outputter, BlockPos outputterPos, TransitRequest request, @Nullable EnumColor color, boolean doEmit, int min) {
```
pupnewfster, line 408: Instead of duplicating methods like this, would it maybe make more sense to have `TileEntityConfigurableMachine` just implement `IAdvancedTransportEjector` and wrap the calls to the implementation on `TileComponentEjector`?

thiakil: Main reason I didn't is that it avoids the custom send home methods being needed

But that could be implemented

PR #8240, opened 2024-09-28
Blasting Highlight Event Visual Bug?

alegian:
This is my first time trying to contribute something to someone else's repo, please be kind.

## Problem
I was looking at the `RenderHighlightEvent` code for the AoE blasting, and I noticed that you are calling recursive  `RenderHighlightEvent`s for each block affected, so that they may have a chance to cancel themselves. 

However, when creating those recursive events, you pass the original `rayTraceResult` as parameter. I believe this means that instead of simulating one event for each block in the AoE, you actually simulate the origin-block's event many times over.

Create a seperate `BlockHitResult` for each `blastingTarget` blockPos. For its parameters, I essentially cloned the original `BlockHitResult`, changing only the blockPos. So for example if the original block was 'hit' at the top left, every AoE block is going to simulate a top left hit aswell.

### DIFF
--- a/src/main/java/mekanism/client/render/RenderTickHandler.java
+++ b/src/main/java/mekanism/client/render/RenderTickHandler.java
@@ -385,7 +385,9 @@ public class RenderTickHandler {
                         Lazy<VertexConsumer> lineConsumer = Lazy.of(() -> renderer.getBuffer(RenderType.lines()));
                         for (Entry<BlockPos, BlockState> block : blocks.entrySet()) {
                             BlockPos blastingTarget = block.getKey();
-                            if (!pos.equals(blastingTarget) && !ClientHooks.onDrawHighlight(levelRenderer, info, rayTraceResult, event.getDeltaTracker(), matrix, renderer)) {
+                            // simulate ray tracing results for all block positions
+                            var currResult = new BlockHitResult(rayTraceResult.getLocation(), rayTraceResult.getDirection(), blastingTarget, rayTraceResult.isInside());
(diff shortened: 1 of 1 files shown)


src/main/java/mekanism/client/render/RenderTickHandler.java
```diff
@@ -385,7 +385,9 @@ public void onBlockHover(RenderHighlightEvent.Block event) {
                         Lazy<VertexConsumer> lineConsumer = Lazy.of(() -> renderer.getBuffer(RenderType.lines()));
                         for (Entry<BlockPos, BlockState> block : blocks.entrySet()) {
                             BlockPos blastingTarget = block.getKey();
-                            if (!pos.equals(blastingTarget) && !ClientHooks.onDrawHighlight(levelRenderer, info, rayTraceResult, event.getDeltaTracker(), matrix, renderer)) {
+                            // simulate ray tracing results for all block positions
+                            BlockHitResult currResult = new BlockHitResult(rayTraceResult.getLocation(), rayTraceResult.getDirection(), blastingTarget, rayTraceResult.isInside());
```
pupnewfster: I believe this can be `rayTraceResult.withPosition(blastingTarget)`. We can also jjust inline it in the call to `ClientHooks#onDrawHighlight` so that we don't have to construct the new object for if the position is already the blasting target.

pupnewfster (changes requested):
Thanks, and I am sorry for taking a while to get to reviewing this. Only a minor cleanup related comment.

src/main/java/mekanism/client/render/RenderTickHandler.java
alegian: how did you know that helper exists 😮

pupnewfster (approved):

PR #8252, opened 2024-11-09
Adapt the "onAlloyIteraction" and "canUpgrade" methods to subclasses of "ITier"

lostmyself8:
1.When adding extra alloys, the "AlloyTier" accepted by "onAlloyIteraction" is too limited. It would be better to replace it with  a subclass of "ITier", at least I don't have to do repetitive work.
2.I modified the parameters in 'canUpgrade' to also adapt to subclasses of 'ITier', reducing unnecessary forced conversions.
3.Please forgive me for using a translator.
Thank you for everything you have done for mek.

### DIFF
--- a/src/api/java/mekanism/api/IAlloyInteraction.java
+++ b/src/api/java/mekanism/api/IAlloyInteraction.java
@@ -19,5 +19,5 @@ public interface IAlloyInteraction {
      * @param stack  - the stack of alloy being right-clicked
      * @param tier   - the tier of the alloy
      */
-    <TIER extends ITier> void onAlloyInteraction(Player player, ItemStack stack, @NotNull TIER tier);
+    void onAlloyInteraction(Player player, ItemStack stack, @NotNull ITier tier);
 }
--- a/src/api/java/mekanism/api/IAlloyInteraction.java
+++ b/src/api/java/mekanism/api/IAlloyInteraction.java
@@ -1,6 +1,6 @@
 package mekanism.api;
 
-import mekanism.api.tier.AlloyTier;
+import mekanism.api.tier.ITier;
 import net.minecraft.world.entity.player.Player;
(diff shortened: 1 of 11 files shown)


thiakil:
Please remove the generic, it's not needed. Just `ITier` should be fine

lostmyself8:
I changed them

pupnewfster:
I am a bit confused on the use case here. Is it to allow other items to be used as alloys as alternates for upgrading? Or is this for adding more tiers? Given with how the logic works it still ends up falling back to BaseTier which is also an enum so for more tiers that doesn't really help in terms of extending it.

lostmyself8:
I have added more tiers, which belong to the new enumeration class. If you want to upgrade the pipeline with alloys, you cannot use AlloyTier anymore. Due to the addition of four new alloys (perhaps more in the future), I cannot directly use AlloyTier to bypass the restrictions, as AlloyTier only defines three tiers.

thiakil:
the point is that https://github.com/mekanism/Mekanism/blob/38b95729a41b9667694808859fd7ff336af8396e/src/main/java/mekanism/common/tile/transmitter/TileEntityTransmitter.java#L398

Takes the `BaseTier`, which is also an enum. So your PR doesnt fix anything meaningful

lostmyself8:
I seem to have forgotten to modify this. Perhaps I need to add a special interface to BaseTier so that I can use this method through the interface.

Because I added an enum AdvancedTier similar to BaseTier, if they have a common interface, I don't need to write an upgradeResult and onExtraAlloyIteraction again.

thiakil (changes requested):
`BaseTier` is still a problem with this, as mentioned in previous comments

pupnewfster:
I believe I expanded things in a cleaner way that should handle your use case, please let me know if it does.

PR #8262, opened 2024-11-22
Add a warning for redstone conditions not being met

bukowski912:

- Make `GuiInsetElement` able to track warnings through a supplier
- Use a blended warning overlay for `GuiInsetElement` during a warning
- Track several activation warnings (based on control type) under `GuiRedstoneControlTab`
- Monitor redstone activation state of `TileEntityMekanism`


src/main/java/mekanism/client/gui/GuiMekanismTile.java
```diff
@@ -39,7 +41,8 @@ protected void addGenericTabs() {
             upgradeWindowTab = addRenderableWidget(new GuiUpgradeWindowTab(this, tile, () -> upgradeWindowTab));
         }
         if (tile.supportsRedstone()) {
-            addRenderableWidget(new GuiRedstoneControlTab(this, tile));
+            BooleanSupplier supplier = trackWarning(WarningType.REDSTONE_UNACTIVATED, () -> !tile.isRedstoneActivated());
+            addRenderableWidget(new GuiRedstoneControlTab(this, tile).warning(supplier));
```
pupnewfster: This should probably be adjusted to just be:
```java
addRenderableWidget(new GuiRedstoneControlTab(this, tile).warning(WarningType.REDSTONE_UNACTIVATED, () -> !tile.isRedstoneActivated()));
```
and have the `warning` method be implemented similar to the one in `GuiSlot`

src/main/java/mekanism/client/gui/element/GuiInsetElement.java
```diff
@@ -55,6 +70,17 @@ protected ResourceLocation getOverlay() {
         return overlay;
     }
 
+    @Override
+    public void draw(@NotNull GuiGraphics guiGraphics) {
```
pupnewfster: This can be protected

src/main/java/mekanism/client/gui/element/GuiInsetElement.java
```diff
@@ -55,6 +70,17 @@ protected ResourceLocation getOverlay() {
...
+        boolean warning = warningSupplier != null && warningSupplier.getAsBoolean();
+        if (warning) {
+            ResourceLocation texture = (left ? WARNING_LEFT : WARNING_RIGHT);
```
pupnewfster: You can just inline the texture so it is `innerDraw(guiGraphics, left ? WARNING_LEFT : WARNING_RIGHT);`

src/datagen/generated/mekanism/assets/mekanism/lang/en_us.json
```diff
@@ -2324,6 +2324,7 @@
   "gui.mekanism.radiation_dose": "Radiation Dose: %1$s",
   "gui.mekanism.radiation_exposure": "Radiation Exposure: %1$s",
   "gui.mekanism.radiation_exposure.entity": "Entity Radiation Exposure: %1$s",
+  "gui.mekanism.issues.redstone_unactivated": "Machine is currently inactive due to its redstone control setting.",
```
pupnewfster: This should be added via datagen (MekanismLangProvider and then running `runData`) rather than being manually added to this file

pupnewfster (changes requested):

bukowski912:
Whups, I pushed a cache file. Is there a reason why it isn't in the `.gitignore`? I'll leave it for now - if this is a mistake, let me know and I will fix it right away. It seems to be already in the repository, at least on my end.

bukowski912:
I hate GitHub mobile.

pupnewfster:
The cache file is so that you don't have to rewrite all the files to disk when you first checkout and run datagen.

src/datagen/main/java/mekanism/client/lang/MekanismLangProvider.java
```diff
@@ -1182,6 +1182,7 @@ private void addMisc() {
         add(MekanismLang.ISSUE_INPUT_DOESNT_PRODUCE_OUTPUT, " - Input does not produce output");
         add(MekanismLang.ISSUE_INVALID_OREDICTIONIFICATOR_FILTER, " - Filter is no longer valid or supported");
         add(MekanismLang.ISSUE_FILTER_HAS_BLACKLISTED_ELEMENT, " - Filter contains at least one element that is blacklisted");
+        add(MekanismLang.ISSUE_REDSTONE_UNACTIVATED, " - Inactive because of the redstone control setting");
```
thiakil: - I don't like the name. Something like `REDSTONE_PREVENTS_ACTIVATION`
- the text would be better as `Redstone conditions not met` or similar (the _setting_ is not the problem, the condition not being met is)

bukowski912: Sure, we'll go with those.

bukowski912:
Note that the special blending technique used for the warning overlay may have performance drawbacks, especially in light of previous optimization (e.g. 609b030). In addition, I renamed `GuiSideHolder#draw()` to `drawHolder()` as I thought that was much clearer. Let me know if that was an overstep.

thiakil:
>  In addition, I renamed GuiSideHolder#draw() to drawHolder() as I thought that was much clearer. Let me know if that was an overstep.

I'd rather that be in a separate PR, just to avoid the noise it creates in the diffs

src/main/java/mekanism/common/tile/base/TileEntityMekanism.java
```diff
@@ -908,6 +909,7 @@ public void addContainerTrackers(MekanismContainer container) {
         }
         if (supportsRedstone()) {
             container.track(SyncableEnum.create(RedstoneControl.BY_ID, RedstoneControl.DISABLED, () -> controlType, value -> controlType = value));
+            container.track(SyncableBoolean.create(this::isPowered, value -> redstone = value));
```
pupnewfster, line 912: Do we also need to sync `redstoneLastTick`?

src/main/java/mekanism/client/gui/GuiMekanismTile.java
```diff
@@ -39,7 +41,10 @@ protected void addGenericTabs() {
             upgradeWindowTab = addRenderableWidget(new GuiUpgradeWindowTab(this, tile, () -> upgradeWindowTab));
         }
         if (tile.supportsRedstone()) {
-            addRenderableWidget(new GuiRedstoneControlTab(this, tile));
+            addRenderableWidget(new GuiRedstoneControlTab(this, tile)
+                .warning(WarningType.REDSTONE_SIGNAL_ABSENT, () -> !tile.isRedstoneActivated() && tile.getControlType() == RedstoneControl.HIGH)
+                .warning(WarningType.REDSTONE_SIGNAL_PRESENT, () -> !tile.isRedstoneActivated() && tile.getControlType() == RedstoneControl.LOW)
+                .warning(WarningType.REDSTONE_PULSE_REQUIRED, () -> !tile.isRedstoneActivated() && tile.getControlType() == RedstoneControl.PULSE));
```
pupnewfster: Can you switch the order of these so it checks the control type matches before checking if it isn't activated?

pupnewfster (changes requested):

pupnewfster (approved):

Issue #8268, opened 2024-12-07
Raised mod incompatibility with MekaSuit HUD


JohnGoten:
### Issue description

Originally discovered in a personal instance (modpack development) of mine, I noticed after having configured the Raised mod to have a more lifted, bedrock-esque, hotbar visual, equipping a MekaSuit would have its HUD moved up by the equal amount as it was cut off at the top already, normally sitting comfortably within what's visible on screen
I was able to recreate this behavior within an isolated instance only using Mekanism and Raised, not even Sodium (+Iris) present. I also created an issue https://github.com/yurisuika/Raised/issues/100 and was shortly informed on their discord of them seemingly only hooking into the existing GUI layers present. Of course given the clearly unintended behavior, I can only assume the MekaSuit HUD is tied to that of the hotbar which should instead be disconnected and resolve this issue (fingers crossed).
![2024-12-07_15 13 32](https://github.com/user-attachments/assets/e54f2a7f-d336-4607-948a-5d2c31ee8311)

### Steps to reproduce

1. Equip a full set of MekaSuit armor
2. Access the Raised config via "`"
3. Move the hotbar (top left has options for chat etc. which are unaffected) on the Y axis using the top right slider
4. Watch as the HUD moves off-screen

### Minecraft version

1.21.1 (Latest)

### NeoForge version

21.1.84

### Mekanism version

10.7.7 (Latest)

### Other relevant versions

Raised 4.0.1 (Latest 1.21.1 as of now) https://modrinth.com/mod/raised/version/NeoForge-1.21.1-4.0.1

### If a (crash)log is relevant for this issue, link it here: (It's almost always relevant)

https://mclo.gs/IjUkWxk

JohnGoten:
Thanks to yurisuika's help, never previously dealt with IntelliJ as your average consumer, I did manage to change the ClientRegistration.java line 318
![image](https://github.com/user-attachments/assets/2890fdbc-6e36-4ecb-b05d-8efdb71ca54d)


`        event.registerAbove(VanillaGuiLayers.HOTBAR, Mekanism.rl("hud"), MekanismHUD.INSTANCE);`

into

`       event.registerAbove(VanillaGuiLayers.CROSSHAIR, Mekanism.rl("hud"), MekanismHUD.INSTANCE);`

as this seemed to be the most reasonable option available, although I tried

`        event.registerAboveAll(Mekanism.rl("hud"), MekanismHUD.INSTANCE);`

[...]

PR #8285, opened 2024-12-24
Raised HUD compatibility

JohnGoten:
As originally discussed in https://github.com/mekanism/Mekanism/issues/8268 I have taken it upon myself to learn how to create a pull request in order to get this pushed and implemented in the foreseeable future. Already done some testing in a personalized instance to ensure no breakage which given the small change has not occurred, luckily. This does however result in the Raised mod's hotbar modification no longer affecting that of really any HUD elements also being moved alongside which is exactly what this intends to do.
 
moving status layer to be below VanillaGuiLayers.SLEEP_OVERLAY rather than above VanillaGuiLayer.HOTBAR, retaining the same type of implementation but with a different variable in use


bukowski912:
I'm going to stick my ~~nose~~ sniffer in:

Why did you chose `VanillaGuiLayers.SLEEP_OVERLAY`? What are its properties beneficial in this case, and can they be expected to hold for the forseeable future?

JohnGoten:
merely following yurisuika's advice as I was given on their discord near 3 weeks ago when this topic originally came up
since I am no expert at this, only going by logical thinking and evidence provided, it may be worthwhile checking on the server directly rather than this simple screenshot (pls don't hate me for sharing this tyvm)
image
this is why I encouraged potential alternatives based on the existing system as I highly doubt Raised needs any changing

pupnewfster:
It is intentional that the HUD renders on the level it does. The boss bar rendering behind the MekaSuit's hud for example is undesirable.

PR #8317, opened 2025-01-24
Disambiguate some SaveHandler names.

kylev:
Similar to in 577523afffbbafafa46de8b4653d632ffff7b744, there are some ambiguous inner-class names that cause some compilers to whine/choke.


pupnewfster:
I am going to go ahead and guess that said compiler is eclipse? (I am not opposed to changing it if so, just want to know a concrete example of what compiler you were having issues with it for)

thiakil (approved):

kylev:
It's actually VS Code's Java plugin, OpenJDK 21 may be coming into play sideways. Builds still work, but the lexical-only syntax checked gets cranky when it can't resolve the implied inner-class ... name collision that still works?

PR #8325, opened 2025-02-01
Set up two new source-sets that allow IntelliJ to resolve module dependencies for runs

shartte:
Basically IJ (well, and Gradle) need a way to know which submodules need to be built when you run the various runs.
There is no sourceSet / module that actually depends on all the parts that are required.

This PR introduces the source sets `runMain` and `runData`, which have no actual sources. Since IJ imports source sets as IJ modules, they can be set as the main module of a IJ run configuration, and IJ will compile the dependent modules correctly.

To this end, I set up runMain to depend on api, main and the secondary modules, while runData inherits everything from runMain, plus all the datagen sets. There's no third source set for the game tests, since the game test source set already has the necessary dependencies for IJ.

Note on delegation: This will make running without delegating the build to Gradle possible, and also fixes running with delegation enabled. In delegation-mode, IJ only runs `:classes` (for the main module), which only compiles *compile time* dependencies of the `main` source set. By setting the module in IJ to `runMain`, it will run `:runMainClasses`, which builds the correct dependences in delegate-to-Gradle mode too.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -127,6 +127,10 @@ def setupExtraSourceSets(SourceSet base, boolean includeExtra = true) {
     if (base != project.sourceSets.gameTest) {
         project.sourceSets.gameTest.compileClasspath += base.output
     }
+    // Add the directory where Gradle would usually put the AP sources explicitly so IntelliJ picks it up
+    // If we don't do this, IJ seems to run a weird heuristic on where to put the AP sources and put them
+    // into the generated resource directory instead
+    base.java.srcDir project.file("build/generated/sources/annotationProcessor/java/${base.name}")
     //Setup and extend configurations for alternate modules. First by making the implementation, compileOnly, runtimeOnly equivalents
     // for those modules extend the main ones
     def baseImplementation = project.configurations.maybeCreate(base.getTaskName(null, 'implementation'))
--- a/build.gradle
+++ b/build.gradle
@@ -312,8 +312,6 @@ neoForge {
                 jvmArguments.addAll('-XX:+IgnoreUnrecognizedVMOptions', '-XX:+AllowEnhancedClassRedefinition')
             }
 
-            gameDirectory = project.file("runs/$name")
-
             // Uncomment this to get verbose debug logging
             // logLevel = org.slf4j.event.Level.DEBUG
 
--- a/build.gradle
+++ b/build.gradle
@@ -73,6 +73,20 @@ sourceSets {
     gameTest {
         compileClasspath += api.output
     }
+    // These source-sets only exist for the sake of assembling a dependency graph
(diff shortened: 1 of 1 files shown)


thiakil:
image

The versions appear to be missing with run as IJ? (they work with gradle, since it does the process resources thing)

Also, perhaps not directly related to this PR, but all the `localRuntime` mods are missing during run/debug

thiakil:
and not sure the Tests should be there? this is just a normal run, not a test run (both run as gradle and IJ)

thiakil:
image

Annotation processor is outputting things in `src/datagen` which is... wrong (run as IJ only)

shartte:
Ah, the lack of JEI et al is actually a problem outside of the PR too, since I forgot to add `localRuntime` to runtime classpath. I think NG defines this configuration implicitly, while MDG does not.

shartte:
Alright, the annotation processor problem seems to be a weird IntelliJ heuristic bug, and we're not the first: https://stackoverflow.com/questions/60455845/intellij-is-using-unexpected-source-directory-from-annotation-processing-output

Ok, the solution seemed to be adding the generated srcDir explicitly to the source set, so that IJ picks it up. That way, its heuristic now picks the same generated sources folder as Gradle does, and the AP works.

shartte:
@thiakil Alrighty, while the versions *can* be fixed (it requires adapting an approach similar to the MDK), if it doesn't hurt in dev, I'd just leave it as-is. The rest should be fixed.

thiakil:
Legend, thanks!

thiakil:
sadly 0c3d6e7c57dbd438264466b6f1a9a3df69e9bb79 doesnt seem to have fixed the annotation processor output for me (still in datagen).

I've worked out that it might be related to the `Store generated sources relative to` setting - it imports as `Module content root` but setting `Module output directory` seems to hide them from git at least.

I seem to be able to make it behave with `options.generatedSourceOutputDirectory = project.buildDir("generated/sources/annotationProcessor/java/${sourceSet.name}")`
BUT, now it seems MDG might be confused?
image

Worst case we might have to patch the project xml manually like https://github.com/palantir/gradle-processors/commit/a8263e21917126cb912edef66f18faa104bd495f

thiakil:
argh, but it refuses to clean it... (so the annotation processor then fails as a file already exists)

`Output path D:\git\mekanism\build\generated\sources\annotationProcessor\java\generators\generated intersects with a source root. Only files that were created by build will be cleaned.`

thiakil:
merging anyway as the root cause for the PR is solved, and Gradle mode build still works

thiakil:
@shartte I'm also getting weird empty folders in the jar now...
image

PR #8341, opened 2025-03-01
Work on moving things towards being more data driven

pupnewfster:
- Refactor Chemical Stacks to keep track of the holder rather than the chemical
- Fix fluid tank dispenser behaviors not actually being added
- Deprecated all our providers in favor of using holders. Mojang is moving towards using holders more and more and this preps us for when they finally finish moving items to being a data driven registry
- Deprecated `Chemical#isGaseous` and add a chemical tag `mekanism:gaseous` that can be used in its place.
- Deprecated `Chemical#getOreTag` in favor of declaring the solid representation for crystallizers via a data map
- Deprecated all chemical attributes adding corresponding values of them that get registered as data maps
- Fixed chemicals returning true from `hasAttributesWithValidation` when the only attribute they have is radiation, and the radiation manager is disabled
- Made collecting and displaying chemical tooltips easier. It is now a method on both the stack and the chemical.
- Reduced usages of ItemLike in favor of Holders where it made sense, as if I had to guess Mojang will eventually do away with it.
- BREAKING: Removed the ethene fuel and hydrogen fuel configs and replaced them with being defined and queried from the datamap.


src/api/java/mekanism/api/chemical/Chemical.java
```diff
@@ -38,83 +56,176 @@ public class Chemical implements IChemicalProvider, IChemicalAttributeContainer<
...
+    private final List<IChemicalAttribute> attributesView = Collections.unmodifiableList(attributes);
     private final ResourceLocation iconLocation;
     private final int tint;
-    private boolean isRadioactive;
+    private double radioactivity;
     private boolean hasAttributesWithValidation;
+
+
+    @Deprecated(forRemoval = true, since = "10.7.11")
+    private final Map<Class<? extends ChemicalAttribute>, ChemicalAttribute> legacyAttributeMap;
```
thiakil: I don't really see the need to maintain a legacy attribute as well as the 'new' ones. Why replacing them anyway?

src/api/java/mekanism/api/chemical/Chemical.java
```diff
@@ -226,44 +378,52 @@ public boolean is(TagKey<Chemical> tag) {
...
-        return MekanismAPI.CHEMICAL_REGISTRY.wrapAsHolder(this);
+        return builtInRegistryHolder();
+    }
+
+    /**
+     * Intrusive holder, similar to vanilla this is deprecated and will eventually be moved away from.
+     *
+     * @since 10.7.11
+     */
+    @Deprecated
```
thiakil: don't deprecate it, mark it as internal api

src/api/java/mekanism/api/chemical/Chemical.java
pupnewfster: The main reason to maintain the legacy attributes is the new ones can be changed on `/reload`, and I need a way to know which ones to clear and which ones won't be added back. And the reason I am replacing them is because I am aiming to move Chemical's to being a datapack registry in 1.22, and it will make the in code ones less convenient to do. Not to mention this will allow pack devs to use datapacks to apply attributes (such as fuel values) to things without having to use CraftTweaker. And then test their changes with a reload instead of having to restart the entire game.

src/api/java/mekanism/api/chemical/Chemical.java
pupnewfster: The reason I had deprecated it was to follow how vanilla marks all their builtInRegistryHolders, as they do want to remove them. I can definitely mark it as internal, but I wonder if we should mark it as both, as ideally we also want to avoid using it even internally if we can say reference the holder through the chemical stack.

src/api/java/mekanism/api/chemical/Chemical.java
thiakil: I think we should just do it all as one

src/api/java/mekanism/api/chemical/Chemical.java
thiakil: and how exactly is vanilla going to remove it? The fact is, it's getting used anyway so in this iteration it's not deprecated as there is no other way to do it (or if there is, you should do that)

src/api/java/mekanism/api/chemical/Chemical.java
pupnewfster: Honestly this won't be in for long (once we port to 1.22 it will be gone) and this is just for backwards compat to avoid breaking addons. If you really want I could break all the stuff now given we are still in alpha, but personally I see no reason to do so (especially as I have already written the backward compat code).

src/api/java/mekanism/api/chemical/Chemical.java
pupnewfster: I have gone ahead and removed the builtinRegistryHolder entirely from this PR as after all the refactoring I had done, it ended up actually being pretty easy to remove.

src/api/java/mekanism/api/MekanismAPI.java
```diff
@@ -135,12 +147,20 @@ private static <T> ResourceKey<Registry<MapCodec<? extends T>>> codecRegistryKey
...
 
-    //TODO: Potentially define this with DeferredHolder for purposes of fully defining them outside of the API
     /**
      * Empty Chemical instance.
      *
      * @since 10.7.0
+     *
+     * @deprecated Use {@link #EMPTY_CHEMICAL_HOLDER}
      */
+    @Deprecated(forRemoval = true, since = "10.7.11")//TODO - 1.22: Inline creation instead of creating the instance in the API
```
thiakil, line 158: while it will be breaking, should we just remove this? chances are slim that someone uses it

src/api/java/mekanism/api/MekanismIMC.java
```diff
@@ -85,8 +87,25 @@ public static void addModuleContainer(ModuleContainerTarget moduleContainer) {
      *
      * @apiNote Call this method during the {@link InterModEnqueueEvent}.
      * @since 10.5.0
+     * @deprecated Call via {@link #addModulesToAll(Holder[])}
```
thiakil: this should be `Use {@link #X} instead` so allow IDEA's heuristic intention to work best

src/api/java/mekanism/api/MekanismIMC.java
```diff
@@ -60,7 +62,7 @@ private MekanismIMC() {
      * @apiNote Call this method during the {@link InterModEnqueueEvent}.
      * @see #addModuleContainer(ModuleContainerTarget)
      */
-    public static void addModuleContainer(ItemLike container, String imcMethod) {
+    public static void addModuleContainer(ItemLike container, String imcMethod) {//TODO - 1.22: Switch this to taking a Holder<Item>
```
thiakil: instead of a todo, can the new one be added now and this one be deprecated?

src/api/java/mekanism/api/MekanismIMC.java
```diff
@@ -192,7 +334,7 @@ public record ModuleContainerTarget(Item container, String imcMethod) {
          * @param container Item that will be the module container.
          * @param imcMethod Method used to add modules as supported to the container.
          */
-        public ModuleContainerTarget(ItemLike container, String imcMethod) {
+        public ModuleContainerTarget(ItemLike container, String imcMethod) {//TODO - 1.22: Switch this to taking a Holder<Item>
```
thiakil: can this not be done now?

src/api/java/mekanism/api/chemical/BasicChemicalTank.java
```diff
@@ -22,38 +23,88 @@ public class BasicChemicalTank implements IChemicalTank, IChemicalHandler {
...
+     *
+     * @since 10.7.11
+     */
+    public static final BiPredicate<Holder<Chemical>, @NotNull AutomationType> holderNotExternal = ConstantPredicates.notExternal();
+    /**
+     * A bi predicate that returns {@code true} for any input when the automation type is manual.
+     *
+     * @since 10.7.11
+     */
+    public static final BiPredicate<Holder<Chemical>, @NotNull AutomationType> holderManualOnly = ConstantPredicates.manualOnly();
```
thiakil: These should not be needed at all, the `ConstantPredicates` methods can be used directly

thiakil: all they're doing is a static import with extra steps

src/api/java/mekanism/api/chemical/BasicChemicalTank.java
```diff
@@ -106,11 +157,30 @@ public static IChemicalTank createAllValid(long capacity, @Nullable IContentsLis
      *
      * @implNote The created tank will always allow {@link AutomationType#MANUAL} extraction, and allow any {@link AutomationType} to insert into it.
      * @since 10.7.0 Previously was in ChemicalTankBuilder
+     * @deprecated Prefer using holders and calling {@link #createModern(long, Predicate, Predicate, IContentsListener)}
```
thiakil: `Use X instead`

src/api/java/mekanism/api/chemical/BasicChemicalTank.java
```diff
@@ -120,7 +190,9 @@ public static IChemicalTank create(long capacity, Predicate<Chemical> canExtract
      * @param listener  Contents change listener.
      *
      * @since 10.7.0 Previously was in ChemicalTankBuilder
+     * @deprecated Prefer using holders and calling {@link #createModern(long, Predicate, IContentsListener)}
```
thiakil: `Use X instead`

src/api/java/mekanism/api/chemical/BasicChemicalTank.java
```diff
@@ -138,7 +228,9 @@
      * @param listener  Contents change listener.
      *
      * @since 10.7.0 Previously was in ChemicalTankBuilder
+     * @deprecated Prefer using holders and calling {@link #inputModern(long, Predicate, IContentsListener)}
```
thiakil: `Use X instead`

src/api/java/mekanism/api/chemical/BasicChemicalTank.java
```diff
@@ -157,7 +267,9 @@
      * @param listener  Contents change listener.
      *
      * @since 10.7.0 Previously was in ChemicalTankBuilder
+     * @deprecated Prefer using holders and calling {@link #inputModern(long, Predicate, Predicate, IContentsListener)}
```
thiakil: `Use X instead`

src/api/java/mekanism/api/chemical/BasicChemicalTank.java
```diff
@@ -281,16 +533,40 @@
     @Nullable
     private final IContentsListener listener;
 
+    @Deprecated(forRemoval = true, since = "10.7.11")
```
thiakil, line 506: add use X instead javadoc

src/api/java/mekanism/api/chemical/BasicChemicalTank.java
```diff
@@ -266,9 +467,60 @@
...
+    private static BiPredicate<Holder<Chemical>, @NotNull AutomationType> wrapAutomationPredicateToModern(BiPredicate<Chemical, @NotNull AutomationType> predicate) {
+        if (predicate == alwaysTrueBi || predicate == internalOnly || predicate == notExternal || predicate == manualOnly) {
+            return (BiPredicate<Holder<Chemical>, @NotNull AutomationType>) (BiPredicate<?, @NotNull AutomationType>) predicate;
+        }
+        return (chemical, automationType) -> predicate.test(chemical.value(), automationType);
+    }
+
+    private final Predicate<Holder<Chemical>> validator;
+    //TODO - 1.22: Rename these to canExtract and canInsert
+    protected final BiPredicate<Holder<Chemical>, @NotNull AutomationType> canExtractModern;
```
thiakil: shouldn't we make these just use the stack?

src/api/java/mekanism/api/chemical/Chemical.java
```diff
@@ -1,120 +1,251 @@
...
+     * @since 10.7.11
+     */
+    public static final Codec<Holder<Chemical>> HOLDER_CODEC = MekanismAPI.CHEMICAL_REGISTRY.holderByNameCodec();//TODO - 1.22: Rename this to just CODEC
     /**
      * A stream codec which can be used to encode and decode chemicals over the network.
      *
      * @since 10.6.0
+     * @deprecated Prefer using {@link #HOLDER_STREAM_CODEC}
      */
+    @Deprecated(forRemoval = true, since = "!0.7.11")
```
thiakil: accidental `!`

src/api/java/mekanism/api/chemical/Chemical.java
```diff
@@ -1,120 +1,251 @@
...
+            return MekanismAPI.EMPTY_CHEMICAL_HOLDER;
+        }
+        Optional<RegistryLookup<Chemical>> chemicalLookup = lookupProvider.lookup(MekanismAPI.CHEMICAL_REGISTRY_NAME);
+        //noinspection OptionalIsPresent - Capturing lambda
+        if (chemicalLookup.isEmpty()) {
+            return MekanismAPI.EMPTY_CHEMICAL_HOLDER;
         }
-        ResourceLocation name = ResourceLocation.tryParse(tag);
-        return name == null ? MekanismAPI.EMPTY_CHEMICAL : MekanismAPI.CHEMICAL_REGISTRY.get(name);
+        return Optional.ofNullable(ResourceLocation.tryParse(tag))
```
thiakil: please not Optional chains

src/api/java/mekanism/api/chemical/Chemical.java
```diff
@@ -1,120 +1,251 @@
...
+    @SuppressWarnings("removal")
     public Chemical(ChemicalBuilder builder) {
-        //Copy the map to support addAttribute
-        this.attributeMap = new HashMap<>(builder.getAttributeMap());
         this.iconLocation = builder.getTexture();
         this.tint = builder.getTint();
-        this.isRadioactive = attributeMap.containsKey(Radiation.class);
-        this.hasAttributesWithValidation = isRadioactive || attributeMap.values().stream().anyMatch(ChemicalAttribute::needsValidation);
-        this.oreTag = builder.getOreTag();
+        //TODO - 1.22: Remove constructor logic that is below here
```
thiakil: can this part be moved to a private static method with that getting the todo?

src/api/java/mekanism/api/chemical/Chemical.java
```diff
@@ -1,120 +1,251 @@
...
+        //Copy the map to support addAttribute
+        this.legacyAttributeMap = new HashMap<>(builder.getAttributeMap());
+        for (ChemicalAttribute legacyAttribute : legacyAttributeMap.values()) {
+            if (legacyAttribute instanceof Radiation radiation) {
+                this.radioactivity = this.legacyRadioactivity = radiation.getRadioactivity();
+            } else if (legacyAttribute.needsValidation()) {
+                //Skip radioactive attributes when checking if we have any that need validation, so that we properly return false
+                // when the radiation manager is disabled
+                this.hasAttributesWithValidation = this.hasLegacyAttributesWithValidation = true;
```
thiakil: i dislike this for a constant assignment

src/api/java/mekanism/api/chemical/Chemical.java
```diff
@@ -1,120 +1,251 @@
...
         return this;
     }
 
     @Override
     public String getTranslationKey() {
         if (translationKey == null) {
-            translationKey = Util.makeDescriptionId("chemical", getRegistryName());
+            //Note: Because chemical registry has a default key, we have to query the name via the resource key so that we can pass null
+            // to makeDescriptionId in cases when our chemical is unregistered
+            translationKey = Util.makeDescriptionId("chemical", MekanismAPI.CHEMICAL_REGISTRY.getResourceKey(this)
```
thiakil, line 218: I'd really prefer this to use `getKey` with a default check (or better, pr to Neo adding a nullable `getKey` for `DefaultedMappedRegistry`)

src/api/java/mekanism/api/chemical/Chemical.java
```diff
@@ -1,120 +1,251 @@
...
+     *
+     * @param type The type of the attribute to check for.
+     *
+     * @return if this chemical has the attribute.
+     *
+     * @since 10.7.11
+     */
+    @SuppressWarnings("removal")
+    @Deprecated(forRemoval = true, since = "10.7.11")
+    public boolean hasLegacy(Class<? extends ChemicalAttribute> type) {
```
thiakil, line 252: the uses of this look like they could be helpers added to this class

src/api/java/mekanism/api/chemical/attribute/ChemicalAttributes.java
```diff
@@ -162,15 +193,43 @@
...
             return heatedChemical.getChemical();
         }
+
+        /**
+         * Attempts to convert this fuel attribute into a modern fuel attribute
+         *
+         * @since 10.7.11
+         */
+        @Nullable
+        public mekanism.api.datamaps.chemical.attribute.CooledCoolant asModern() {
```
thiakil, line 230: should be marked as internal api

src/api/java/mekanism/api/chemical/attribute/ChemicalAttributes.java
```diff
@@ -191,16 +270,43 @@
...
             return cooledChemical.getChemical();
         }
+
+        /**
+         * Attempts to convert this fuel attribute into a modern fuel attribute
+         *
+         * @since 10.7.11
+         */
+        @Nullable
+        public mekanism.api.datamaps.chemical.attribute.HeatedCoolant asModern() {
```
thiakil, line 309: should be marked as internal api

src/api/java/mekanism/api/chemical/attribute/ChemicalAttributes.java
```diff
@@ -255,8 +373,39 @@
             return energyDensity.getAsLong() / ticks;
         }
 
+        /**
+         * Attempts to convert this fuel attribute into a modern fuel attribute
+         *
+         * @since 10.7.11
+         */
+        @Nullable
+        public ChemicalFuel asModern() {
```
thiakil, line 413: internal api

src/api/java/mekanism/api/chemical/ChemicalStack.java
```diff
@@ -456,28 +506,85 @@
...
-        int code = 1;
-        code = 31 * code + getChemical().hashCode();
-        code = 31 * code + Long.hashCode(getAmount());
-        return code;
+        if (isEmpty()) {
+            return 0;
+        }
+        //Note: chemical is not null here, and we know it isn't empty so we can just directly reference it
+        // rather than having to check if it is empty again
+        return Objects.hash(chemical, amount);
```
thiakil: while 'clean', this isn't a good method to use, as it's varargs

src/api/java/mekanism/api/datamaps/chemical/attribute/IChemicalAttribute.java
```diff
@@ -0,0 +1,39 @@
...
+ * @since 10.7.11
+ */
+public interface IChemicalAttribute {
+
+    /**
+     * Converts this attribute into a legacy variant.
+     */
+    @SuppressWarnings("removal")
+    @Deprecated(forRemoval = true, since = "10.7.11")
+    mekanism.api.chemical.attribute.ChemicalAttribute toLegacyAttribute();
```
thiakil, line 20: internal api annotation instead, probably not deprecated?

src/api/java/mekanism/api/providers/IModuleDataProvider.java
```diff
@@ -1,11 +1,15 @@
...
+import java.util.Objects;
+import mekanism.api.MekanismAPI;
 import mekanism.api.gear.ICustomModule;
 import mekanism.api.gear.ModuleData;
 import net.minecraft.MethodsReturnNonnullByDefault;
 import net.minecraft.resources.ResourceLocation;
 
+@SuppressWarnings("removal")
 @MethodsReturnNonnullByDefault
+@Deprecated(forRemoval = true, since = "10.7.11")
```
thiakil, line 12: should this have a replace with comment?

src/api/java/mekanism/api/gear/ModuleData.java
```diff
@@ -22,22 +22,27 @@
...
 @NothingNullByDefault
-public class ModuleData<MODULE extends ICustomModule<MODULE>> implements IModuleDataProvider<MODULE> {
+@SuppressWarnings("removal")
+public class ModuleData<MODULE extends ICustomModule<MODULE>> implements mekanism.api.providers.IModuleDataProvider<MODULE> {
 
     private final Function<@NotNull IModule<MODULE>, @NotNull MODULE> constructor;
     private final Int2ObjectMap<ConstructedConfigData> configData;
+    @Deprecated(forRemoval = true, since = "10.7.11")
     private final IItemProvider itemProvider;
+    @Nullable//TODO - 1.22: Make this nonnull
```
thiakil, line 44: why is it nullable?

src/datagen/generated/mekanism/assets/minecraft/atlases/blocks.json
```diff
@@ -50,15 +50,15 @@
     },
     {
       "type": "minecraft:single",
-      "resource": "mekanism:infuse_type/bio"
```
thiakil, line 53: uhh why did these change order? were they not being sorted before?

src/main/java/mekanism/client/gui/element/custom/GuiQIOCrystallizerScreen.java
```diff
@@ -68,22 +71,31 @@
...
                     iterStacks.clear();
-                    if (!prevSlurry.isEmptyType() && !prevSlurry.is(MekanismAPITags.Chemicals.DIRTY)) {
-                        TagKey<Item> oreTag = prevSlurry.getOreTag();
-                        if (oreTag != null) {
-                            for (Holder<Item> ore : BuiltInRegistries.ITEM.getTagOrEmpty(oreTag)) {
-                                iterStacks.add(new ItemStack(ore));
+                    if (!prevSlurry.is(MekanismAPI.EMPTY_CHEMICAL_KEY)) {
+                        ChemicalSolidTag tag = chemical.getData(IMekanismDataMapTypes.INSTANCE.chemicalSolidTag());
+                        if (tag != null) {
+                            tag.lookupTag().stream()
```
thiakil: i hate this. Much better to unwrap it
```
Named<Item> tagContents = tag.lookupTag().orElse(null);
if (tagContents != null) {
      for (Holder<Item> entry : tagContents) {
          iterStacks.add(new ItemStack(entry));
      }
}
```

src/main/java/mekanism/client/recipe_viewer/emi/ChemicalEmiIngredientSerializer.java
```diff
@@ -28,10 +29,10 @@ public String getType() {
...
 
     void addEmiStacks(EmiRegistry emiRegistry) {
-        for (Chemical chemical : MekanismAPI.CHEMICAL_REGISTRY) {
-            if (!chemical.isEmptyType()) {//Don't add the empty type. We will allow EMI to filter out any that are hidden from recipe viewers
-                emiRegistry.addEmiStack(new ChemicalEmiStack(chemical, 1));
-            }
-        }
+        MekanismAPI.CHEMICAL_REGISTRY.holders()
+              //Don't add the empty type. We will allow EMI to filter out any that are hidden from recipe viewers
+              .filter(chemical -> !chemical.is(MekanismAPI.EMPTY_CHEMICAL_KEY))
```
thiakil: i'd prefer this was just a `forEach` lambda rather than a multi-step stream

src/main/java/mekanism/client/recipe_viewer/jei/ChemicalStackHelper.java
```diff
@@ -41,9 +43,10 @@ public String getDisplayName(ChemicalStack ingredient) {
     }
 
     @Override
-    @Deprecated(forRemoval = true)
+    @SuppressWarnings("removal")
+    @Deprecated(forRemoval = true, since = "19.9.0")
```
thiakil: 19?

src/main/java/mekanism/common/lib/radiation/RadiationManager.java
```diff
@@ -147,6 +145,16 @@ private void markDirty() {
         }
     }
 
+    @Override
+    public double baselineRadiation() {
+        return 0.000_000_100;//100 nSv/h
```
thiakil: i dont understand why removing the constant

src/api/java/mekanism/api/MekanismIMC.java
pupnewfster: It is technically a tiny breaking change as I want to switch ModuleContainerTarget's record to use `Holder<Item>`, but as we are in alpha, and also [literally no one except us is trying to call the container method](https://waifu.neoforged.net/d/aec29a5ymel1cc/waifu-classes?orgId=1&from=now-1y&to=now&timezone=browser&var-version=1.21.1-neoforge&var-class=mekanism%2Fapi%2FMekanismIMC$ModuleContainerTarget&var-method=.%2A&var-field=.%2A) I am just going to go ahead and do it now.

src/main/java/mekanism/client/recipe_viewer/jei/ChemicalStackHelper.java
pupnewfster: That's the JEI version that it was deprecated for removal in, it is abstract though so we are still required to implement it.

src/main/java/mekanism/client/recipe_viewer/emi/ChemicalEmiIngredientSerializer.java
pupnewfster: Changed, I initially did it to avoid capturing lambda's, but realistically you are probably right, and it isn't like this method gets called frequently anyway.

src/datagen/generated/mekanism/assets/minecraft/atlases/blocks.json
pupnewfster: They weren't exactly being sorted by us, and when I switched to iterating the registry by entryset instead of value, it changed. This is in part due to me having messed up the switch. I then noticed it when doing the code review of the PR myself, and fixed that, but I decided to keep in the sorting logic I added just so that it is a bit more consistent overall.

src/api/java/mekanism/api/gear/ModuleData.java
pupnewfster: Because if someone instantiates it via the deprecated `IItemProvider` way, we don't know for certain (given this is during registration) that calling `IItemProvider#asItem` won't cause an exception of some sort to be thrown. I doubt it would, given the module data registry happens after vanilla ones, but I figured just to be safe we were better off just lazily resolving the deprecated way.

src/api/java/mekanism/api/providers/IModuleDataProvider.java
pupnewfster: Unsure? Basically it is just use `Holder<ModuleData<?>>` or `DeferredHolder<ModuleData<?>, ...>` in places. There isn't a different thing that should be implemented in place of it.

src/api/java/mekanism/api/MekanismAPI.java
pupnewfster: It seems to be [used](https://waifu.neoforged.net/d/aec29a5ymel1cc/waifu-classes?orgId=1&from=2025-03-08T12:07:11.627Z&to=2025-03-08T18:07:11.627Z&timezone=browser&var-version=1.21.1-neoforge&var-class=mekanism%2Fapi%2FMekanismAPI&var-method=.%2A&var-field=%5EEMPTY_CHEMICAL$&viewPanel=panel-20) by a few mods

src/api/java/mekanism/api/chemical/BasicChemicalTank.java
pupnewfster: I was wondering about that, so I will go ahead and so so, even though we don't care about the actual amount and we don't have data components for chemicals.

src/api/java/mekanism/api/chemical/Chemical.java
pupnewfster: We can't directly just have a default check, as we want the empty chemical to have a translatable name. Personally I think we just leave it like this for now, and maybe at some point make a PR to Neo (I can add a TODO in the code about it if you want), or maybe when moving to data pack registries this will just go away to an extent.

src/api/java/mekanism/api/chemical/Chemical.java
pupnewfster: Personally I am a bit against that, mostly since we use each only in one place, and I wouldn't want to be just calling the helpers directly as I don't think the helpers make sense to keep after the legacy way is gone.

src/api/java/mekanism/api/chemical/Chemical.java
pupnewfster: PR made: https://github.com/neoforged/NeoForge/pull/2013, and I will add a TODO in code so that we don't have to hold merging back on waiting for it, and can just switch to it when we are able to.

src/main/java/mekanism/client/recipe_viewer/jei/ChemicalStackHelper.java
thiakil: ahh, perhaps we can add `JEI: ` to the string?

thiakil (approved):

PR #8360, opened 2025-03-22
Extract `IFreeRunnersItem`, add integration with Curio

xbt573:
This patch introduces integration of Free Runners with Curio slots. As side effect it introduces `IFreeRunners` interface and extracts shared logic into it. It also renames `FreeRunnerMode` to `FreeRunnersMode` to comply to other class names, let me know if it needs to get renamed back

It is my first contribution to Minecraft mod (and Java project too), so I might make some mistakes (although I tested it myself)

### DIFF
--- a/src/main/java/mekanism/common/item/gear/ItemFreeRunners.java
+++ b/src/main/java/mekanism/common/item/gear/ItemFreeRunners.java
@@ -1,20 +1,14 @@
 package mekanism.common.item.gear;
 
-import com.mojang.serialization.Codec;
-import io.netty.buffer.ByteBuf;
 import java.util.List;
-import java.util.Locale;
 import java.util.function.Consumer;
-import java.util.function.IntFunction;
-import mekanism.api.IIncrementalEnum;
-import mekanism.api.annotations.NothingNullByDefault;
+
 import mekanism.api.text.EnumColor;
-import mekanism.api.text.IHasTextComponent.IHasEnumNameTextComponent;
-import mekanism.api.text.ILangEntry;
 import mekanism.common.Mekanism;
 import mekanism.common.MekanismLang;
 import mekanism.common.config.MekanismConfig;
-import mekanism.common.item.gear.ItemFreeRunners.FreeRunnerMode;
+import mekanism.common.item.interfaces.IFreeRunnersItem;
+import mekanism.common.item.interfaces.IFreeRunnersItem.FreeRunnersMode;
 import mekanism.common.item.interfaces.IHasConditionalAttributes;
 import mekanism.common.item.interfaces.IItemHUDProvider;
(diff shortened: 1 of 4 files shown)


src/main/java/mekanism/common/item/interfaces/IFreeRunnersItem.java
```diff
@@ -0,0 +1,101 @@
...
+import net.minecraft.world.entity.EquipmentSlot;
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.item.ItemStack;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Locale;
+import java.util.function.IntFunction;
+import java.util.function.Predicate;
+
+public interface IFreeRunnersItem {
```
thiakil, line 27: I don't really see the need for the interface?

src/main/java/mekanism/common/item/gear/ItemFreeRunners.java
```diff
@@ -126,56 +116,13 @@ public boolean supportsSlotType(ItemStack stack, @NotNull EquipmentSlot slotType
...
     @Override
     public void adjustAttributes(ItemAttributeModifierEvent event) {
-        if (getMode(event.getItemStack()) == FreeRunnerMode.NORMAL) {
+        if (getMode(event.getItemStack()) == FreeRunnersMode.NORMAL) {
             event.addModifier(Attributes.MOVEMENT_EFFICIENCY, MOVEMENT_EFFICIENCY, EquipmentSlotGroup.FEET);
         }
     }
 
-    @NothingNullByDefault
-    public enum FreeRunnerMode implements IIncrementalEnum<FreeRunnerMode>, IHasEnumNameTextComponent, StringRepresentable {
```
thiakil, line 135: please at least restore the original enum name, it creates unnecessary noise in the diff

thiakil (changes requested):

src/main/java/mekanism/common/item/interfaces/IFreeRunnersItem.java
xbt573: Jetpacks also have their own interface to make more clear that operation is happening on some jetpack rather that concrete type, in my opinion this increases clarity and consistency over code base.

If you think that it is REALLY unnecessary I will remove it and make it operate on ItemFreeRunners.

src/main/java/mekanism/common/item/gear/ItemFreeRunners.java
xbt573: Is no noise in diff matters more than consistency and readability? If yes, I will fix this

thiakil:
merged manually

PR #8363, opened 2025-03-29
Paxels now can have custom durability

HwakYuk:
`TieredItem` calls `Item.Properties#durability` in its constructor, which overrides any prior durability configurations. To address this, I've introduced `LockableProperties`, enabling the durability to be locked. This allows tools like paxels to maintain its custom durability settings.  May not be an ideal solution though.


pupnewfster:
I will try to double check this later, but I do think a custom properties extension is the only way to fix it properly. I had to do similar in another mod to prevent tiered item from applying the durability components to an item of mine that wasn’t meant to have durability.

Issue #8361, opened 2025-03-23
java.lang.IllegalStateException: Do not call getCapability on an invalid cache or from the invalidation listener!


ZwiebelTVDE:
### Issue description

Our server experieces frequent crashes when playing AllTheMods 10 with Mekanism version 10.7.12 it appeares to be linked to some advanced logistigal pipes in one of our bases.

### Steps to reproduce

1. Create a server with neoforge and the modpack ATM10
2. Place Advanced Logisical pipes to connect some inventories
3. Leave the server and join back on the server

### Minecraft version

1.21.1 (Latest)

### NeoForge version

21.1.133

### Mekanism version

10.7.12

### Other relevant versions

See bottom of crashlog:

### If a (crash)log is relevant for this issue, link it here: (It's almost always relevant)

https://gist.github.com/ZwiebelTVDE/d541a16bca337a403ed7fa1d830ce9e7

NanobotZ:
I have an identical problem, for me it was a logistical pipe connected to an AE2 block (ME Interface).
If I had to guess, the chunk with the pipe loaded before the entire AE2 network was loaded.

Issue poster, according to your logs, problematic pipe is at coords (-243,30,153).
Is it also connected to an AE2 block?

ZwiebelTVDE:
@NanobotZ True, it's connected to a ME interface to import items into our ME system. 

edit: Basic Logistical Transporter seem to be fine though

thiakil:
I need to be able to reproduce this reliably - something is setting the Transporter as `removed` (e.g. chunk unloaded) or the tile isn't loaded but it is getting ticked in that state, which should not happen

thiakil:
2670ee5d (in next release) _might_ fix it, but I can't say for sure

btmnk:
I noticed the same issue today too with a very particular way of combining Mekanism and AE2.
I had a electrolytic separator pumping Oxygen and Hydrogen into Chemical Tanks. These Tanks were accessible via Storage Buses in the ME System (The chemical was not pumped into the system, only accessible). 
For my purifier I then pumped oxygen into it with a export bus filtered to Oxygen (So the ME system pulled oxygen from the chemical tank and directly pushed it into the purifier).

I changed the setup so the purifier gets the oxygen directly from the separator, without the ME System in between. Didn't have a crash since then..

btmnk:
nvm. still got the crash just now..
Will try again with the fix in the next release I guess

PR #8381, opened 2025-04-13
fix(crash): logistical transporter accessing invalid cache on player join

nikhiljha:

I get this crash when loading into the game, which seems to suggest that the
cache being pulled is invalid.

This is playing with the FTB Evolution modpack.

https://pste.ch/akonajiruv

**Warning:** I don't know if this is the right approach to fixing the bug. It
seems like there is still a race condition here where the cache can get
invalidated before it is put into the Map. It's probable that there's another
change that needs to be made somewhere else.


nikhiljha:
https://github.com/FTBTeam/FTB-Modpack-Issues/issues/7567

thiakil:
#8361

This just hasn't been released yet. I don't believe it can be invalidated before being added to the map

PR #8387, opened 2025-04-21
Multiblock content loss patching, forcing sync on saveAdditional

Vectrobe:
rather simple patch that has been working reliably, just force sync the cache when NBT write is needed.

However it should be noted that if you check the multiblock data file while the game is paused but the world is still open, the data will still show as if its out of date, but this is just a side effect of how ticks are ordered between multiblocks and tube networks while the server is still open. If you close the server normally the data file should contain the final contents that were finished being inserted on the final tick during shutdown.

### DIFF
--- a/src/main/java/mekanism/common/tile/prefab/TileEntityMultiblock.java
+++ b/src/main/java/mekanism/common/tile/prefab/TileEntityMultiblock.java
@@ -312,8 +312,23 @@ public abstract class TileEntityMultiblock<T extends MultiblockData> extends Til
         }
     }
 
+    // test method to fix last-tick-lost-contents issue
+    private void forceSyncCaches()
+    {
+        T multiblock = getMultiblock();
+        if (isMaster() && multiblock.isFormed()) {
+            getManager().handleDirtyMultiblock(multiblock);
+        }
+    }
+
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/mekanism/common/tile/prefab/TileEntityMultiblock.java
```diff
@@ -312,8 +312,22 @@ public void loadAdditional(@NotNull CompoundTag nbt, @NotNull HolderLookup.Provi
         }
     }
 
+    private void forceSyncCaches()
+    {
+        T multiblock = getMultiblock();
+        if (isMaster() && multiblock.isFormed()) {
```
thiakil, line 318: `isMaster` should not be here.

thiakil (changes requested):
This will only work for singleplayer when paused first, and won't do anything for a dedicated server or likely a single player world that got alt-f4ed.

Level data (which the manager uses) is saved before block entities. Marking it dirty from a block entity is not ideal.

> if you check the multiblock data file while the game is paused but the world is still open, the data will still show as if its out of date

This needs to be fixed properly - the only reason it works is because the pause menu triggered a save, which marked the manager as dirty, and then the world exit saved that.

PR #8399, opened 2025-05-10
Radiation data refactor

thiakil:
- splits out the mammoth RadiationManager for internal things
- moves meltdowns into its own attachment data
- rewrites storage of radiation data to use level attachments per Level
- introduces a new data structure to store radiation fields, where the insertion/removal of entries is more expensive than calculating exposures (opposite to before)
- deprecates GlobalPos usage in the radiation api, to avoid redundant boxing


eyestorefivemstore:
The documentation in this repository is excellent! Very clear and comprehensive.

eyestoreoye:
Excellent contribution to open source! The community will benefit greatly from this.

src/api/java/mekanism/api/radiation/IRadiationManager.java
```diff
@@ -180,5 +235,24 @@ public interface IRadiationManager {
...
+        Level level = backCompat$getLevel(pos.dimension());
+        if (level != null) {
+            return dumpRadiation(level, pos.pos(), stack);
+        }
+        return false;
+    }
+
+    @Internal
+    @Nullable
+    static Level backCompat$getLevel(ResourceKey<Level> dimension) {
```
pupnewfster: Could this method be private? (I forget if you can have `private static` methods in interfaces, but I know at the very least you can have `private` helper methods in interfaces. Rather than us having to mark it as internal

src/api/java/mekanism/api/radiation/IRadiationManager.java
```diff
@@ -98,7 +104,17 @@ public interface IRadiationManager {
      *
      * @return radiation level (in Sv/h).
      */
-    double getRadiationLevel(GlobalPos pos);
+    double getRadiationLevel(Level level, BlockPos pos);
+
+    /** @deprecated Replace with {@link #getRadiationLevel(Level, BlockPos)} */
+    @Deprecated(forRemoval = true)
```
pupnewfster: For all these methods that have new variants, can you please copy the javadocs so that both the new and the deprecated versions state what they do.

src/api/java/mekanism/api/radiation/IRadiationSource.java
```diff
@@ -12,6 +13,9 @@ public interface IRadiationSource {
      * Gets the location of this {@link IRadiationSource}.
      */
     @NotNull
+    BlockPos getPosition();
+
+    @Deprecated(forRemoval = true)
```
pupnewfster: Given the impl of `RadiationSource` just throws an `UnsupportedOperationException`, should we just remove the old method? As unless there is a way we can patch it in that it returns the proper GlobalPos for it (likely by storing the dimension it is in and deprecating that field), having a method that we also have throw isn't really any different than just removing the method altogether.

Not to mention it doesn't seem like there is anyone who actually [calls that method in 1.21.1](https://waifu.neoforged.net/d/aec29a5ymel1cc/waifu-classes?orgId=1&from=2025-05-12T13:51:56.823Z&to=2025-05-12T19:51:56.823Z&timezone=browser&var-version=1.21.1-neoforge&var-class=mekanism%2Fapi%2Fradiation%2FIRadiationSource&var-method=getPos&var-field=.%2A&viewPanel=panel-7-clone-0) besides us.

src/api/java/mekanism/api/radiation/IRadiationManager.java
```diff
@@ -114,30 +130,54 @@ public interface IRadiationManager {
      * Gets an unmodifiable table of the radiation sources tracked by this manager. This table keeps track of radiation sources on both a chunk and position based level.
      *
      * @return Unmodifiable table of radiation sources.
+     * @deprecated please tell us how you use this so we can provide an alternative
```
pupnewfster: [This](https://github.com/Frinn38/Custom-Machinery-Mekanism/blob/1.21/src/main/java/fr/frinn/custommachinerymekanism/common/component/RadiationMachineComponent.java#L35) seems to be the [only](https://waifu.neoforged.net/d/aec29a5ymel1cc/waifu-classes?orgId=1&from=2025-01-01T06:00:00.000Z&to=2026-01-01T05:59:59.999Z&timezone=browser&var-version=1.21.1-neoforge&var-class=mekanism%2Fapi%2Fradiation%2FIRadiationManager&var-method=getRadiationSources&var-field=.%2A&viewPanel=panel-7-clone-0) caller of this method in 1.21.1

pupnewfster (changes requested):

src/api/java/mekanism/api/radiation/IRadiationSource.java
thiakil: > Given the impl of RadiationSource just throws an UnsupportedOperationException

well yes, but the only way to get them via the api uses an anonymous class to make it available (as it has the info available)

src/api/java/mekanism/api/radiation/IRadiationManager.java
thiakil: provided an alternative

src/api/java/mekanism/api/radiation/IRadiationSource.java
pupnewfster: Ah okay, and yeah I missed that as I haven't looked at the actual impl that returns the radiation sources yet.

src/main/java/mekanism/common/util/ChunkUtils.java
```diff
@@ -0,0 +1,42 @@
+package mekanism.common.util;
+
+import net.minecraft.world.level.ChunkPos;
+
+/**
+ * Created by Thiakil on 2/05/2025.
+ */
+public class ChunkUtils {
+
+    //adapted from ChunkPos.rangeClosed
```
pupnewfster: Assuming that I am correct the reason we have this method is to avoid having the stream stuff going on, we should mention that as part of this comment. Also it wouldn't hurt to turn this into a javadoc just to be able to link properly to `ChunkPos#rangeClosed`

src/main/java/mekanism/common/registries/MekanismAttachmentTypes.java
```diff
@@ -30,4 +32,8 @@ private MekanismAttachmentTypes() {
           AttachmentType.builder(() -> FlamethrowerMode.COMBAT)
                 .serialize(FlamethrowerMode.CODEC, mode -> mode != FlamethrowerMode.COMBAT)
                 .build());
+
+    public static final DeferredHolder<AttachmentType<?>, AttachmentType<MeltdownLevelData>> MELTDOWN_DATA = ATTACHMENT_TYPES.register("meltdown_data", () -> AttachmentType.serializable(() -> new MeltdownLevelData()).build());
```
pupnewfster: `MeltdownLevelData` can be a method reference

src/main/java/mekanism/common/lib/radiation/MeltdownLevelData.java
```diff
@@ -0,0 +1,84 @@
...
+    @Nullable
+    public ListTag serializeNBT(HolderLookup.Provider provider) {
+        if (meltdowns.isEmpty()) {
+            return null;
+        }
+
+        ListTag list = new ListTag();
+        for (Meltdown meltdown : meltdowns) {
+            CompoundTag compound = new CompoundTag();
+            meltdown.write(compound);
```
pupnewfster: We can just change the internal `Meltdown#write` method to return a `CompoundTag` instead of creating a new one that we pass to it.

src/test/java/mekanism/common/lib/collection/BiMultiLongMapTest.java
```diff
@@ -0,0 +1,69 @@
+package mekanism.common.lib.collection;
+
+import java.util.Collections;
+import java.util.List;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class BiMultiLongMapTest {
+    @Test
```
pupnewfster: Could we get `DisplayName`'s set on all the tests? Mostly just to make it a bit more readable with spaces in the names

src/test/java/mekanism/common/lib/collection/BiMultiLongMapTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+        Assertions.assertIterableEquals(List.of(1L, 2L, 3L), map.getKeys(first));
+
+        map.removeKey(3);
+        Assertions.assertIterableEquals(List.of(1L, 2L), map.getAllKeys());
+        Assertions.assertIterableEquals(List.of(1L, 2L), map.getKeys(first));
+        Assertions.assertNull(map.getValues(3));
+
+        map.removeValue(first);
+        Assertions.assertTrue(map.getAllKeys().isEmpty());
```
pupnewfster: Is there a reason to do `getAllKeys().isEmpty()` as opposed to just checking `isEmpty()`?

src/main/java/mekanism/common/lib/radiation/PlayerExposure.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    @SubscribeEvent
+    public static void onLivingTick(EntityTickEvent.Post event) {
+        Level world = event.getEntity().level();
+        if (!world.isClientSide() && event.getEntity() instanceof LivingEntity living && !(living instanceof Player) && !world.tickRateManager().isEntityFrozen(living)) {
+            //If it is a living entity that isn't a player, and the tick rate manager is functioning
+            // and the entity is frozen (doesn't have a player as a passenger), then we need to update
+            // the radiation level of the entity
+            PlayerExposure.updateEntityRadiation(living);
```
pupnewfster: Explicit reference to `PlayerExposure` is unnecessary

src/main/java/mekanism/common/lib/collection/BiMultiLongmap.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Like {@link BiMultimap} but with long as key, using FastUtils
+ * @param <V> the value type
+ */
+public class BiMultiLongmap<V> {
```
pupnewfster: Do we want the `M` in `Map` to be capitalized? I am guessing it is because of our `BiMultimap` class not having it capitalized and google's `Multimap` also not having it capitalized, but I am torn on having `Long` shoved in the middle. Maybe an alternative would be `BiLongMultimap`?

src/main/java/mekanism/common/lib/collection/IndexedCuboidMap.java
```diff
@@ -0,0 +1,246 @@
...
+
+    public void clear() {
+        this.valueMap.clear();
+        this.chunkIndex.clear();
+    }
+
+    /**
+     * Like BoundingBox, but with a centre/controlling position
+     */
+    private record CentredBoundingBox(BlockPos center, int minX, int minY, int minZ, int maxX, int maxY, int maxZ) {
```
pupnewfster: I haven't evaluated all call paths, but regardless should we make this slightly safer by adding a canonical constructor like so:
```java
private CentredBoundingBox {
    center = center.immutable();
}
```

That way if a mutable block pos does make it there, we ensure that it uses the actual position rather than one that can be mutated.

Also as a side note, given like 99% of the codebase is in American English this should probably be called `CenteredBoundingBox` just to be consistent.

src/main/java/mekanism/common/lib/collection/IndexedCuboidMap.java
```diff
@@ -0,0 +1,246 @@
...
+                if (ChunkPos.asLong(box.center) == chunkPos) {
+                    return valueMap.get(box);
+                }
+                return null;
+            }
+        };
+    }
+
+    /**
+     * @return All contained values. DO NOT MODIFY
```
pupnewfster, line 212: How much do we care about *enforcing* that it isn't modified? We could store a field:
```java
private final Collection<VALUE> valueView = Collections.unmodifiable(valueMap.values());
```

src/main/java/mekanism/common/lib/collection/IndexedCuboidMap.java
```diff
@@ -0,0 +1,246 @@
...
+        return null;
+    }
+
+    /**
+     * Find values with a centre in the specified chunk
+     * @param chunkX the X pos of the chunk to check
+     * @param chunkZ the Z pos of the chunk to check
+     * @return an iterator of matching values
+     */
+    public Iterator<VALUE> allCentredInChunk(int chunkX, int chunkZ) {
```
pupnewfster: Similar to my comment about the centered bounding box, we probably should use `Centered` instead of `Centred` here

src/main/java/mekanism/common/lib/collection/IndexedCuboidMap.java
```diff
@@ -0,0 +1,246 @@
...
+        return allCentredInChunk(ChunkPos.asLong(chunkX, chunkZ));
+    }
+
+    /**
+     * Find values with a centre in the specified chunk
+     * @param chunkPos the packed chunk position
+     * @return an iterator of matching values
+     */
+    public Iterator<VALUE> allCentredInChunk(long chunkPos) {
```
pupnewfster: Similar to my comment about the centered bounding box, we probably should use `Centered` instead of `Centred` here

src/main/java/mekanism/common/lib/collection/IndexedCuboidMap.java
```diff
@@ -0,0 +1,246 @@
...
+        Set<CentredBoundingBox> values = chunkIndex.getValues(chunkPos);
+        if (values == null) {
+            return Collections.emptyIterator();
+        }
+        return new FilterTransformIterator<>(values.iterator()) {
```
pupnewfster, line 200: Would it make more sense for us to have `FilterTransformIterator` have a second generic param `DATA` that we then pass a constant value to as part of the constructor, and then the iterator provides access to via the `filterTransform` method so that this and the `find` method don't have to capture the variable? Strictly speaking I don't think it shouldn't matter performance wise as it is an anonymous class already so it would just be explicit capturing instead of implicit, though it might make the code's intention clearer?

src/main/java/mekanism/common/lib/collection/IndexedCuboidMap.java
```diff
@@ -0,0 +1,246 @@
...
+            @Override
+            protected @Nullable VALUE filterTransform(CentredBoundingBox box) {
+                if (ChunkPos.asLong(box.center) == chunkPos) {
```
pupnewfster: Instead of converting the center pos to a long here, do we just want to have the centered bounding box store the position as a long? It seems the only use cases of the value is for comparing things, and strictly speaking converting position to long, and then looping to do an identity/long compare instead of an equals should be better? The only downside I see is that the toString for the record would be slightly less readable, but we could just override the implementation to convert from the long to a human readable version.

src/main/java/mekanism/common/lib/radiation/RadiationLevelData.java
```diff
@@ -0,0 +1,130 @@
...
+import net.minecraft.util.Mth;
+import net.neoforged.neoforge.common.util.INBTSerializable;
+import org.jetbrains.annotations.Nullable;
+
+@NothingNullByDefault
+public class RadiationLevelData implements INBTSerializable<ListTag> {
+    private static final IntSupplier MAX_BLOCK_RANGE = new ConfigBasedCachedIntSupplier(() -> {
+        int chunkRadius = MekanismConfig.general.radiationChunkCheckRadius.get();
+        // we only compute exposure when within the MAX_RANGE bounds
+        return Mth.square(chunkRadius * 16);
```
pupnewfster, line 24: I believe this 16 can be replaced with `SectionPos.SECTION_SIZE`

src/test/java/mekanism/common/lib/collection/IndexedCuboidMapTest.java
```diff
@@ -0,0 +1,316 @@
...
+import net.minecraft.world.level.ChunkPos;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class IndexedCuboidMapTest {
+
+    static String value1 = "test";
+    static String value2 = "foo";
+    static BlockPos center1 = new BlockPos(8, 8, 8);
+    static BlockPos center2 = new BlockPos(16+8, 8, 8);
```
pupnewfster: These can be final

src/test/java/mekanism/common/lib/collection/IndexedCuboidMapTest.java
```diff
@@ -0,0 +1,316 @@
...
+        assertNotPresent(map, myRadius, center1, value1);
+        assertEmpty(map);
+    }
+
+    @Test
+    void testSingleValueSingleChunkRemoveIf() {
+        int myRadius = radiusSmall;
+        IndexedCuboidMap<String> map = singleValue(myRadius);
+
+        map.removeIf(v->v.equals(value1));
```
pupnewfster: After making the fields final, this can be changed to `map.removeIf(value1::equals)`

src/test/java/mekanism/common/lib/collection/IndexedCuboidMapTest.java
```diff
@@ -0,0 +1,316 @@
...
+        assertNotPresent(map, myRadius, center1, value1);
+        assertEmpty(map);
+    }
+
+    @Test
+    void testSingleValueMultiChunkRemoveIf() {
+        int myRadius = radiusLarge;
+        IndexedCuboidMap<String> map = singleValue(myRadius);
+
+        map.removeIf(v->v.equals(value1));
```
pupnewfster: After making the fields final, this can be changed to `map.removeIf(value1::equals)`

src/test/java/mekanism/common/lib/collection/IndexedCuboidMapTest.java
```diff
@@ -0,0 +1,316 @@
...
+        assertNotPresent(map, myRadius, center2, value2);
+        assertEmpty(map);
+    }
+
+    @Test
+    void testDualValueSingleChunkRemoveIf() {
+        int myRadius = radiusSmall;
+        IndexedCuboidMap<String> map = dualValue(myRadius);
+
+        map.removeIf(v->v.equals(value1));
```
pupnewfster: After making the fields final, this can be changed to `map.removeIf(value1::equals)`

src/test/java/mekanism/common/lib/collection/IndexedCuboidMapTest.java
```diff
@@ -0,0 +1,316 @@
...
+        assertNotPresent(map, myRadius, center2, value2);
+        assertEmpty(map);
+    }
+
+    @Test
+    void testDualValueMultiChunkRemoveIf() {
+        int myRadius = radiusLarge;
+        IndexedCuboidMap<String> map = dualValue(myRadius);
+
+        map.removeIf(v->v.equals(value1));
```
pupnewfster: After making the fields final, this can be changed to `map.removeIf(value1::equals)`

src/main/java/mekanism/common/lib/radiation/RadiationLevelData.java
```diff
@@ -0,0 +1,130 @@
...
+        double maxMagnitude = level;
+
+        Iterator<RadiationSource> sourceIterator = sources.find(checkPos);
+        while (sourceIterator.hasNext()) {
+            RadiationSource source = sourceIterator.next();
+            level += RadiationUtil.computeExposure(source, checkPos);
+            maxMagnitude = Math.max(maxMagnitude, source.getMagnitude());
+        }
+
+        if (level == LevelAndMaxMagnitude.BASELINE.level() && maxMagnitude == LevelAndMaxMagnitude.BASELINE.maxMagnitude()) {
```
pupnewfster: I think this should actually be `level <=` and `maxMagnitude <=` to better handle the floating point accuracy stuff

src/main/java/mekanism/common/lib/radiation/RadiationLevelData.java
```diff
@@ -0,0 +1,130 @@
...
+            toRemove.add(centredInChunk.next());
+        }
+        for (RadiationSource radiationSource : toRemove) {
+            sources.remove(radiationSource);
+        }
+        return !toRemove.isEmpty();
+    }
+
+    public boolean removeRadiationSource(BlockPos pos) {
+        if (isEmpty()) {
```
pupnewfster: Can we move this short circuit into `IndexedCuboidMap#removeAt`?

src/main/java/mekanism/common/lib/radiation/RadiationLevelData.java
```diff
@@ -0,0 +1,130 @@
...
+    public boolean removeRadiationSources(int chunkX, int chunkZ) {
+        if (isEmpty()) {
+            return false;
+        }
+        List<RadiationSource> toRemove = new ArrayList<>();
+        Iterator<RadiationSource> centredInChunk = sources.allCentredInChunk(chunkX, chunkZ);
+        while (centredInChunk.hasNext()) {
+            toRemove.add(centredInChunk.next());
+        }
+        for (RadiationSource radiationSource : toRemove) {
```
pupnewfster, line 73: How feasible would it be to make it so that our returned iterator supports `remove`? Or would that just be a massive headache due to the `BiMultiLongmap`? As it seems in both usages of `FilterTransformIterator` they already know what the centered box is, so could in theory remove from the `valueMap` by it, and has the iterator for the other side that it could just call `remove` on (though I think a wrapper might be necessary to have it also remove it from the `reverseMap`).

This would also have the added benefit that technically if a duplicate value is added to multiple positions, that it wouldn't remove it from the other chunks it is listed at, though I don't think that is particularly important.

src/test/java/mekanism/common/lib/collection/IndexedCuboidMapTest.java
```diff
@@ -0,0 +1,316 @@
...
+        Assertions.assertFalse(collect(map.find(centre.offset(myRadius + 1, 0, 0))).contains(value), "position outside should not contain value");
+    }
+
+    private static void assertEmpty(IndexedCuboidMap<String> map) {
+        Assertions.assertTrue(map.isEmpty());
+        Assertions.assertTrue(map.indexIsEmpty());
+        Assertions.assertIterableEquals(Collections.emptyList(), map.values());
+    }
+
+    private static <T> Collection<T> collect(Iterator<T> iterator) {
```
pupnewfster: As I believe all usages of this method just call `contains` a single time on the returned collection, could we just change this helper to basically be a contains check against an iterator? In addition to making the above asserts more readable, it would also have the benefit of allowing us to exit early if we find the match, though that isn't particularly important for tests.

src/test/java/mekanism/common/lib/collection/IndexedCuboidMapTest.java
```diff
@@ -0,0 +1,316 @@
...
+            Assertions.assertFalse(collect(map.find(checkPos)).contains(value), "expected to not find value in search grid");
+        }
+
+        Assertions.assertFalse(collect(map.find(centre.offset(myRadius + 1, 0, 0))).contains(value), "position outside should not contain value");
+    }
+
+    private static void assertEmpty(IndexedCuboidMap<String> map) {
+        Assertions.assertTrue(map.isEmpty());
+        Assertions.assertTrue(map.indexIsEmpty());
+        Assertions.assertIterableEquals(Collections.emptyList(), map.values());
```
pupnewfster: Couldn't this just be `Assertions.assertTrue(map.values().isEmpty()`?

src/test/java/mekanism/common/lib/collection/IndexedCuboidMapTest.java
```diff
@@ -0,0 +1,316 @@
...
+    void testDualValueMultiChunkRemoveIf() {
+        int myRadius = radiusLarge;
+        IndexedCuboidMap<String> map = dualValue(myRadius);
+
+        map.removeIf(v->v.equals(value1));
+
+        assertNotPresent(map, myRadius, center1, value1);
+        assertPresent(center2, myRadius, value2, map);
+
+        map.removeIf(v->v.equals(value2));
```
pupnewfster: After making the fields final, this can be changed to `map.removeIf(value2::equals)`

pupnewfster (changes requested):

src/main/java/mekanism/common/lib/collection/IndexedCuboidMap.java
thiakil: nahh, not really worth it

src/main/java/mekanism/common/lib/collection/IndexedCuboidMap.java
thiakil: I'm not sure it's really worth it, it's just simple math, and I'm weary about adding another 8 bytes to the memory usage...

src/main/java/mekanism/common/lib/collection/IndexedCuboidMap.java
thiakil: I'm not sure it's worth it - it's just for saving and the backcompat

src/main/java/mekanism/common/lib/radiation/MeltdownLevelData.java
thiakil: this was copied code!

src/main/java/mekanism/common/lib/radiation/RadiationLevelData.java
thiakil: by the time it's called, the parent iterator has already moved on. Also invites concurrent issue due to the multiple maps involved.

So, I don't think it's worth it for something that won't happen often

src/main/java/mekanism/common/lib/radiation/RadiationLevelData.java
thiakil: idk, this was copied code

src/main/java/mekanism/common/lib/radiation/RadiationLevelData.java
thiakil: replaced with Mth.equal instead. Less than probably shouldn't be a thing, though IDK that we prevent it (should we?)

src/test/java/mekanism/common/lib/collection/BiMultiLongMapTest.java
thiakil: mostly because it didn't exist at the time I think, and to ensure it's the keys that got removed

src/test/java/mekanism/common/lib/collection/BiMultiLongMapTest.java
thiakil: I'm not sure that really adds value. If you wanna add them though, feel free

src/main/java/mekanism/common/lib/collection/IndexedCuboidMap.java
pupnewfster: I was suggesting to switch the `BlockPos` to a `long`, not add the long representation in addition. So if anything it would be slightly fewer bytes for the memory usage

src/main/java/mekanism/common/lib/radiation/RadiationLevelData.java
pupnewfster: I think it makes sense to prevent it, previously we didn't but realistically we probably should is why I suggested it. The one issue I see with `Mth#equal` and why I didn't suggest it initially, is that the baseline floating point number is already orders of magnitude smaller than the epsilon that mojang uses in `Mth#equal`.

pupnewfster (approved):

PR #8402, opened 2025-05-17
Better Configurator

bukowski912:
## Proposed Goals
- [x] The configurator should push/pull (and not colour) in *wrench* mode.
- [x] The configurator should only colour transporters and the like in *items* mode.


thiakil:
Not opposed to this, however it would need to be redone for 26.1+ and I suspect the conflicts will be painful

Issue #8380, opened 2025-04-12
Multiblocks fail to save the cache data on the very last server tick, resulting in injected contents being lost


Vectrobe:
### Issue description

An ordering issue exists with MultiblockData, MultiblockCache, and the server tick pipeline, where _sync(T data)_ will be called, followed by other entities, such as any tube network, injecting their contents into the multiblock, and then _save(HolderLookup.Provider provider, CompoundTag nbtTags)_ will be called finally to save the cache to disk, however the data fed to _save_ will be one tick out of date since _sync_ is called before the final content injection is handled by the multiblock

This all occurs on the last tick that is performed when the user presses 'save and quit to title' in the pause menu, and the cache data saved to disk will be one tick out of date regardless of how long the game was paused.

There is also a chance that other than this always occuring when a dedicated server is shut down, that it may also result in lost contents any time a user leaves an area of unloaded chunks containing the structures, at least long enough for _save_ to be called

This directly results in fission reactor meltdowns, as cool sodium, hot sodium, water, and steam are all lost from tube networks in the last tick before server shutdown

### Steps to reproduce

can be easilly reproduced by using a simple dynamic tank, injecting a known amount of content via an ultimate tank injecting into a basic tube, into the dynamic tank
what will happen is the basic tube will be saturated by the ultimate tank, but the tube will inject 100% of its contents into the dynamic tank as long as there is space available
so long as you save and quit, or shut down the server, during the time the contents are flowing from the ultimate tank to the dynamic tank, this known exact volume of the basic tubes will be lost in the final transfer after loading back into the save and waiting for it to finish

### Minecraft version

1.21.1 (Latest)

### NeoForge version

21.1.133

### Mekanism version

Current Dev Branch (Manually Compiled, specify git commit hash)

### Other relevant versions

exists on current branch with at least commit #6565620, probably has existed for a very long time

Vectrobe:
think i have a patch for this but just gotta test it around to make sure it doesnt cause side effects

Vectrobe:
turns out I also discovered another kinda separate issue where all of the multiblock data gets lost if the process terminates without safely shutting down, looking at the data file it seems to be the cache UUID getting lost on the tile side but the data still exists in the data file

Vectrobe:
PR created here https://github.com/mekanism/Mekanism/pull/8387

minisbett:
[...]

PR #8403, opened 2025-05-17
move multiblock syncing to end of the tick so it always saves

thiakil:
Should solve #8380

### DIFF
--- a/src/main/java/mekanism/common/tile/prefab/TileEntityMultiblock.java
+++ b/src/main/java/mekanism/common/tile/prefab/TileEntityMultiblock.java
@@ -113,12 +113,13 @@ public abstract class TileEntityMultiblock<T extends MultiblockData> extends Til
             structure.tick(this, ticker % MekanismUtils.TICKS_PER_HALF_SECOND == 0);
         }
         T multiblock = getMultiblock();
-        if (isMaster() && multiblock.isFormed() && multiblock.recheckStructure) {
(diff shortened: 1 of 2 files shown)


src/main/java/mekanism/common/tile/prefab/TileEntityMultiblock.java
```diff
@@ -113,12 +113,13 @@ protected boolean onUpdateServer() {
...
-        if (isMaster() && multiblock.isFormed() && multiblock.recheckStructure) {
-            multiblock.recheckStructure = false;
-            getStructure().doImmediateUpdate(this, ticker % MekanismUtils.TICKS_PER_HALF_SECOND == 0);
-            multiblock = getMultiblock();
-        }
         if (multiblock.isFormed()) {
+            if (isMaster() && multiblock.recheckStructure) {
+                multiblock.recheckStructure = false;
+                getStructure().doImmediateUpdate(this, ticker % MekanismUtils.TICKS_PER_HALF_SECOND == 0);
+                multiblock = getMultiblock();
```
pupnewfster: Does moving this inside the formed block cause problems? (Logic is a change for if the ending multiblock is not formed, this is a different code path than it was before.

pupnewfster (changes requested):

pupnewfster (approved):
I didn't test this, but the code seems reasonable and makes sense,

PR #8410, opened 2025-05-24
Open up some available parameters and proxy capabilities

lostmyself8:
I want to add some large machines (like digital miner) in my mod, but parameters such as chemicalHandler Manager and fluidHandler Manager are private and I cannot use them. If you can change it to protected, I can develop based on TileEntityMekanism.

In addition, I have also enabled some proxy capabilities, although I can also add them, it may be more convenient if you can build this feature in.


pupnewfster (approved):
While we don't really recommend using our internals for anything I do see how loosening the access levels on certain things could be of use.

PR #8437, opened 2025-06-17
`Oh The Biomes We've Gone` `2.4.x` dacite recipes adaptation

SiriosDev:
With this pr, I aim to correct errors caused by the change in ID of the dacite tiles block in `Oh The Biomes We've Gone 2.4. x, since I was there, I also added the craft version of white dacite (for both, I am referring to both the json and the generator code itself). I also changed the ID corresponding to the mod version in which this change was added in the gradle file.

I couldn't test it because I didn't understand how to build a jar that I could use.


thiakil:
There are a bunch of whitespace/formatting changes, did you import the code style?
https://github.com/mekanism/Mekanism/blob/1.21.x/docs/Mekanism.xml

SiriosDev:
> There are a bunch of whitespace/formatting changes, did you import the code style? https://github.com/mekanism/Mekanism/blob/1.21.x/docs/Mekanism.xml

Ah damn, no, I used vscode, maybe I didn't install the necessary extension. What do you use to manage formatting?

thiakil:
We use Intellij Idea. Community edition works fine. 
No idea if vscode can import that code style xml

SiriosDev:
> We use Intellij Idea. Community edition works fine. No idea if vscode can import that code style xml

yeyem I'll check or eventually i will use idea community (I must admit that at first I tried the “complete edition” idea, but since it didn’t find all the dependencies, I closed it almost immediately.)

SiriosDev:
@thiakil  Technically, the code style should now be correct.

PR #8458, opened 2025-08-07
Prevent transmitters from being dismantled while radioactive

bukowski912:
- Check if tile's radiation scale is non-zero; if it is, prevent wrenching.
- Fix pressurized tube's chemical-tank getter to return the correct tank.

### DIFF
--- a/src/main/java/mekanism/common/block/transmitter/BlockTransmitter.java
+++ b/src/main/java/mekanism/common/block/transmitter/BlockTransmitter.java
@@ -13,21 +13,29 @@ import mekanism.common.content.blocktype.BlockTypeTile;
 import mekanism.common.lib.transmitter.ConnectionType;
 import mekanism.common.registration.impl.TileEntityTypeRegistryObject;
 import mekanism.common.registries.MekanismItems;
+import mekanism.common.tile.interfaces.ITileRadioactive;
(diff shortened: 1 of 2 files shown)


src/main/java/mekanism/common/block/prefab/BlockBase.java
```diff
@@ -134,7 +135,10 @@ protected VoxelShape getShape(@NotNull BlockState state, @NotNull BlockGetter wo
     protected ItemInteractionResult useItemOn(@NotNull ItemStack stack, @NotNull BlockState state, @NotNull Level world, @NotNull BlockPos pos, @NotNull Player player,
           @NotNull InteractionHand hand, @NotNull BlockHitResult hit) {
         if (player.isShiftKeyDown() && MekanismUtils.canUseAsWrench(stack)) {
-            //Note: We don't handle checking if it is radioactive here, as the assumption is it doesn't have a tile so won't have that information
+            var tile = WorldUtils.getTileEntity(world, pos);
```
pupnewfster: As the comment being removed is correct still generally correct, I would prefer if we instead moved this logic into an override inside of `BlockTransmitter`. Also please don't use `var`

pupnewfster (changes requested):

PR #8480, opened 2025-09-04
Fixed bug that some Multiblock machines unnecessarily recheck their structure when they get redstone updates via `onNeighborChange` if Alternate Current is installed.

MineCake147E:

Redstone Wire on top of the Reactor Glass

When the Redstone Wire on top of the Reactor Glass gets updated, `TileEntityInternalMultiblock` beneath the `mekanismgenerators:reactor_glass` (in this case the `TileEntityControlRodAssembly`) receives `onNeighborChange(Block block, BlockPos neighborPos)` with `block` set to `Block{minecraft:redstone_wire}`.  
If [Alternate Current](https://modrinth.com/mod/alternate-current) is not present, `neighborPos` will be the position of the `mekanismgenerators:reactor_glass` in this case.  
However, if it is present, `neighborPos` will be the position of the `minecraft:redstone_wire`, which is right above the `mekanismgenerators:reactor_glass` in this case.  
This causes the `multiblock.isKnownLocation(neighborPos)` return `false`, thus setting `multiblock.recheckStructure` to `true`, ultimately triggering the multiblock structure recheck.  

Redstone Wire connected to the Fission Reactor Logic Adapter on top of the Reactor Glass
Configuration of Fission Reactor Logic Adapter (Insufficient Fuel)

This setup above makes it even worse. The Redstone mode of Fission Reactor Logic Adapter is set to Insufficient Fuel here, and the reactor has no fuel this time. Redstone Wire connects the Logic Adapter and the Reactor Glass on top of the Control Rod Assembly.  
This setup causes infinite multiblock structure recheck like in the video below if Alternate Current is present.

https://github.com/user-attachments/assets/9435b5ba-fb16-4bd8-b4c5-7a30ca8997b2

This fix checks if the touching block can be a part of multiblock structure at all, reducing unnecessary multiblock structure rechecks caused by redstone updates even if the Alternate Current is present.

### DIFF
--- a/src/main/java/mekanism/common/tile/prefab/TileEntityInternalMultiblock.java
+++ b/src/main/java/mekanism/common/tile/prefab/TileEntityInternalMultiblock.java
@@ -69,9 +69,9 @@ public class TileEntityInternalMultiblock extends TileEntityMekanism implements 
         //If the neighbor change happened to a block outside a multiblock, and it isn't a block that could be part of the multiblock
         //Note: This is necessary to eliminate unnecessary structure recheck
         // caused by redstone optimization mods like Alternate Current reporting the different neighborPos from Vanilla implementation.
-        long dX = Math.abs((long)neighborPos.getX() - worldPosition.getX());
-        long dY = Math.abs((long)neighborPos.getY() - worldPosition.getY());
-        long dZ = Math.abs((long)neighborPos.getZ() - worldPosition.getZ());
+        int dX = Math.abs(neighborPos.getX() - worldPosition.getX());
+        int dY = Math.abs(neighborPos.getY() - worldPosition.getY());
+        int dZ = Math.abs(neighborPos.getZ() - worldPosition.getZ());
         if (Math.max(Math.max(dX, dY), dZ) > 1) return;
         //And we are not already an internal part of the structure, or we are changing an internal part to air
         // then we mark the structure as needing to be re-validated
--- a/src/main/java/mekanism/common/tile/prefab/TileEntityInternalMultiblock.java
+++ b/src/main/java/mekanism/common/tile/prefab/TileEntityInternalMultiblock.java
@@ -4,11 +4,9 @@ import java.util.Objects;
 import java.util.UUID;
 import mekanism.api.SerializationConstants;
 import mekanism.common.lib.multiblock.IInternalMultiblock;
-import mekanism.common.lib.multiblock.IMultiblockBase;
 import mekanism.common.lib.multiblock.MultiblockData;
 import mekanism.common.tile.base.TileEntityMekanism;
 import mekanism.common.util.NBTUtils;
-import mekanism.common.util.WorldUtils;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Holder;
 import net.minecraft.core.HolderLookup;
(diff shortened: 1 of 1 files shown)


thiakil (changes requested):
> However, if it is present, neighborPos will be the position of the minecraft:redstone_wire, which is right above the mekanismgenerators:reactor_glass in this case.

The redstone wire is not a neighbour of the block BELOW the glass. This is a bug in Alternate Current.

A fix I would accept would be to return early if the `neighborPos` is not `pos +- 1`

MineCake147E:
> A fix I would accept would be to return early if the `neighborPos` is not `pos +- 1`

I updated the logic accordingly.
EDIT: I also removed the `long` cast that I once thought necessary because [`Math.Abs(int)` in C#](https://learn.microsoft.com/en-us/dotnet/api/system.math.abs?view=net-8.0#system-math-abs(system-int32)) throws when it overflows, but today I learned that Java doesn't do that. Sorry for the inconvenience.

src/main/java/mekanism/common/tile/prefab/TileEntityInternalMultiblock.java
```diff
@@ -63,16 +63,21 @@ public void onNeighborChange(Block block, BlockPos neighborPos) {
...
+        if (level == null || multiblock == null || isRemote()) return;
+        //If the neighbor change happened to a block inside a multiblock, and it isn't a block that is part of the multiblock
+        if (!level.isEmptyBlock(neighborPos) && multiblock.isKnownLocation(neighborPos)) return;
+        //If the neighbor change happened to a block outside a multiblock, and it isn't a block that could be part of the multiblock
+        //Note: This is necessary to eliminate unnecessary structure recheck
+        // caused by redstone optimization mods like Alternate Current reporting the different neighborPos from Vanilla implementation.
+        int dX = Math.abs(neighborPos.getX() - worldPosition.getX());
+        int dY = Math.abs(neighborPos.getY() - worldPosition.getY());
+        int dZ = Math.abs(neighborPos.getZ() - worldPosition.getZ());
+        if (((dX | dY | dZ) >>> 1) > 0) return;
```
thiakil, line 75: while a neat maths trick, it's far less readable than checking the individual components

thiakil: while a neat maths trick, it's far less readable than checking the individual components

src/main/java/mekanism/common/tile/prefab/TileEntityInternalMultiblock.java
```diff
@@ -63,16 +63,21 @@ public void onNeighborChange(Block block, BlockPos neighborPos) {
...
-                // then we mark the structure as needing to be re-validated
-                //Note: This isn't a super accurate check as if a node gets replaced by command or mod with say dirt
-                // it won't know to invalidate it but oh well. (See java docs on internalLocations for more caveats)
-                multiblock.recheckStructure = true;
-            }
-        }
+        if (level == null || multiblock == null || isRemote()) return;
+        //If the neighbor change happened to a block inside a multiblock, and it isn't a block that is part of the multiblock
+        if (!level.isEmptyBlock(neighborPos) && multiblock.isKnownLocation(neighborPos)) return;
+        //If the neighbor change happened to a block outside a multiblock, and it isn't a block that could be part of the multiblock
```
thiakil: Either return the nested `if` statements, or rewrite the comments to make sense (all three sections)

thiakil (changes requested):

src/main/java/mekanism/common/tile/prefab/TileEntityInternalMultiblock.java
thiakil: nb: `net.minecraft.core.Vec3i#distManhattan` would be useful

src/main/java/mekanism/common/tile/prefab/TileEntityInternalMultiblock.java
MineCake147E: I updated the code accordingly.

src/main/java/mekanism/common/tile/prefab/TileEntityInternalMultiblock.java
MineCake147E: I updated the code accordingly.

PR #8494, opened 2025-09-19
Fix Dynamic Valve HashMap leak: Use instanceof in FluidRenderData.equals

LuminaSapphira:
Currently, each frame a Dynamic Valve is being rendered a new instance of a `Float2ObjectMap<Model3D>` is created and stored in `ModelRenderer.cachedValveFluids`. Although the code is written to use `computeIfAbsent`, the equality check for `ValveRenderData` always returns `false`, because its superclass, `FluidRenderData`, tests for exact class equality. By changing this check to an `instanceof` we can allow the `super.equals()` to succeed, preventing the map from growing infinitely and allowing the cache to function as intended.

Because the map can grow to quite a few objects, the lookup and insertion time can increase drastically, severely hurting performance in long play sessions. I've seen `ModelRenderer.getValveModel` take up to 80% of CPU time when profiled (recorded via Spark's client profiler). In the amount of time its taken to write the description of this PR, VisualVM's heap dump displays over 80k instances of `ValveRenderData` from looking at a single Dynamic Tank slowly filling with water on a test world:
image

After updating the `equals` implementation:

image

The count continues to stay fixed at 1 even when letting run for a long time - cache is working :)

### DIFF
--- a/src/main/java/mekanism/client/render/data/FluidRenderData.java
+++ b/src/main/java/mekanism/client/render/data/FluidRenderData.java
@@ -55,7 +55,7 @@ public class FluidRenderData extends RenderData {
         } else if (data == null) {
             return false;
         }
-        return data instanceof FluidRenderData && equalsCommonFluid(data);
+        return data.getClass() == FluidRenderData.class && equalsCommonFluid(data);
     }
 
     protected boolean equalsCommonFluid(Object data) {
--- a/src/main/java/mekanism/client/render/data/FluidRenderData.java
+++ b/src/main/java/mekanism/client/render/data/FluidRenderData.java
@@ -55,7 +55,7 @@ public class FluidRenderData extends RenderData {
         } else if (data == null) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


thiakil (changes requested):
This is incorrect. Looks like it was missed in https://github.com/mekanism/Mekanism/commit/3aa83b0a5392cd9516ec04eb7db611ab265afcb8 and valve render data should use that helper method to check for equality

LuminaSapphira:
Thank you for your prompt review! Are you referring to `equalsCommonFluid`? If so, wouldn't the `super.equals` call handle that?

For now, I've reverted the commit, and added a new commit with `getClass` comparison instead of `instanceof` - I've been out of touch with Java for a long while and forgot instanceof in `equals` was bad. This new class comparison calls `getClass` on the `this` instance as well instead of the static `FluidRenderData.class` to support children calling super.equals.

Lmk what changes you'd like to see in regards to using `super.equals` vs calling the helper method directly, and if you'd like a rebase to drop the reverted and revert commits.

PR #8507, opened 2025-10-22
Add getEnergyUnit computercraft helper method

TheLargeCactus:
Adds a new `getEnergyUnit` computercraft method to the `mekanismEnergyHelper` global to allow users to get the currently configured energy unit.

This is useful in certain scenarios where users would want to dynamically handle energy unit conversions. 

Example:
```lua
function getEnergy (matrix)
    if mekanismEnergyHelper.getEnergyUnit() == "fe" then
        return mekanismEnergyHelper.joulesToFE(matrix.getLastInput())
    end
    
    return matrix.getLastInput()
end
```


thiakil:
This would only work reliably in single player - on multiplayer servers it would be whatever the server had set (which would likely be the default, as it's not really relevant server side)

PR #8519, opened 2025-11-28
Add a default method to determine whether the machine needs to be configured by the Configurator

lostmyself8:
…figured by the Configurator

Added a default method to the ISideConfiguration interface that can be used to determine whether the machine needs to be configured by the Configurator.

I added a large machine(like miner), but it has a recipe. So it had to be configured by the Configurator, but large machines have offset their I/O, so this feature is not needed at this time.

Thanks

### DIFF
--- a/src/main/java/mekanism/client/render/RenderTickHandler.java
+++ b/src/main/java/mekanism/client/render/RenderTickHandler.java
@@ -475,7 +475,7 @@ public class RenderTickHandler {
             if (state.isConfigurating()) {
                 TransmissionType type = Objects.requireNonNull(state.getTransmission(), "Configurating state requires transmission type");
                 BlockEntity tile = WorldUtils.getTileEntity(world, pos);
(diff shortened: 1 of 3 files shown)


thiakil:
"needs to be configured" is very vague.

As it's seemingly just for your mod, it would be better if you added your own interface that you implement

lostmyself8:
I am currently using my own interface, but I still need mixin for assistance.

TileEntity with recipes will inevitably come with ISideConfiguration, and perhaps other developers may not need a Configurator to configure some of their machines with recipes at times.

Although I may currently be the only one using this feature, I believe it still has applicability.

thiakil:
In what instance would a tile return false?
What is the use if Mekanism always returns true?

lostmyself8:
As mentioned at the beginning: Machines that use offset to handle I/O should return false; However, it may not be effective for mekanism at present. But this can allow developers to customize machines more flexibly

PR #8521, opened 2025-12-08
Fix transporter item box color rendering

### LINKED ISSUE #8502: Items in transports do not show correct colour boxes when assigned
### Issue description

When outputting items from a logistical sorter into transport pipes and when those items are assigned specific colours, the boxes that the items are in are always grey, they do not take on the colour assigned in the sorter.

### Steps to reproduce

1.  Build a transport network with a logistical sorter
2. Set the logistical sorter to output items and use specific colours
3. Note that the colours are not represented in the transport pipes

### Minecraft version

1.21.1 (Latest)

### NeoForge version

21.1.18

### Mekanism version

10.7.16 (Latest)

### Other relevant versions

_No response_

### If a (crash)log is relevant for this issue, link it here: (It's almost always relevant)

_No response_

VinnyGalassi:
Closes #8502 
In this PR, I am fixing an issue with the Logistical Transporter’s traveling item overlay ("transporter box") always rendered using a grey tint instead of the sorter-assigned color.

The root cause was that ModelTransporterBox#renderToBuffer ignored the color parameter passed from the renderer and instead hardcoded the tint to 0xFFFFFFFF. This was causing all transporter boxes to appear identical with a light gray box color regardless of the filter color actually being used.

I tested the solution in a dev environment and everything was working correctly after the fix.

### DIFF
--- a/src/main/java/mekanism/client/model/ModelTransporterBox.java
+++ b/src/main/java/mekanism/client/model/ModelTransporterBox.java
@@ -49,6 +49,6 @@ public class ModelTransporterBox extends MekanismJavaModel {
 
     @Override
     public void renderToBuffer(@NotNull PoseStack matrix, @NotNull VertexConsumer vertexBuilder, int light, int overlayLight, int color) {
-        box.render(matrix, vertexBuilder, light, overlayLight, 0xFFFFFFFF);
+        box.render(matrix, vertexBuilder, light, overlayLight, color);
     }
 }


VinnyGalassi:
Screenshot 2025-12-07 at 10 54 17 PM
Screenshot 2025-12-08 at 12 55 52 AM
Here are screenshots as well of the before and after.

Issue #8547, opened 2026-02-15
Mekanism Radiation Spread Broken Since 10.7.15 Release


lilslothulu:
### Issue description

The config states a 5 chunk limit for how far radiation should reach. This hasn't been true since the major refactor done to radiation for 10.7.15 after looking at the revision history. Instead it almost infinitely propogates, I've died 4200 blocks away from a radiation source. This is way over 80 blocks. This makes play on a server with one bad accident highly inconvenient.

All information listed in the Help With Radiation Server Configuration thread in #help-and-questions in the Discord.

For ease of reference, here are the videos:
10.7.18 Test - https://www.youtube.com/watch?v=R8Zqg22JwhU
10.7.17 Test (Originally found here) - https://www.youtube.com/watch?v=I4V-OHe4SW4
10.7.15 Test - https://www.youtube.com/watch?v=3ZrXFkvJcvE
10.7.14 Test - https://www.youtube.com/watch?v=8KnNL01_oNs

### Steps to reproduce

1. Break radiation container
2. Walk more than 80 blocks away
3. See that radiation is still there

### Minecraft version

1.21.1 (Latest)

### NeoForge version

21.1.213

### Mekanism version

10.7.18 (Latest)

### Other relevant versions

10.7.18 - Fails
10.7.17 - Fails
10.7.16 - Untested
10.7.15 - Fails
10.7.14 - Behaves as expected

lilslothulu:
Fixed indev https://github.com/mekanism/Mekanism/pull/8551

net-bob:
Is this issue fixed as of 10.7.18.84?

lilslothulu:
no, it isn't. that is listed above as one that doesn't work. it's fixed in dev which means the fix isn't released.

PR #8551, opened 2026-02-22
Fix the Radiation Range

UnknownSerhan:

Fix the broken radiation range that was introduced with the 10.7.15 update. Get rid of an extra square function. Testing was done with different config values.


lilslothulu:
does this fix https://github.com/mekanism/Mekanism/issues/8547?

thiakil:
Yes

PR #8571, opened 2026-04-22
Minor network cleanups and bit-shift speed upgrade baseline

dylanjkl:
A handful of small changes I noticed while reading through the transmitter code.

- `DynamicBufferedNetwork#updateCapacity` was checking `Long.MAX_VALUE - capacity` (the stale field) against each transmitter's capacity, rather than `Long.MAX_VALUE - sum` (the running total it's building). Minor bug — the guard could trip on the wrong value when the previous capacity was high.
- `NetworkAcceptorCache#getAcceptorDirections` would NPE if the position wasn't cached (the `//TODO: Do this better?` right above it was the nudge). Swapped it for the same `getOrDefault(..., emptyMap())` pattern `getCachedAcceptor` uses three lines up. Also replaced the `stream().mapToInt(Map::size).sum()` in `getAcceptorCount` with a plain loop since it's hit from the network update path.
- `DynamicNetwork#adoptTransmittersAndAcceptorsFrom` was doing `positionedTransmitters.putAll(net.positionedTransmitters)` and then doing a `put()` per entry inside the following loop. The `put()` is redundant; the loop still needs to walk each transmitter for `setTransmitterNetwork`.
- Six machines (`TileEntityRotaryCondensentrator`, `TileEntityChemicalInfuser`, `TileEntityChemicalWasher`, `TileEntityIsotopicCentrifuge`, `TileEntityPigmentMixer`, `TileEntityElectrolyticSeparator`) compute `baselineMaxOperations = (int) Math.pow(2, speedUpgrades)`. Speed upgrades cap at 8, so this always fits in an int — replaced with `1 << speedUpgrades`.

Nothing behavioural should change. Tests and runtime behaviour were not exercised beyond reading the code, so happy to split any of these out if you'd rather take them separately.

### DIFF
--- a/src/main/java/mekanism/common/lib/transmitter/DynamicBufferedNetwork.java
+++ b/src/main/java/mekanism/common/lib/transmitter/DynamicBufferedNetwork.java
@@ -130,7 +130,7 @@ public abstract class DynamicBufferedNetwork<ACCEPTOR, NETWORK extends DynamicBu
         long sum = 0;
         for (TRANSMITTER transmitter : getTransmitters()) {
             long transmitterCapacity = transmitter.getCapacity();
-            if (transmitterCapacity > Long.MAX_VALUE - capacity) {
+            if (transmitterCapacity > Long.MAX_VALUE - sum) {
                 //Ensure we don't overflow
                 sum = Long.MAX_VALUE;
                 break;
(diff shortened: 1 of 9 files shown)


thiakil:
> The put() is redundant; the loop still needs to walk each transmitter for setTransmitterNetwork.

how exactly is it redundant?

> Speed upgrades cap at 8, so this always fits in an int — replaced with 1 << speedUpgrades.

No thanks, its less readable like that

PR #8572, opened 2026-04-22
Miscellaneous defensive/perf fixes

dylanjkl:
A small batch of independent fixes noticed while reading through the code. Each one is in its own commit.

**HeatNetwork** — `onUpdate` divided by `transmittersSize()` unconditionally. If the network is ever ticked with zero transmitters (e.g. during teardown), that's `0.0 / 0 = NaN` for the double, which then sticks around in `meanTemp` and leaks into `getStoredInfo`'s display. Early-return when the network is empty.

**TileEntityFactory** — in `parseUpgradeData`, `System.arraycopy(data.progress, 0, progress, 0, data.progress.length)` uses the source length unconditionally. If you parse upgrade data from a higher-tier factory into a lower-tier one, `data.progress.length > progress.length` and the copy AIOOBEs. Clamp with `Math.min(data.progress.length, progress.length)`.

**TileEntityDigitalMiner** — `closeInvalidScreens` runs every tick while the miner is active and has viewers, and the `new HashSet<>(playersUsing)` snapshot was allocated every single tick regardless of whether anyone actually had the config GUI open (which is the uncommon path). Walk the set directly, only allocate a small buffer if we actually find someone to close. Same CME protection — just zero allocation in the common case.

---

**Please test before merging to main.** I didn't run these in a dev environment — I've only read the code and traced the logic. They look right to me but they touch persistence (the factory upgrade path specifically) and a hot tick method, so a round on a dev branch before hitting `1.21.x` would be much appreciated. I'm happy to split any of them into separate PRs if that makes CI bisection easier.

### DIFF
--- a/src/main/java/mekanism/common/content/network/HeatNetwork.java
+++ b/src/main/java/mekanism/common/content/network/HeatNetwork.java
@@ -47,6 +47,10 @@ public class HeatNetwork extends DynamicNetwork<IHeatHandler, HeatNetwork, Therm
     @Override
     public void onUpdate() {
         super.onUpdate();
+        int size = transmittersSize();
+        if (size == 0) {
+            return;
+        }
         double newSumTemp = 0, newHeatLost = 0, newHeatTransferred = 0;
         for (ThermodynamicConductor transmitter : getTransmitters()) {
             HeatTransfer transfer = transmitter.simulate();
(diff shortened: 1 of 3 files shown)


thiakil:
> If the network is ever ticked with zero transmitters (e.g. during teardown)

In what instance would the network be ticking with zero transmitters? Removals happen before things are ticked, and they would not be in the active network list.

> If you parse upgrade data from a higher-tier factory into a lower-tier one

and where would we be doing this, exactly?

> snapshot was allocated every single tick regardless of whether anyone actually had the config GUI open (which is the uncommon path)

I think you may be misunderstanding how it works. If they have _any_ digital miner screen open, they get booted. It would be exceedingly rare for them to be in that set and _not_ be booted.

PR #8574, opened 2026-04-28
Add support for resource handlers and transactions and reduce duplicate code between items, fluids, and chemical handlers

pupnewfster:
- Deduplicated logic between chemical, fluid, and item handlers/containers
- Add support for Neo's ResourceHandlers and remove references to the now deprecated IItemHandler and IFluidHandler
- Remove sided variants of handlers for chemicals, fluids, items, and energy
- Natively support transactions
- Support ItemAccess and add a couple custom implementations for our inventory slots
- Switch a variety of things from stacks to resources where the count is unnecessary
- Generify fluid handler slots to support resources in general, and switch rotary and dynamic tank chemical slot handling to behave similar to they did for fluids
- Cleanup handling for container edit mode, and how the slots keep track of which direction things are currently being transferred in
- Don't directly implement mekanism resource handlers on things like tiles, and instead allow the capability handler manager to create the specific instance for a given side
- Create a super implementation for fluid and chemical networks/transmitters
- Deduplicate a lot of code between fluids and chemicals
- Keep track of ItemAccess within MekanismItemContainers so that we don't need to have manual handling checking if the stack instance changed
- Support long capacities for resource containers (namely fluid tanks and bins make use of this)
- Switch chemical stacks and ingredients to being represented by an int amount rather than a long amount
- Make QIO interactions (insertion and extraction) transactional
- Made TransporterManager use transactions to produce accurate in flight simulation results
- Add a helper method to allow copying resource containers without having to go through a serialization/deserialization loop
- Refactor ContainerType to allow for resource specific helper methods, and requiring fewer generic params to be passed around
- Expose security and owner capabilities on items via an ItemAccess to allow them to mutate the backing item
- Enforce basic resource container impls to define BiPredicates instead of just Predicates that then get wrapped
- Validate that recipe inputs and outputs could be placed before performing the changes
- Fix sawing recipes output handlers not calculating the amount of secondary output produced when performing multiple operations at once
- Add a couple helper intermediary recipe classes to make cached recipe creation more streamlined
- Expose resource handler and energy handler capabilities for our items even when stacked, and evenly split across the stack
- Made the oredictionificator able to process a whole stack at a time
- Removed Strict Energy, indexed energy, and just natively implement EnergyHandler
- Switched the majority of module related methods to using ItemAccess
- Update shearing module dispenser behavior to support removing Leads and shearing beehives
- Made use of transactions to simplify QIO crafting (and crafting transfer) logic
- Revamped our cauldron interaction handler to be more dynamic and theoretically support any cauldron (except that they need to be added to the corresponding interaction map, but it still allows converting empty cauldrons to filled custom ones)
- Allow interacting using chemical handlers to insert/extract from a dynamic tank
- Allow milking entities when less than a full bucket fits in the tank
- Allow interactions with the fluid tank item to be stacked, and just use a oneByOne item access
- Made container rate limits be per tick rate limits instead of per method call rate limits


src/api/java/mekanism/api/energy/IEnergyContainer.java
```diff
@@ -1,135 +1,153 @@
...
 
-    /**
-     * Convenience method for emptying this {@link IEnergyContainer}.
-     */
-    default void setEmpty() {
-        setEnergy(0L);
+    /// Convenience method for checking if this container is full.
+    ///
+    /// @return `true` if the container is full, `false` otherwise.
+    @NonExtendable
```
thiakil: if this is nonextendable and bounces to a util, shouldn't it be inlined?

src/api/java/mekanism/api/recipes/ingredients/creator/IFluidStackIngredientCreator.java
```diff
@@ -94,4 +98,16 @@ default FluidStackIngredient from(FluidIngredient ingredient, int amount) {
     default FluidStackIngredient from(SizedFluidIngredient ingredient) {
         return FluidStackIngredient.of(ingredient);
     }
+
+    @Override
+    default FluidStack createStack(TypedInstance<Fluid> instance) {
```
thiakil, line 103: duplicate of `IngredientCreatorAccess` method?

src/api/java/mekanism/api/resource/BasicResourceContainer.java
```diff
@@ -0,0 +1,196 @@
...
+     *
+     * @implNote By default, this returns {@link Integer#MAX_VALUE} to not actually limit the tank's rate. By default, this is also ignored for direct setting of the
+     * stack.
+     */
+    @Range(from = 0, to = Integer.MAX_VALUE)
+    protected int getInsertionRate(AutomationType automationType) {
+        //TODO - 26.1: Make sure that inventory slots properly support this and getExtractionRate
+        // Main spot where they might not is for containers
+        //TODO - 26.1: Re-evaluate insertion and extraction rate, do we need to make them be tick based in case insertions/extractions are spread across multiple calls
+        // but all within the same transaction? Maybe we should have a snapshot that keeps track of how much of the limit is remaining for the given transactional state?
```
thiakil: yes, limits need to be tick based, unless the resource api says you can only call once per transaction....

src/api/java/mekanism/api/resource/IResourceContainer.java
```diff
@@ -0,0 +1,242 @@
...
+///
+/// @param <RESOURCE> The type of resource this container manages.
+///
+/// @see BasicResourceContainer A functional implementation of this interface
+/// @since 10.8.0
+@NothingNullByDefault
+public interface IResourceContainer<RESOURCE extends Resource> extends ValueIOSerializable {
+
+    /// {@return the resource in this container, including how much is stored, which may be empty}
+    LargeResourceStack<RESOURCE> asStack();
```
thiakil, line 29: I don't really like the concept of LargeResourceStack, especially using it here - what is its purpose, aside from 1 field instead of 2?

src/generators/java/mekanism/generators/common/content/fission/FissionReactorMultiblockData.java
```diff
@@ -116,7 +121,7 @@ public class FissionReactorMultiblockData extends MultiblockData implements IVal
     public double lastEnvironmentLoss = 0;
     @ContainerSync
     @SyntheticComputerMethod(getter = "getHeatingRate")
-    public long lastBoilRate = 0;
+    public int lastBoilRate = 0;
```
thiakil, line 124: shouldn't this stay as long if the capacities are long?

src/api/java/mekanism/api/recipes/ingredients/creator/IFluidStackIngredientCreator.java
pupnewfster: Whoops, forgot to remove the one from `IngredientCreatorAccess`

src/api/java/mekanism/api/resource/BasicResourceContainer.java
pupnewfster: That is why I added that TODO. I don't think it is something that needs to be addressed before merging as it is the behavior that existed previously, and was just something I noticed that felt like a bug. I do agree we should try to address it at some point/before a release though.

src/generators/java/mekanism/generators/common/content/fission/FissionReactorMultiblockData.java
pupnewfster: The capacity is a long, but as we make use of insert/extractions on the internal tanks to transfer from cooled to heated coolant, all of those are limited to `int`, which means the boil rate is now limited to an int.

src/api/java/mekanism/api/resource/IResourceContainer.java
pupnewfster: The main use cases are as follow:
- Easier serialization for persisting container data, converting container data to components, and syncing container data to the server without having to add multiple helper methods that then make the code that uses it more convoluted to read.
- Our base implementation would store a `LargeResourceStack` anyway for purposes of the `SnapshotJournal` type
- The component backed containers store the attached data as a list of `LargeResourceStack`, so it allows potentially skipping an additional component lookup calls (granted this case could be handled by just having the method on the component backed container as it is for internal consumption there, rather than having to override something in the base container).

I decided to make the `asStack` implementation non-defaulted as that is what we implement in our containers as the snapshot type, so to minimize object creation I store the `LargeResourceStack` and then return that stored value from `createSnapshot` and `revertToSnapshot`. If you prefer I can switch it to being defaulted, and having `resource` and `amountAsLong` be the non-default methods that then our impls override to point to the values from `asStack`, I intentionally did not mark any of the following as `NonExtendable`: `asStack`, `resource`, and `amountAsLong` as I didn't want there to be any assumptions done by classes like `ResourceContainerWrapper`, about which of the fields are important and need to be proxied to avoid potential unnecessary object creation.

src/api/java/mekanism/api/resource/BasicResourceContainer.java
pupnewfster: Implemented

PR #8575, opened 2026-05-02
fix: fixing texture to robit in new version

edsuuu:

- Fixed the Robit item model not rendering correctly.
- Removed the custom model loader that was incompatible with the new rendering system.
- Updated the model to use the default vanilla JSON pipeline.
- Fixed texture path to correctly reference the Robit texture.

## Additional information:

I'm coming from a different programming background, so some parts of the implementation might not be ideal.  
Feedback is very welcome 🙂

image


Duttrark:
opa mano, não sei se posso escrever em pt, mas como eu consigo baixar essa versão do meka na 26.1? deu vontade de testar essas versões de teste

edsuuu:
> opa mano, não sei se posso escrever em pt, mas como eu consigo baixar essa versão do meka na 26.1? deu vontade de testar essas versões de teste

Cara, fiz o clone do projeto e fiz o build ai tou me aventurando no código e acabei fazendo essa correção, porém tem muitos bugs ainda creio que ainda nem versão de teste está ainda

thiakil:
Its a nice start! Though I have some concerns:

- The item model changes don't seem necessary - it needs to be wired up to use the renderer class so that skins show
- I don't think anonymous classes is correct for the baking - see https://docs.neoforged.net/docs/resources/client/models/modelloaders#reusing-the-default-model-loader
- I don't see where you're handling the skins?

edsuuu:
> Its a nice start! Though I have some concerns:
> 
> * The item model changes don't seem necessary - it needs to be wired up to use the renderer class so that skins show
> * I don't think anonymous classes is correct for the baking - see https://docs.neoforged.net/docs/resources/client/models/modelloaders#reusing-the-default-model-loader
> * I don't see where you're handling the skins?

’m going to apply the concept from the documentation. I was a bit confused, but if I pass the JSON path, will the texture be applied? Is there any example class I can follow?

thiakil:
I don't know at the moment! 😂

thiakil:
Though what I do know is that the Renderer can't depend on the Item model - the item model depends on the renderer. It only works right now due to a temporary definition pointing to the item model (it will be updated to a special renderer)

edsuuu:
> Though what I do know is that the Renderer can't depend on the Item model - the item model depends on the renderer. It only works right now due to a temporary definition pointing to the item model (it will be updated to a special renderer)

i understand now, but it worked correctly in the inventory. What I’m not understanding is how to properly render the Robit on the ground. It looks completely broken haha.

I tried two approaches:

* Mapping it via JSON, but without using MapCodec
* Using MapCodec, but then it rendered in a weird way

Could you help me figure out the correct way to do this? I’ve spent hours trying to get it working.

Captura de Tela 2026-05-03 às 15 54 51

edsuuu:
total broken, I also feel a bit confused with the code because there are so many classes involved, which makes it harder to understand how everything connects.

edsuuu:
I ended up reverting the changes that broke it, but I was building the geometry directly in the model class. Does that seem correct, or should I be using the JSON instead?

thiakil:
It needs to support loading the `robit.json` and `robit_allay.json` models (and I think have the capability for others)

edsuuu:
> It needs to support loading the `robit.json` and `robit_allay.json` models (and I think have the capability for others)

I took a pause and managed to fix it. Could you review the code to see if it complies with the new rendering system in version 26.1?

image

thiakil:
Still doesn't seem to be handling the skins

edsuuu:
I’ll look for a better approach and open a new fully requested pull request.

PR #8595, opened 2026-07-03
Move upgrades from an enum to a datapack registry

pupnewfster:
- Allows custom upgrade types to be added (mainly of use for addons and machines they may add)
- Allows overriding the color and max installable amount for upgrades (config for maxUpgradeMultiplier still exists so there still isn't much of a reason to do so, but hopefully it will help stop people from using mixin to change this value)


src/datagen/main/java/mekanism/client/model/MekanismModelProvider.java
```diff
@@ -596,13 +597,7 @@ private void registerManualItemModels(ItemModelGenerators itemModels) {
...
         itemModels.declareCustomModelItem(MekanismItems.ULTIMATE_CONTROL_CIRCUIT.asItem());
         itemModels.declareCustomModelItem(MekanismItems.ULTIMATE_TIER_INSTALLER.asItem());
-        itemModels.declareCustomModelItem(MekanismItems.ANCHOR_UPGRADE.asItem());
-        itemModels.declareCustomModelItem(MekanismItems.CHEMICAL_UPGRADE.asItem());
-        itemModels.declareCustomModelItem(MekanismItems.ENERGY_UPGRADE.asItem());
-        itemModels.declareCustomModelItem(MekanismItems.FILTER_UPGRADE.asItem());
-        itemModels.declareCustomModelItem(MekanismItems.MUFFLING_UPGRADE.asItem());
-        itemModels.declareCustomModelItem(MekanismItems.SPEED_UPGRADE.asItem());
-        itemModels.declareCustomModelItem(MekanismItems.STONE_GENERATOR_UPGRADE.asItem());
+        itemModels.itemModelOutput.accept(MekanismItems.UPGRADE.asItem(), UpgradeItemModel.Unbaked.INSTANCE);
```
thiakil: this should probably be moved to a different place, this method is for the static jsons

src/datagen/generated/mekanism/assets/mekanism/lang/en_us.json
```diff
@@ -2524,13 +2524,7 @@
   "item.mekanism.teleportation_core": "Teleportation Core",
   "item.mekanism.ultimate_control_circuit": "Ultimate Control Circuit",
   "item.mekanism.ultimate_tier_installer": "Ultimate Tier Installer",
-  "item.mekanism.upgrade_anchor": "Anchor Upgrade",
```
thiakil, line 2527: it looks like the DPR items didnt get a lang key generated?

src/main/java/mekanism/client/gui/element/custom/GuiSupportedUpgrades.java
```diff
@@ -70,15 +76,19 @@ public void drawBackground(GuiGraphicsExtractor guiGraphics, int mouseX, int mou
...
-        for (int i = 0; i < EnumUtils.UPGRADES.length; i++) {
-            Upgrade upgrade = EnumUtils.UPGRADES[i];
-            UpgradePos pos = getUpgradePos(i);
-            int xPos = relativeX + 1 + pos.x;
-            int yPos = relativeY + 1 + pos.y;
-            gui().renderItem(guiGraphics, UpgradeUtils.getStack(upgrade), xPos, yPos, 0.75F);
-            if (!supportedUpgrades.contains(upgrade)) {
-                //Make the upgrade appear faded if it is not supported
-                guiGraphics.fill(xPos, yPos, xPos + ELEMENT_SIZE, yPos + ELEMENT_SIZE, backgroundColor);
+        Registry<Upgrade> upgrades = gui().registryAccess().lookupOrThrow(MekanismRegistries.Keys.UPGRADES);
```
thiakil: I feel like the `Registry<>` could be cached on the tile/wrapper? once the level gets set of course

src/main/java/mekanism/client/model/item/UpgradeItemModel.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private static final ModelDebugName DEBUG_NAME = () -> "MekanismUpgradeItemModel";
+
+    private final Map<ResourceKey<Upgrade>, ItemModel> cache = new IdentityHashMap<>(); // contains all the baked models since they'll never change
+
+    private final ItemTransforms itemTransforms;
+    private final Matrix4fc transformation;
+    private final BakingContext context;
+
+    private UpgradeItemModel(BakingContext context, Matrix4fc transformation) {
```
thiakil: the 3 items used from the context should be constructor params & fields, rather than the context itself

src/main/java/mekanism/client/sound/SoundHandler.java
```diff
@@ -362,9 +369,13 @@ private float getTileVolumeFactor() {
             BlockEntity tile = WorldUtils.getTileEntity(Minecraft.getInstance().level, BlockPos.containing(getX(), getY(), getZ()));
             float retVolume = 1.0F;
 
-            if (tile instanceof IUpgradeTile upgradeTile && upgradeTile.supportsUpgrade(Upgrade.MUFFLING)) {
-                int mufflerCount = Math.min(upgradeTile.getUpgrades(Upgrade.MUFFLING), Upgrade.MUFFLING.getMax());
-                retVolume = 1.0F - (mufflerCount / (float) Upgrade.MUFFLING.getMax());
+            if (tile instanceof IUpgradeTile upgradeTile) {
+                Holder<Upgrade> mufflingUpgrade = mufflingReference.get();
```
thiakil: this block should probably go into `IUpgradeTile`? maybe we can get rid of the `mufflingReference`?

src/main/java/mekanism/client/model/item/UpgradeItemModel.java
```diff
@@ -0,0 +1,101 @@
...
+        ModelBaker modelBaker = context.blockModelBaker();
+        Material.Baked bakedMaterial = modelBaker.materials().get(new Material(texture), DEBUG_NAME);
+        ModelRenderProperties properties = new ModelRenderProperties(false, bakedMaterial, this.itemTransforms);
+        QuadCollection quads = modelBaker.compute(new ItemModelGenerator.ItemLayerKey(bakedMaterial, BlockModelRotation.IDENTITY, 0));
+        return new CuboidItemModelWrapper(Collections.emptyList(), quads, properties, transformation);
+    }
+
+    public record Unbaked() implements ItemModel.Unbaked {
+
+        private static final Identifier GENERATED_MODEL = ModelLocationUtils.decorateItemModelLocation("generated");
```
thiakil: I suspect it would be better to not lock it down to the generated model. Instead, a basic item json that is parented to generated, with layer 0 being a tex ref of `#upgrade` which is then remapped to the correct tex on bake

thiakil (changes requested):

src/datagen/generated/mekanism/assets/mekanism/lang/en_us.json
pupnewfster: We already have lang keys for the individual upgrades. Currently the item is called Upgrade and it has "Upgrade Type: Speed" as the first line of the tooltip. I was undecided whether to override the display name and add a lang key "%s Upgrade" that then combines it automatically. Should I do that? And if so do I leave the tooltip as well or replace it with this?

thiakil: Ohhh, that's fine then, I'll trust you looked in game

Issue #3402, opened 2016-06-24
[1.9.4] ThermoElectric Boiler and Steam Turbine not showing informations


Xiaminou:
The ThermoElectric Boiler is not longer showing the Boil Rate and Max Boil, they're showing 0 when it should be 390.
![2016-06-24_15 07 37](https://cloud.githubusercontent.com/assets/8670392/16338497/673377c2-3a1d-11e6-80fb-ece0508ef353.png)

The Steam Turbine doesn't show the right informations either, it's showing 45mb/t and 642 RF/t instead of 5k RF/t. It also looks like it's full of steam, when it's not.
![2016-06-24_15 01 50](https://cloud.githubusercontent.com/assets/8670392/16338509/855c41ac-3a1d-11e6-9e80-d8842a9775e9.png)
![2016-06-24_15 01 55](https://cloud.githubusercontent.com/assets/8670392/16338512/856024e8-3a1d-11e6-81cb-88972145b49a.png)
![2016-06-24_15 02 03](https://cloud.githubusercontent.com/assets/8670392/16338510/855df47a-3a1d-11e6-9880-b35c0f9fef2f.png)

It's producing power just fine but I don't know what caused it, and have no idea how to reproduce it, or make it behave normally again for that matter.

Xiaminou:
Broke a replaced a block, now the Boiler is fine but the turbine shows half the production.

Xiaminou:
I do have a theory about this, I think it's the same kind of bug as [this one](http://bt.industrial-craft.net/view.php?id=1857).
I added a tree farm nearby, and every time a tree grows or gets chopped down I see the particle effects as if the multiblock was re-forming.

aidancbrady:
I'm fairly certain this is an interaction issue, after a bunch of testing with the boiler and turbine I can't get this to happen.  Let me know if you can prove otherwise.

PR #8607, opened 2026-07-18
fix: multiple bugfixes and NeoForge version bump for 26.2

windy664:
## Summary

Multiple bugfixes and improvements for the 26.2 branch:

- **Fix client-side security owner assignment** — correct security owner handling on client side
- **Refactor getLookup methods** — use controller variable for cleaner frequency lookups
- **Guard missing frequency lookup on placement** — prevent NPE when frequency lookup fails during block placement
- **Avoid Ingredient acceptsItem lookup** — optimize ingredient matching to avoid unnecessary item lookups
- **Fix jetpack hover client state sync** — fix client-side jetpack hover state not syncing properly
- **Clamp universal cable saved energy share** — prevent energy share values from exceeding bounds
- **Bump NeoForge version to 26.2.0.23-beta** — update minimum NeoForge version requirement

All fixes are tested and working on the 26.2 branch.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,11 +7,11 @@ minecraft_version=26.2
 major_minecraft_version=26.2
 previous_minecraft_version=1.21.1
 previous_minor_minecraft_version=26.1.2
-neo_version=26.2.0.18-beta
+neo_version=26.2.0.23-beta
 mod_version=10.8.0
 #This determines the minimum version of forge required to use Mekanism
 # Only bump it whenever we need access to a feature in forge that is not available in earlier versions
-neo_version_range=[26.2.0.18-beta,)
+neo_version_range=[26.2.0.23-beta,)
 minecraft_version_range=[26.2,26.3)
 #This specifies what type of release it will be uploaded to CurseForge and Modrinth as
 # options are: alpha, beta, release
--- a/src/main/java/mekanism/common/content/network/transmitter/UniversalCable.java
+++ b/src/main/java/mekanism/common/content/network/transmitter/UniversalCable.java
@@ -178,12 +178,11 @@ public class UniversalCable extends BufferedTransmitter<EnergyHandler, EnergyNet
         if (hasTransmitterNetwork()) {
             EnergyNetwork transmitterNetwork = getTransmitterNetworkNN();
             if (!transmitterNetwork.energyContainer.isEmpty() && saveShareJournal.saveShare > 0) {
-                //TODO: I believe I fixed the save share distribution bug that caused this to be necessary. If this comes back up look into it again
-                // or reinstate the clamping
+                long networkEnergy = transmitterNetwork.energyContainer.getAmountAsLong();
                 //Clamp the value so that we can't error if the network's energy is less than the amount we are saving
-                //saveShareJournal.saveShare = Math.min(transmitterNetwork.energyContainer.getAmountAsLong(), saveShareJournal.saveShare);
-                transmitterNetwork.energyContainer.setEnergy(transmitterNetwork.energyContainer.getAmountAsLong() - saveShareJournal.saveShare, transaction);
-                buffer.setEnergy(saveShareJournal.saveShare, transaction);
+                long saveShare = Math.min(networkEnergy, saveShareJournal.saveShare);
+                transmitterNetwork.energyContainer.setEnergy(networkEnergy - saveShare, transaction);
+                buffer.setEnergy(saveShare, transaction);
             }
         }
     }
--- a/src/api/java/mekanism/api/recipes/ingredients/ItemStackIngredient.java
+++ b/src/api/java/mekanism/api/recipes/ingredients/ItemStackIngredient.java
@@ -70,12 +70,8 @@ public final class ItemStackIngredient implements InputIngredient<Item, ItemStac
     public boolean testType(TypedInstance<Item> instance) {
         Objects.requireNonNull(instance);
         Ingredient vanillaIngredient = ingredient.ingredient();
-        if (vanillaIngredient.isCustom()) {
-            //Component data might be necessary, make it into a stack and test it
-            return vanillaIngredient.test(IngredientCreatorAccess.item().createStack(instance));
-        }
-        //Vanilla ingredients don't need to check component data so we can just skip converting the resource to a stack
-        return vanillaIngredient.acceptsItem(instance.typeHolder());
+        //Component data might be necessary, make it into a stack and test it
+        return vanillaIngredient.test(IngredientCreatorAccess.item().createStack(instance));
     }
 
     @Override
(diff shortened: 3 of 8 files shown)


thiakil:
> Avoid Ingredient acceptsItem lookup — optimize ingredient matching to avoid unnecessary item lookups

No. What you've done is bad for performance

gradle.properties
```diff
@@ -7,11 +7,11 @@ minecraft_version=26.2
 major_minecraft_version=26.2
 previous_minecraft_version=1.21.1
 previous_minor_minecraft_version=26.1.2
-neo_version=26.2.0.18-beta
+neo_version=26.2.0.23-beta
 mod_version=10.8.0
 #This determines the minimum version of forge required to use Mekanism
 # Only bump it whenever we need access to a feature in forge that is not available in earlier versions
-neo_version_range=[26.2.0.18-beta,)
```
pupnewfster, line 14: Why

src/main/java/mekanism/common/content/network/transmitter/UniversalCable.java
```diff
@@ -178,12 +178,11 @@ public void takeShare(@Nullable TransactionContext transaction) {
         if (hasTransmitterNetwork()) {
             EnergyNetwork transmitterNetwork = getTransmitterNetworkNN();
             if (!transmitterNetwork.energyContainer.isEmpty() && saveShareJournal.saveShare > 0) {
-                //TODO: I believe I fixed the save share distribution bug that caused this to be necessary. If this comes back up look into it again
-                // or reinstate the clamping
+                long networkEnergy = transmitterNetwork.energyContainer.getAmountAsLong();
```
pupnewfster, line 181: Why, do you have a crash log?

windy664: [crash-2026-07-14_13.58.33-server.txt](https://github.com/user-attachments/files/30152505/crash-2026-07-14_13.58.33-server.txt)

[crash-2026-07-15_08.08.05-server.txt](https://github.com/user-attachments/files/30152508/crash-2026-07-15_08.08.05-server.txt)

thiakil: capping it is just masking the problem, we need to know specifics of how much the amounts were before the overage. Can you add some logging for what `networkEnergy` and `saveShareJournal.saveShare` were at the time of overage?

If it was just by 1 I wouldn't care, but those amounts are something more serious wrong.

windy664: > capping it is just masking the problem, we need to know specifics of how much the amounts were before the overage. Can you add some logging for what `networkEnergy` and `saveShareJournal.saveShare` were at the time of overage?
> 
> If it was just by 1 I wouldn't care, but those amounts are something more serious wrong.

1f8d0af80b9cf9f768461a1c4cdfaaac

A power generation rate of only 80FE with full ethylene production.

thiakil: Not what I asked for, and certainly don't match the numbers in your crash logs

icedmoca:
## Still some issues.

**FluidType.Properties.isWaterLike is never set, so entities cannot move in any Mekanism fluid**


```
Branch: 26.2
Minecraft: 26.2 · NeoForge: 26.2.0.57
```

### Summary

Entities get no movement logic at all inside any Mekanism fluid. A player who walks into one stands on the surface, unable to move in any direction. Flying down into it works, because creative flight does not take that code path, and once inside movement is still impossible.

The cause is in `FluidDeferredRegister.getMekBaseBuilder():` the `FluidType.Properties` never sets `isWaterLike`, and `MekanismFluidType` does not override `FluidType#move`. NeoForge's contract on move is explicit:

▎ *If using custom movement logic, the method should return true. Otherwise, the movement logic will default to water if `getIsWaterLike()` returns `true` or no movement if it returns `false`.*

`isWaterLike` defaults to false and `FluidType#move` returns false by default, so neither branch applies and the entity receives no movement handling.

### Reproduction

1. Singleplayer, Mekanism only.
2. Place any Mekanism fluid with a bucket (reproduced with hydrogen, sodium, lithium, uranium oxide and heavy water).
3. Walk into it in survival.

**Expected:** the entity enters the fluid and moves through it.
Actual: the entity rests on the surface and cannot move. Flying in works; movement inside is still impossible.

### Why it looks like a NeoForge bug but isn't

Biomes O' Plenty's blood behaves correctly on the same instance, and it is the only fluid tested that calls .isWaterLike(true). Energized Power has the identical omission and the identical symptom. Vanilla water and lava are unaffected.

### Also ruled out, in case it saves anyone time:

- Block collision. Every Mekanism fluid block reports passable=true, liquid=true, solid=false, identical to minecraft:water.
- density / viscosity. Values span 71 (hydrogen) to 10,970 (uranium oxide) and all fail identically. Nothing reads them, because the movement path never runs.
- canSwim / canPushEntity. Both default to true and are correct.
- NeoForge #3402 ("Fix custom fluid movement only working when eyes are in fluid", in 26.2.0.57). Does not address this; the symptom persists on that build.

# Suggested fixes

Minimal, in getMekBaseBuilder():
```
return FluidType.Properties.create()
      .isWaterLike(true)
      .sound(SoundActions.BUCKET_FILL, SoundEvents.BUCKET_FILL)
      .sound(SoundActions.BUCKET_EMPTY, SoundEvents.BUCKET_EMPTY);
```

This restores movement in one line, but isWaterLike is broader than it appears. NeoForge derives isInWater, wasEyeInWater, splash effects, boat handling and fish AI from it, so every chemical would behave as water and the density and viscosity taken from real constants would still never be read.

### Preferred, overriding move in MekanismFluidType so the declared numbers matter:

```
@Override
public boolean move(LivingEntity entity, Vec3 movementVector, double gravity) {
    double relative = Mth.clamp(1000.0D / Math.max(1, this.getViscosity()), 0.15D, 1.6D);
    entity.moveRelative((float) (0.02D * relative), movementVector);
    entity.move(MoverType.SELF, entity.getDeltaMovement());

    double horizontalDrag = Mth.clamp(0.8D * relative, 0.4D, 0.89D);
    entity.setDeltaMovement(entity.getDeltaMovement().multiply(horizontalDrag, 0.8D, horizontalDrag));

    if (!entity.isNoGravity()) {
        double buoyancy = Mth.clamp(1000.0D / Math.max(1, this.getDensity()), 0.25D, 2.0D);
        entity.setDeltaMovement(entity.getDeltaMovement().add(0.0D, -gravity / 16.0D * buoyancy, 0.0D));
    }
    return true;
}
```

**Tested locally:** hydrogen feels thin and quick, sulfuric acid sluggish, uranium oxide close to tar. The clamps are only there to keep the extremes playable and are not tuned to any particular preference.

Note this overload requires building against `NeoForge 26.2.0.57` or newer; the branch currently targets `neo_version=26.2.0.23-beta`, which predates it. neo_version_range already permits it at runtime.

PR #8625, opened 2026-09-04
Expose datagen source set to test classpath

ghshhf:
## What this does
Adds `datagenMain.output` to the `test` source set's `compileClasspath`, mirroring the existing wiring that already exposes `main` and `api`. This lets unit tests reference classes that live in `src/datagen/main/java` (e.g. `FormatSplitter` and the language providers), which are currently invisible to `src/test`.

## Why
`FormatSplitter` is a pure-Java, zero-Minecraft-dependency utility, so it is a great candidate for fast unit tests — but the test source set could not see it. Rather than move it into `main` (changing its module), this just widens test visibility.

## Impact
- No production code changed. No runtime behavior changed.
- Only effect: `src/test/java` can now `import` classes from `src/datagen/main/java`.

[no-changelog]

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,6 +107,9 @@ configurations {
 
 //Add all extra source sets that the main sourceSet should have
 setupExtraSourceSets(sourceSets.main)
+//Expose datagen classes to the test source set so unit tests can cover datagen-only utilities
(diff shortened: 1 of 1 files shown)


thiakil:
pretty sure datagen would have its own automatic test sourceset, this should not be necessary

PR #8630, opened 2026-09-17
Add support for OpenComputers: Rebooted

CaitlynMainer:
This adds optional OpenComputers: Rebooted compatibility using the existing opencomputers ModID. Mekanism does not require OC to be installed for normal use.

## Changes proposed in this pull request
Add the OC:R API as a compile-only dependency and declare OpenComputers: Rebooted as an optional dependency.
Expose Mekanism's existing computer methods through OpenComputers: Rebooted components for blocks that already support computer integration.
Connect those components to the OpenComputers: Rebooted network so adjacent computers can discover and interact with them.
Keep the integration completely optional when OpenComputers: Rebooted is not installed.

This should also remain compatible with other OpenComputers forks that continue to use the opencomputers ModID.

Testing

I did some basic in-game testing with Mekanism and OC:R installed. An adjacent Basic Bin was correctly detected as a basicBin component, its methods were exposed, and the methods I tested behaved the same way as they do through Mekanism's CC:T integration.


build.gradle
```diff
@@ -730,7 +732,7 @@ if (System.getenv('CURSEFORGE_KEY') != null || project.hasProperty('curseforgeKe
             //Add optional deps
             addOptional(
                     //Computers
-                    'cc-tweaked', 'oc2',
+                    'cc-tweaked', 'oc2', 'opencomputers',
```
pupnewfster: This should be the CF's slug (the thing from the url of the curseforge page)

pupnewfster (commented):
Is there any meaningful difference to `OC: Rebooted` vs `OC2` which recentlyy got updated to 1.21.1 <https://www.curseforge.com/minecraft/mc-mods/oc2/files/8890230>? Is OC: Rebooted a version of the original OC mod but on newer mc versions?

build.gradle
CaitlynMainer: Yes, OC: Rebooted is the original OC, updated to 1.21.1, and *not* OC2, that was recently updated.
Also, apologies, I thought this wanted the ModID, which has not changed from the original "opencomputers"  I can update this to the CF ModSlug if that is what's needed.

